PR #77, opened 2023-06-19
price list settings

benefarr:


src/Http/Livewire/Settings/PriceLists.php
```diff
@@ -0,0 +1,224 @@
...
+                    'discounts' =>
+                        fn ($query) => $query->where('category_price_list.price_list_id', $priceList->id)
+                            ->select(['id', 'discount', 'is_percentage']),
+                ])
+                ->get()
+                ->toArray();
+        }
+
+        $this->editModal = true;
+        $this->resetErrorBag();
```
patrickweh, line 93: no need to reset this i guess as you are using the toast notifications to display validation errors

src/Http/Requests/CreatePriceListRequest.php
```diff
@@ -17,6 +17,10 @@ public function rules(): array
             'price_list_code' => 'required|string|unique:price_lists,price_list_code',
             'is_net' => 'required|boolean',
             'is_default' => 'boolean',
+
+            'discount' => 'exclude_without:parent_id|exclude_if:parent_id,NULL|array',
+            'discount.discount' => 'present|numeric|nullable',
```
patrickweh, line 22: should be between 1 and -1 if percentage

resources/views/livewire/settings/price-lists.blade.php
```diff
@@ -0,0 +1,134 @@
...
+                            <th class="w-14"></th>
+                        </x-slot>
+                        <template x-for="(productCategory, index) in productCategories">
+                            <tr>
+                                <td class="text-center">
+                                    <div x-text="productCategory.name" class="mr-2"></div>
+                                </td>
+                                <td>
+                                    <div class="flex justify-center">
+                                        <x-input x-model="productCategory.discounts[0].discount"/>
```
patrickweh, line 74: `<x-inputs.number x-model="productCategory.discounts[0].discount"/>`

please divide by 100 in livewire component if its percentage and multiply by 100 on load, for the user its not intuitive to add a below 0 value

src/Http/Livewire/Settings/PriceLists.php
```diff
@@ -0,0 +1,224 @@
...
+    public bool $editModal = false;
+
+    public function mount(): void
+    {
+        $this->priceLists = PriceList::query()
+            ->get()
+            ->toArray();
+    }
+
+    public function showEditModal(int $priceListId = null): void
```
patrickweh, line 53: reset the discounts if priceListId is null

patrickweh (changes requested):

src/Actions/PriceList/UpdatePriceList.php
```diff
@@ -26,18 +29,46 @@ public function performAction(): Model
     {
         $priceList = PriceList::query()
             ->whereKey($this->data['id'])
+            ->with(['discount'])
```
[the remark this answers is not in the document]
SirSplasch: it may be unnecessary, but this way func_get_args() will not be called, and should therefore be a little faster

patrickweh (approved):

PR #110, opened 2023-07-28
add action events

SirSplasch:


src/Actions/BaseAction.php
```diff
@@ -73,9 +89,45 @@ public function getRules(): array
         return $this->rules;
     }
 
-    public function validate(): static
+    final public function execute()
```
patrickweh, line 92: missing return type (mixed would be better than nothing imho)

src/Actions/BaseAction.php
```diff
@@ -73,9 +89,45 @@ public function getRules(): array
...
+
+        return $this;
+    }
+
+    protected function validateData(): void
     {
         $this->data = Validator::validate($this->data, $this->rules);
+    }
+
+    final public function withEvents(): static
```
patrickweh, line 121: this is to enable events if previously disabled?

src/Actions/BaseAction.php
patrickweh, line 1: i think it would be better to rename this file to FluxAction.
I have an action in my package that extends "BaseAction", would be more clear if it extends FluxAction

src/Actions/BaseAction.php
```diff
@@ -90,4 +142,24 @@ protected function convertEmptyStringToNull(array $data): array
             return $value === '' ? null : $value;
         }, $data);
     }
+
+    protected function fireActionEvent(string $event, object $object = null, bool $halt = true)
```
patrickweh, line 146: i think it would be better to always have the Action class in the event and set the return value to a property in the execute method. this gives more context for the event listener

patrickweh (changes requested):

src/Actions/BaseAction.php
SirSplasch: yes, for example if you want to disable events on validating but enable them again on execute

src/Actions/FluxAction.php
```diff
@@ -0,0 +1,176 @@
...
+
+    protected array $data;
+
+    protected array $rules = [];
+
+    protected mixed $result = null;
+
+    protected static Dispatcher $dispatcher;
+
+    protected static bool $isBroadcasting = true;
```
patrickweh, line 26: i think this should default to false and require an active opt in in the action class

src/Actions/FluxAction.php
```diff
@@ -0,0 +1,176 @@
...
+use Illuminate\Contracts\Container\BindingResolutionException;
+use Illuminate\Contracts\Events\Dispatcher;
+use Illuminate\Events\NullDispatcher;
+use Illuminate\Support\Facades\Validator;
+use Illuminate\Support\Str;
+use Spatie\Permission\Exceptions\UnauthorizedException;
+
+abstract class FluxAction
+{
+    use Makeable, BroadcastAction;
```
patrickweh, line 16: the trait should be used in the subclass (active opt in)

patrickweh (changes requested):

patrickweh (approved):

PR #129, opened 2023-08-18
refactor allowed fields to work with related records on api

SirSplasch:


src/Helpers/QueryBuilder.php
```diff
@@ -67,6 +62,39 @@ public static function filterModel(object $model, Request $request = null): Lara
...
+        $allowed = $model::getColumns();
+        $allowedFields = array_flip(
+            array_map(
+                fn ($item) => $model->getTable() . '.' . $item,
+                $allowed->pluck('Field')->toArray()
+            )
+        );
+        $queryBuilder->allowedFields(array_keys(array_merge($allowedFields, $relatedAllowedFields)));
+
+        if (count($includes) > 0) {
```
patrickweh, line 94: $includes could be null count on null is deprecated

patrickweh (changes requested):

src/Helpers/QueryBuilder.php
SirSplasch: array_diff returns an array, this can never be null

patrickweh (approved):

PR #139, opened 2023-08-30
Add cross selling feature

patrickweh:


src/Actions/Product/CreateProduct.php
```diff
@@ -54,6 +56,13 @@ public function performAction(): Product
             $product->prices()->createMany($prices);
         }
 
+        if ($productCrossSellings) {
```
SirSplasch, line 59: if statement unnecessary, foreach is enough because its defaulted to empty array. Its only an issue if you try to loop through something that isn't countable.

src/Actions/Product/CreateProduct.php
```diff
@@ -54,6 +56,13 @@ public function performAction(): Product
...
+            foreach ($productCrossSellings as $productCrossSelling) {
+                $productCrossSelling['product_id'] = $product->id;
+                CreateProductCrossSelling::make($productCrossSelling)->checkPermission()->validate()->execute()->id;
```
SirSplasch, line 62: if you want to check the permission do this beforehand, outside the foreach loop. In this case for each productCrossSelling the same permission is checked although its extremly unlikely that the users permissions do change while looping. And even if they change, this can cause an unexpected behaviour for the user.

SirSplasch, line 62: You know that this will eventually throw an exception. This may cause some confusion, because the product is created, but an exception is thrown. Rethink if you need to validate. Also the permission should be checked before anything happens with the product to avoid confusion.

SirSplasch, line 62: remove ->id doesn't do anything

src/Actions/Product/UpdateProduct.php
```diff
@@ -85,6 +90,30 @@ public function performAction(): Model
...
+            $activeProductCrossSellings = [];
+            foreach ($productCrossSellings as $productCrossSelling) {
+                $productCrossSelling['product_id'] = $product->id;
+                if ($productCrossSelling['id'] ?? false) {
+                    $action = UpdateProductCrossSelling::make($productCrossSelling);
+                } else {
+                    $action = CreateProductCrossSelling::make($productCrossSelling);
+                }
+
+                $activeProductCrossSellings[] = $action->checkPermission()->validate()->execute()->id;
```
SirSplasch, line 103: same as always, rethink if its that smart if an exception is thrown at this point in the code.

src/Actions/Product/UpdateProduct.php
```diff
@@ -85,6 +90,30 @@ public function performAction(): Model
...
+            }
+
+            $removedProductCrossSellings = $product->productCrossSellings
+                ?->whereNotIn('id', $activeProductCrossSellings)
+                ->pluck('id')
+                ->toArray();
+            foreach ($removedProductCrossSellings as $removedProductCrossSelling) {
+                DeleteProductCrossSelling::make(['id' => $removedProductCrossSelling])
+                    ->checkPermission()
```
SirSplasch, line 112: really need to check the permission for every deletion? Also if an exception is thrown this can cause some confusion for the user, because the product is already updated at this point.

src/Helpers/PriceHelper.php
SirSplasch, line 1: why these changes? For my understanding the original code was perfectly fine.
Is it because you want isCalculated set to true only if there is no direct price from product in price list?

Even if this is the case, you don't need to copy the price into an extra variable, just set a variable with value true.

src/Http/Livewire/Product/Product.php
```diff
@@ -120,6 +124,20 @@ public function save(): bool
                 ->toArray();
         }
 
+        if ($this->productCrossSellings !== null) {
+            $this->product['product_cross_sellings'] = collect($this->productCrossSellings)
```
SirSplasch, line 128: just use array_map() no need to collect()->toArray()

src/Http/Livewire/Product/Product.php
```diff
@@ -120,6 +124,20 @@ public function save(): bool
...
+                ->map(function (array $productCrossSelling) {
+                    $productCrossSelling['products'] = collect($productCrossSelling['products'])
```
SirSplasch, line 130: same here again, just user array_map()

src/Http/Livewire/Product/Product.php
```diff
@@ -120,6 +124,20 @@ public function save(): bool
...
+                        ->map(function (array $product) {
```
SirSplasch, line 131: callback can be one-lined fn ($product) => $product['id']

SirSplasch (changes requested):

src/Helpers/PriceHelper.php
patrickweh: the original code added isCalculated = true even if it was a "real" price model with an database entry.
i need to know if the price is real or just virutal, the change works fine for this purpose

src/Http/Requests/UpdateProductCrossSellingRequest.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+namespace FluxErp\Http\Requests;
+
+use FluxErp\Rules\ExistsWithIgnore;
+use Illuminate\Foundation\Http\FormRequest;
```
SirSplasch, line 6: remove unused import

src/Actions/Product/UpdateProduct.php
```diff
@@ -85,6 +91,35 @@ public function performAction(): Model
...
+                $productCrossSelling['product_id'] = $product->id;
+                if ($productCrossSelling['id'] ?? false) {
+                    $action = UpdateProductCrossSelling::make($productCrossSelling);
+                } else {
+                    $action = CreateProductCrossSelling::make($productCrossSelling);
+                }
+
+                try {
+                    $activeProductCrossSellings[] = $action->checkPermission()->validate()->execute();
+                } catch (ValidationException) {}
```
SirSplasch, line 106: also catch UnauthorizedException

SirSplasch (changes requested):

SirSplasch (approved):

PR #140, opened 2023-08-31
Fix bundle order positions

patrickweh:


src/Actions/OrderPosition/CreateOrderPosition.php
```diff
@@ -82,6 +83,39 @@ public function performAction(): OrderPosition
...
+                        'purchase_price' => 0,
+                        'vat_rate_percentage' => 0,
+                        'is_net' => $orderPosition->is_net,
+                        'is_free_text' => false,
+                        'is_bundle_position' => true,
+                    ];
+                })
+                ->each(function (array $bundleProduct) {
+                    CreateOrderPosition::make($bundleProduct)
+                        ->validate()
```
SirSplasch, line 114: can throw a validation exception, which can cause some confusion, because the initial order position is already created at this point. Please rethink if ->validate() is needed, or if the exception should be caught.

src/Actions/OrderPosition/CreateOrderPosition.php
```diff
@@ -82,6 +83,39 @@ public function performAction(): OrderPosition
         unset($orderPosition->discounts);
         $orderPosition->save();
 
+        if ($this->data['product_id'] ?? false) {
+            $product = $orderPosition->product()->with('bundleProducts')->first();
+            $sortNumber = $orderPosition->sort_number;
+            $product->bundleProducts
+                ->map(function (Product $bundleProduct) use ($orderPosition, &$sortNumber) {
+                    $sortNumber++;
```
SirSplasch, line 91: this can screw the ordering, because there could be positions in this order with the same sort_number.
Therefore all other positions with higher sort_numbers than the initial order position should be updated accordingly.
Maybe you can find a satisfying solution for this issue.

SirSplasch (changes requested):

SirSplasch (approved):

PR #175, opened 2023-10-16
fix wrong data point for total net price

patrickweh:


src/Livewire/Contacts/Contact.php
```diff
@@ -84,14 +84,13 @@ public function mount(int $id = null): void
...
 
-        $contact->main_address = $contact->addresses
-            ->where('is_main_address', true)
-            ->first()
-            ->toArray();
-
         $this->avatar = $contact->getAvatarUrl();
 
         $this->contact = $contact->toArray();
+        $this->contact['main_address'] = $contact->addresses
```
SirSplasch, line 90: addresses() otherwise the collection will be queried, not the database

src/Livewire/DataTables/OrderPositionList.php
```diff
@@ -98,6 +98,20 @@ public function getResultFromQuery(Builder $query): array
         return $tree;
     }
 
+    public function getReturnKeys(): array
+    {
+        return array_merge(
+            parent::getReturnKeys(),
+            [
```
SirSplasch, line 105: order by table column order

helpers.php
```diff
@@ -389,3 +389,10 @@ function flux_path(string $path = ''): string
         return __DIR__ . ($path ? DIRECTORY_SEPARATOR . $path : $path);
     }
 }
+
+if (! function_exists('model')) {
+    function model(string $class): Illuminate\Database\Eloquent\Model
+    {
+        return app($class);
```
SirSplasch, line 396: This alone is not solving the issue. Is the intent to have something like a config file or anything?
Like 'user' => 'App\\Models\\User', and then every time the User model is needed, this function gets called with the parameter given from 'user' key? Or what is the Plan?

SirSplasch (changes requested):

PR #176, opened 2023-10-17
add new button on tickets list

SirSplasch:
add save and delete on ticket view
add notifications to ticket types settings via roles


resources/views/livewire/ticket/ticket.blade.php
```diff
@@ -76,25 +95,26 @@
                         </div>
                     </x-slot:header>
                     <div class="space-y-4">
-                        <x-state wire:model.live="ticketState" formatters="formatter.state" avialable="availableStates"/>
+                        <x-state wire:model.live="ticket.state" formatters="formatter.state" available="availableStates"/>
```
patrickweh, line 98: as this change will always be active after pressing save it doesnt have to be live anymore i guess

resources/views/livewire/ticket/ticket.blade.php
```diff
@@ -109,14 +129,16 @@
                                 <x-label>
                                     {{ __('Author') }}
                                 </x-label>
+                                <x-toggle :left-label=" __('User') " :label=" __('Contact') " wire:model.live="authorType" />
```
patrickweh, line 132: also not necessary to be live i guess

resources/views/livewire/ticket/tickets.blade.php
```diff
@@ -0,0 +1,77 @@
+<div>
+    <x-modal.card z-index="z-30" wire:model="showTicketModal" :title="__('New Ticket')">
+        <div class="space-y-8 divide-y divide-gray-200">
+            <div class="space-y-8 divide-y divide-gray-200">
+                <div x-data="{
+                        selectedAdditionalColumns: @entangle('selectedAdditionalColumns'),
+                        additionalColumns: @entangle('additionalColumns'),
+                        ticket: @entangle('ticket'),
```
patrickweh, line 8: better use $wire.entangle('ticket')

src/Livewire/DataTables/TicketList.php
```diff
@@ -19,6 +31,10 @@ class TicketList extends DataTable
         'created_at',
     ];
 
+    public array $columnLabels = [
+        'ticket_type.name' => 'Ticket Type',
```
patrickweh, line 35: i think you dont need this with the new data table version, would be more consistant with ticket type -> name in the label

src/Livewire/DataTables/TicketList.php
```diff
@@ -65,4 +145,84 @@ public function getFilterableColumns(string $name = null): array
...
+            $this->saveFileUploadsToMediaLibrary('attachments', $ticket->id, Ticket::class);
+        } catch (\Exception $e) {
+            exception_to_notifications($e, $this);
+        }
+
+        $this->notification()->success(__('Ticket created…'));
+
+        $this->showTicketModal = false;
+        $this->skipRender();
+        $this->loadData();
```
patrickweh, line 198: i would prefer a redirect to the newly created ticket

src/Livewire/DataTables/TicketList.php
```diff
@@ -65,4 +145,84 @@ public function getFilterableColumns(string $name = null): array
...
+    }
+
+    public function updatedAttachments(): void
+    {
+        $this->prepareForMediaLibrary('attachments');
+
+        $this->skipRender();
```
patrickweh, line 205: you can remove all skipRender() as with the new data table version rendering is skipped by default after initializing the component

src/Livewire/Settings/TicketTypeEdit.php
```diff
@@ -43,10 +47,15 @@ public function mount(): void
             null
         );
 
-        $appModels = get_subclasses_of(Model::class, 'FluxErp\\');
-        $moduleModels = get_subclasses_of(Model::class, 'Modules\\');
+        $fluxModels = get_subclasses_of(Model::class, 'FluxErp\\Models\\');
```
patrickweh, line 50: better use modelfinder from data tables for this as in the SearchBar component.
Should be more performant because its cached.

src/Livewire/Ticket/Ticket.php
```diff
@@ -23,24 +24,13 @@ class Ticket extends Component
...
     public array $additionalColumns = [];
 
-    public string $ticketState;
-
     public array $ticketTypes;
 
     public array $states;
 
-    public string $tab = 'features.comments.comments';
+    public bool $authorType = true;
```
patrickweh, line 31: $authorTypeContact would be a better name for this…

patrickweh (changes requested):

resources/views/livewire/ticket/ticket.blade.php
SirSplasch: This needs to be live, because there is a updatedAuthorType() function which updates authenticatable_type and authenticatable_id

patrickweh (approved):

PR #181, opened 2023-10-25
Add new widgets

patrickweh:


resources/js/apex-charts.js
```diff
@@ -1,3 +1,101 @@
...
+
+                    return series;
+                });
+
+                this.options = Object.assign(this.options, options);
+
+                this.renderChart();
+            });
+        },
+        updateData() {
```
SirSplasch, line 54: duplicate code, updateData() and getLivewireData() are nearly the exact same code, this can surely be extracted.

src/Livewire/Charts/Chart.php
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    public function boot(): void
+    {
+        if ($this->series) {
+            $this->skipRender();
+        }
+    }
+
+    public function calculateChart()
```
SirSplasch, line 79: abstract public function calculateChart(): void;
Also move after properties, but before any concrete function

src/Livewire/Charts/PieChart.php
```diff
@@ -0,0 +1,20 @@
...
+{
+    public ?array $chart = [
+        'type' => 'pie',
+    ];
+
+    public bool $showTotals = true;
+
+    public function render(): View|Factory
+    {
+        return view('flux::livewire.charts.bar-chart');
```
SirSplasch, line 18: mustn't this be flux::livewire.charts.donut-chart?

resources/views/livewire/charts/donut-chart.blade.php
SirSplasch, line 1: Why is donut the default form of circular charts? I would expect that a pie chart should be the "default"

resources/views/components/layouts/head/head.blade.php
```diff
@@ -6,12 +6,14 @@
 <meta name="csrf-token" content="{{ csrf_token() }}">
 <meta name="pusher-key" content="{{ config('broadcasting.connections.pusher.key') }}">
 <meta name="pusher-cluster" content="{{ config('broadcasting.connections.pusher.options.cluster', 'mt1') }}">
+<meta name="currency-code" content="{{ \FluxErp\Models\Currency::default()?->iso }}">
```
SirSplasch, line 9: Is this ever used anywhere? I don't see it, but I may also be blind.

src/Livewire/Widgets/RevenueBySalesRepresentative.php
```diff
@@ -0,0 +1,43 @@
...
+
+    public function calculateChart(): void
+    {
+        $baseQuery = Order::query()
+            ->whereNotNull('invoice_date')
+            ->whereNotNull('invoice_number')
+            ->whereNotNull('agent_id')
+            ->where('invoice_date', '>=', now()->subYear());
+
+        $revenueBySalesRepresentative = $baseQuery->clone()
```
SirSplasch, line 27: why needs the query to be cloned? It's only used once.

src/Livewire/Widgets/RevenueBySalesRepresentative.php
```diff
@@ -0,0 +1,43 @@
...
+use FluxErp\Models\Order;
+use Illuminate\Support\Str;
+
+class RevenueBySalesRepresentative extends DonutChart implements UserWidget
+{
+    public bool $showTotals = false;
+
+    public static function getLabel(): string
+    {
+        return Str::headline(class_basename(self::class));
```
SirSplasch, line 16: localize?

src/Livewire/Widgets/RevenuePurchasesProfitChart.php
```diff
@@ -0,0 +1,103 @@
...
+use FluxErp\Models\Order;
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Str;
+
+class RevenuePurchasesProfitChart extends BarChart implements UserWidget
+{
+    public static function getLabel(): string
+    {
+        return Str::headline(class_basename(self::class));
```
SirSplasch, line 17: localize?

src/Livewire/Widgets/RevenuePurchasesProfitChart.php
SirSplasch, line 1: multiple scales are set to non-zero and are then cast into an integer, that doesn't seem right. (calculations with bcmath)

src/Livewire/Widgets/RevenuePurchasesProfitChart.php
```diff
@@ -0,0 +1,103 @@
...
+        foreach ($keys as $key) {
+            $monthlyRevenue[$key] ??= 0;
+            $monthlyPurchases[$key] ??= 0;
+            $monthlyProfit[$key] ??= 0;
+        }
+
+        // remove all values that are zero in all series
+        foreach ($monthlyRevenue as $key => $value) {
+            if ($value === 0 && $monthlyPurchases[$key] === 0 && $monthlyProfit[$key] === 0) {
+                unset($monthlyRevenue[$key]);
```
SirSplasch, line 63: can be done in one unset() statement, no need for 3 separate ones

SirSplasch (changes requested):

resources/views/livewire/placeholders/horizontal-bar.blade.php
```diff
@@ -0,0 +1,9 @@
+<div class="p-6 w-full">
+    <p class="mb-4 animate-pulse text-base text-neutral-700 dark:text-white">
```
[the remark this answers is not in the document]
patrickweh: no need for relative in that case

resources/views/components/layouts/head/head.blade.php
patrickweh: yes, if no currency is available for a formatter this tag is used as currency.

src/Livewire/Widgets/RevenuePurchasesProfitChart.php
patrickweh: apex charts needs the value to be numeric, i use bcmath to calculate the correct float and cast it to int to just remove the decimals as this isnt important in that case

SirSplasch (approved):

PR #214, opened 2023-12-04
add make commands for flux stack

patrickweh:


src/Console/Commands/ControllerMakeCommand.php
SirSplasch, line 1: Not a huge fan of this command. Also this command on its own does not really work. Often times it throws an exception due to missing stub.

SirSplasch (changes requested):
On model make command, requests are only created if selected in combination with resource controller option.
This is especially an issue if you select Form Requests and Actions options.

PR #303, opened 2024-01-16
add table fields for is dark mode on users and addresses

patrickweh:


SirSplasch (changes requested):
migration missing which adds is_dark_mode column to users table

SirSplasch (approved):

PR #362, opened 2024-02-07
add helper function to create/update/delete related records

SirSplasch:
add cycle check in UpdateUser action

### DIFF
--- a/src/Helpers/Helper.php
+++ b/src/Helpers/Helper.php
@@ -102,15 +102,17 @@ class Helper
         string $updateAction,
         string $deleteAction
     ): void {
-        $existing = $model->$relation()->pluck('id')->toArray();
-        $model->$relation()->whereNotIn('id', Arr::pluck($related, 'id'))->delete();
+        $relatedKeyName = $model->$relation()->getRelated()->getKeyName();
+
+        $existing = $model->$relation()->pluck($relatedKeyName)->toArray();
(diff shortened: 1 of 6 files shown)


src/Helpers/Helper.php
```diff
@@ -94,4 +92,49 @@ public static function getHtmlInputFieldTypes(): array
             'select',
         ];
     }
+
+    public static function updateRelatedRecords(
```
patrickweh, line 96: would be better to use ->getKeyName() in case the key is not named id

src/Helpers/Helper.php
```diff
@@ -94,4 +92,49 @@ public static function getHtmlInputFieldTypes(): array
...
+    ): void {
+        $existing = $model->$relation()->pluck('id')->toArray();
+        $model->$relation()->whereNotIn('id', Arr::pluck($related, 'id'))->delete();
+
+        $canCreate = $createAction::canPerformAction(false);
+        $canUpdate = $updateAction::canPerformAction(false);
+        $updated = [];
+        foreach ($related as $item) {
+            $item = array_merge($item, [$foreignKey => $model->id]);
+            if (! ($item['id'] ?? false)) {
```
patrickweh, line 113: use data_get, looks better

patrickweh (changes requested):

PR #364, opened 2024-02-08
add product suppliers

SirSplasch:
add ModelExists Rule


resources/views/components/product/general.blade.php
```diff
@@ -38,6 +38,52 @@
             </div>
         @show
     </x-card>
+    <x-card class="space-y-2.5" :title="__('Suppliers')">
```
patrickweh, line 41: class="flex flex-col gap-4"

src/Http/Requests/CreateCustomEventRequest.php
```diff
@@ -8,11 +8,6 @@
 
 class CreateCustomEventRequest extends BaseFormRequest
```
patrickweh, line 9: I would mark this as deprecated...

src/Http/Requests/CreateValueListRequest.php
```diff
@@ -8,11 +8,6 @@
 
 class CreateValueListRequest extends BaseFormRequest
```
patrickweh, line 9: Probably also deprecated

patrickweh (changes requested):
Missing relation from address to products

patrickweh (approved):

PR #365, opened 2024-02-11
Add purchase invoices

patrickweh:


.github/workflows/run-tests-on-pr.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     name: Laravel (PHP ${{ matrix.php-versions }})
     runs-on: test_runner
     env:
-      DB_DATABASE: laravel
+      DB_DATABASE: testing
```
patrickweh, line 14: default database in sail is testing, gets created for every sail instance

SirSplasch (approved):

PR #400, opened 2024-03-05
fix mail sync and purchase invoice upload

patrickweh:
fix invalid media attachments in CreateMailMessage.php

add possibility to pass existing media id to CreatePurchaseInvoice.php

add possibility to add mime type on media upload

### DIFF
--- a/src/Actions/PurchaseInvoice/CreatePurchaseInvoice.php
+++ b/src/Actions/PurchaseInvoice/CreatePurchaseInvoice.php
@@ -74,22 +74,16 @@ class CreatePurchaseInvoice extends FluxAction
 
         $errors = [];
         $media = data_get($this->data, 'media');
-        $filePath = (
-            is_string($media) && is_file($media)
-                ? $media
-                : null
-        ) ?? (
-            is_a($media, \SplFileInfo::class)
-                ? $media->getRealPath()
-                : null
-        ) ?? (
-            data_get($this->data, 'media.id')
-                ? app(Media::class)->query()
-                    ->whereKey(data_get($this->data, 'media.id'))
-                    ->first()
-                    ->getPath()
-                : null
-        );
+        $filePath = match (true) {
+            is_string($media) && is_file($media) => $media,
+            is_a($media, \SplFileInfo::class) => $media->getRealPath(),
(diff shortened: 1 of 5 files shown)


src/Jobs/SyncMailAccountJob.php
```diff
@@ -60,15 +62,13 @@ public function handle(): void
...
-                fn (MailFolder $folder) => DeleteMailFolder::make(['id' => $folder->id])->validate()->execute()
-            );
+            ->each(fn (MailFolder $folder) => DeleteMailFolder::make(['id' => $folder->id])->validate()->execute());
 
         if ($this->onlyFolders) {
             return;
         }
 
-        foreach ($folders as $folder) {
+        foreach ($folders->reverse() as $folder) {
```
SirSplasch, line 71: why reverse?

src/Jobs/SyncMailAccountJob.php
SirSplasch, line 1: Resolve all Model (queries) via app container app(Model::class)->query()

src/Jobs/SyncMailAccountJob.php
```diff
@@ -126,8 +126,8 @@ private function getNewMessages(Folder $folder): void
                 ->get()
                 ->first()
                 ?->getUid() - 1
```
SirSplasch, line 128: null - 1 == true | this leads to unexpected behaviour

src/Actions/PurchaseInvoice/CreatePurchaseInvoice.php
```diff
@@ -64,8 +73,26 @@ protected function validateData(): void
         parent::validateData();
 
         $errors = [];
+        $media = data_get($this->data, 'media');
+        $filePath = (
```
SirSplasch, line 77: why not use match (true) {}?

SirSplasch (changes requested):

src/Jobs/SyncMailAccountJob.php
patrickweh: the imap package fetches in reversed order compared to any regular client, so the inbox is at bottom not at top

SirSplasch (approved):

PR #420, opened 2024-03-25
Update laravel 11

patrickweh:


composer.json
```diff
@@ -64,11 +63,17 @@
...
+            "Workbench\\App\\": "workbench/app/",
+            "Workbench\\Database\\Factories\\": "workbench/database/factories/",
+            "Workbench\\Database\\Seeders\\": "workbench/database/seeders/"
         }
     },
     "scripts": {
         "post-autoload-dump": [
+            "@clear",
+            "@prepare",
+            "@dusk:install-chromedriver",
```
SirSplasch, line 76: why install chrome driver on post-autoload-dump?

SirSplasch (changes requested):
every uuid column migration needs to be changed to char('uuid', 36)

SirSplasch (approved):

PR #429, opened 2024-04-03
add vite config keys for reverb

patrickweh:


routes/channels.php
SirSplasch, line 1: Shouldn't only be channels for models registered, which actually use the BroadcastsEvents trait?

src/FluxServiceProvider.php
SirSplasch, line 1: MorphMap should be enforced in the boot method, not in register

SirSplasch (changes requested):

routes/channels.php
```diff
@@ -24,89 +26,23 @@
...
-});
-
-Broadcast::channel('FluxErp.Models.Order', function ($user) {
-    return $user->can(channel_to_permission((new Order())->broadcastChannelRoute()));
-});
-
-Broadcast::channel('FluxErp.Models.Project.{project}', function ($user) {
-    return $user->can(channel_to_permission((new Project())->broadcastChannelRoute()));
-});
+foreach (Relation::morphMap() as $alias => $class) {
```
SirSplasch, line 29: $alias is an unused variable, or am I wrong?

SirSplasch (changes requested):

SirSplasch (approved):

PR #481, opened 2024-06-03
add create variants action

patrickweh:


src/Actions/Product/Variant/CreateVariants.php
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    public function performAction(): Collection
+    {
+        $parentProduct = app(Product::class)->query()
+            ->whereKey($this->data['product_id'])
+            ->with('clients:id')
+            ->first();
+        $product = $parentProduct->toArray();
+        unset(
```
SirSplasch, line 32: no need to unset most of these, just only select the fields you want.

SirSplasch (changes requested):

SirSplasch (approved):

PR #494, opened 2024-06-13
refactor user modifications to only hold the name and id of the cause…

patrickweh:
…r in the created_by, updated_by and deleted_by


SirSplasch (approved):

SirSplasch:
fix failed test

PR #500, opened 2024-06-26
add rounding

SirSplasch:


database/migrations/2024_06_20_165116_add_rounding_columns_to_price_lists_table.php
```diff
@@ -0,0 +1,30 @@
...
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    public function up(): void
+    {
+        Schema::table('price_lists', function (Blueprint $table) {
+            $table->string('rounding_method_enum')->default('none')->after('price_list_code');
+            $table->integer('rounding_precision')->nullable()->after('rounding_method_enum');
```
patrickweh, line 13: can this be negative?

src/Helpers/PriceHelper.php
```diff
@@ -334,4 +359,9 @@ private function getRootPrice(?PriceList $priceList, ?Price $price): ?Price
 
         return $parentPrice ?? $price;
     }
+
+    private function fireEvent(string $event): void
```
patrickweh, line 363: better use protected, gives the possibility to override the function in the project even if this is propably never the case here, doesnt hurt

src/Livewire/Forms/PriceListForm.php
```diff
@@ -0,0 +1,81 @@
...
+
+    public ?int $rounding_number = null;
+
+    public ?string $rounding_mode = null;
+
+    public ?bool $is_net = true;
+
+    public ?bool $is_default = false;
+
+    public ?array $discount = [
```
patrickweh, line 35: can this be null?

patrickweh (changes requested):

database/migrations/2024_06_20_165116_add_rounding_columns_to_price_lists_table.php
SirSplasch: yes, intended, so that it is possible to round to hundreds, etc.

src/Livewire/Forms/PriceListForm.php
SirSplasch: not really, I made this nullable, because it was easier to circumvent the issue if a price list has no discount. Otherwise I would get an exception during fill() method.

patrickweh (approved):

PR #509, opened 2024-07-09
add merge option to install assets command

patrickweh:


src/Console/Commands/InstallAssets.php
```diff
@@ -15,7 +16,8 @@ class InstallAssets extends Command
      */
     protected $signature = 'flux:install-assets
         {directory? : The directory to install the assets}
-        {--force : Overwrite existing files}';
+        {--force : Overwrite existing files}
+        {--merge : If a files i json parsable merge it recursively}';
```
SirSplasch, line 20: fix description, this sentence cannot be a correct.

src/Console/Commands/InstallAssets.php
SirSplasch, line 1: This current implementation of the merge option is partially misleading.
Currently only the package.json (or other json files) are being merged, all js files are being overwritten even though I may not want them to be overwritten.

src/Console/Commands/InstallAssets.php
```diff
@@ -113,18 +120,50 @@ public static function copyStubs(?array $files = null, bool $force = false, ?\Cl
...
+                                    Arr::dot($decodedContent),
+                                    Arr::dot($decodedOldContent)
+                                )
+                            ),
+                            JSON_PRETTY_PRINT
+                        )
+                    );
+                }
+
+                $merged = true;
```
SirSplasch, line 161: move this to the if statement above, it is only merged if $content is actually merged.

SirSplasch (changes requested):

SirSplasch (approved):

PR #512, opened 2024-07-11
add model query caching

patrickweh:


src/Models/Scopes/UserClientScope.php
```diff
@@ -13,25 +13,28 @@
...
     public function apply(Builder $builder, Model $model): void
     {
         // Don't apply scope if no user is authenticated
         if (! Auth::hasUser()) {
             return;
         }
 
-        $clients = ($user = Auth::user()) instanceof User ?
+        // write the clients to a static property to avoid multiple queries if the scope is used multiple times
+        static::$clients ??= ($user = Auth::user()) instanceof User ?
```
SirSplasch, line 26: Can't the user's clients be cached? They surely will not be updated that often. And even then could the cache be regenerated.

src/Support/Database/CachedBuilder.php
```diff
@@ -0,0 +1,57 @@
...
+        if ($modelQueryCacheResult) {
+            return $modelQueryCacheResult;
+        }
+
+        // if the result is not in the cache, we'll fetch it from the database
+        $result = parent::get($columns);
+
+        // store the result in the cache
+        Cache::put(
+            $this->cacheKey($this->getModel()),
```
SirSplasch, line 31: shouldn't this be called static?

src/Traits/CacheModelQueries.php
```diff
@@ -0,0 +1,28 @@
...
+
+use FluxErp\Support\Database\CachedBuilder;
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Support\Facades\Cache;
+
+trait CacheModelQueries
+{
+    public static function bootCacheModelQueries(): void
+    {
+        static::$builder = CachedBuilder::class;
```
SirSplasch, line 13: how is this working? On app start i get an exception: access to undeclared static property.

SirSplasch (changes requested):

SirSplasch (approved):

PR #514, opened 2024-07-12
don't run schedules if app is in maintenance

SirSplasch:
translate edit/delete calendar event confirm dialogs fix AddressMediaScope


src/Http/Requests/DownloadMultipleMediaRequest.php
```diff
@@ -7,7 +7,7 @@ class DownloadMultipleMediaRequest extends BaseFormRequest
     public function rules(): array
     {
         return [
-            'ids' => 'required|array|string',
```
patrickweh, line 10: Why no array? The route accepts post, so it could be an array

patrickweh (changes requested):

patrickweh (approved):

PR #527, opened 2024-07-23
fix event edit

patrickweh:


src/Livewire/Features/Calendar/FluxCalendar.php
```diff
@@ -152,7 +154,8 @@ public function saveEvent(array $attributes): array|false
                 return false;
             }
 
-            $actionResult = $this->event->getActionResult();
+            $actionResult = data_get($this->event->getActionResult(), 'created')
```
SirSplasch, line 157: why did you do this? Do you even grasp how this worked beforehand?

resources/views/vendor/tall-calendar/components/event-edit.blade.php
SirSplasch, line 1: Removing this view it not correct. This needs to be loaded.

SirSplasch (changes requested):

SirSplasch (approved):

PR #539, opened 2024-08-02
add map to address list

patrickweh:


src/Livewire/DataTables/AddressList.php
```diff
@@ -93,4 +101,47 @@ public function save(): false|RedirectResponse|Redirector
...
+    public function loadData(): void
+    {
+        parent::loadData();
+
+        if ($this->showMap) {
+            $this->updatedShowMap();
+        }
+    }
+
+    public function updatedShowMap(): void
```
SirSplasch, line 114: can these two functions be renderless?

SirSplasch (changes requested):
Only show map if at least one entry with lat, lng exists.

SirSplasch (approved):

PR #561, opened 2024-08-25
fix payment state being ignored

patrickweh:
fix discounts created as meta records on order positions

fix cover image not beeing resetted when deleting the current cover image

fix unable to change locked order state

add change payment state from payment reminder table

fix gross profit and margin beeing visible on purchase orders

### DIFF
--- a/src/Livewire/Order/Order.php
+++ b/src/Livewire/Order/Order.php
@@ -407,6 +407,8 @@ class Order extends OrderPositionList
         $this->order->address_delivery = $this->order->address_delivery ?: [];
         try {
             $action = UpdateOrder::make($this->order->toArray())->checkPermission()->validate();
+
+            $this->getAvailableStates('state');
         } catch (ValidationException|UnauthorizedException $e) {
             exception_to_notifications($e, $this);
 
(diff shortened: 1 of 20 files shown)


src/Livewire/Order/Order.php
```diff
@@ -407,6 +407,8 @@ public function save(): bool
         $this->order->address_delivery = $this->order->address_delivery ?: [];
         try {
             $action = UpdateOrder::make($this->order->toArray())->checkPermission()->validate();
+
+            $this->getAvailableStates('state');
```
SirSplasch, line 411: shouldn't the available states for delivery_state (and maybe payment_state) also be updated?

src/Livewire/Order/Order.php
```diff
@@ -526,9 +528,23 @@ public function fetchContactData(bool $replicate = false): void
...
+                'id' => $this->order->id,
+                'state' => $this->order->state,
+                'payment_state' => $this->order->payment_state,
+                'delivery_state' => $this->order->delivery_state,
+            ])
+                ->checkPermission()
+                ->validate()
+                ->execute();
+
+            $this->getAvailableStates('state');
```
SirSplasch, line 544: shouldn't the available states for delivery_state (and maybe payment_state) also be updated?

SirSplasch (changes requested):

SirSplasch (approved):

PR #568, opened 2024-09-02
fix seeders

patrickweh:
fix menu manager

add RendersWidgets trait

### DIFF
--- a/database/seeders/CurrencyTableSeeder.php
+++ b/database/seeders/CurrencyTableSeeder.php
@@ -9,10 +9,12 @@ class CurrencyTableSeeder extends Seeder
 {
     public function run(): void
     {
+        Currency::factory()->create([
+            'name' => 'Euro',
+            'iso' => 'EUR',
+            'symbol' => '€',
+            'is_default' => true,
+        ]);
         Currency::factory()->count(3)->create();
-
-        if (! Currency::query()->where('is_default', true)->exists()) {
-            Currency::factory()->create(['is_default' => true]);
-        }
     }
 }
--- a/database/seeders/CurrencyTableSeeder.php
+++ b/database/seeders/CurrencyTableSeeder.php
@@ -9,12 +9,15 @@ class CurrencyTableSeeder extends Seeder
 {
     public function run(): void
     {
-        Currency::factory()->create([
-            'name' => 'Euro',
-            'iso' => 'EUR',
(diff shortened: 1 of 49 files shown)


resources/views/components/dashboard/edit-dashboard.blade.php
```diff
@@ -0,0 +1,54 @@
+@use('FluxErp\Enums\TimeFrameEnum')
+<div class="flex flex-col md:flex-row gap-1.5">
+    <div x-cloak x-show="!editGrid" class="flex flex-col md:flex-row gap-1.5 items-center text-sm">
+        <x-select
+            class="p-2"
+            :options="array_map(function (TimeFrameEnum $timeFrame) {
```
SirSplasch, line 6: just use TimeFrameEnum::valuesLocalized() and then set option-key-value (instead of option-value and option-label)

src/Traits/RendersWidgets.php
SirSplasch, line 1: The naming of saveDashboard and cancelDashboard methods does not really fit the trait.

database/seeders/LanguageTableSeeder.php
SirSplasch, line 1: This way this seeder can not be run more than once without failing.
The 'language_code' column has a unique constraint.

database/seeders/CurrencyTableSeeder.php
SirSplasch, line 1: This way this seeder can not be run more than once without failing.
The 'iso' column has a unique constraint.

SirSplasch (changes requested):

src/Traits/RendersWidgets.php
```diff
@@ -0,0 +1,117 @@
...
+        // create new widgets, update existing widgets
+        foreach ($this->widgets as &$widget) {
+            $savedWidget = auth()->user()->widgets()->updateOrCreate(['id' => $widget['id']], $widget);
+            $widget['id'] = $savedWidget->id;
+        }
+
+        $this->widgets();
+    }
+
+    public function cancelWidgets(): void
```
SirSplasch, line 75: May I suggest renaming this method to resetWidgets?
To me cancelWidgets seems a little strange in this context in reference what it actually does.

SirSplasch (changes requested):

SirSplasch (approved):

PR #627, opened 2024-10-23
add default to vat_rates

SirSplasch:
set clients and vat_rat_id on CreateProduct based on parent product, or default values require vat_rate_id on products

### DIFF
--- a/src/Actions/VatRate/UpdateVatRate.php
+++ b/src/Actions/VatRate/UpdateVatRate.php
@@ -31,4 +31,16 @@ class UpdateVatRate extends FluxAction
 
         return $vatRate->withoutRelations()->fresh();
     }
+
+    protected function prepareForValidation(): void
+    {
+        if (($this->data['is_default'] ?? false)
+            && ! resolve_static(VatRate::class, 'query')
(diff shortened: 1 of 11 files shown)


src/Actions/VatRate/UpdateVatRate.php
```diff
@@ -31,4 +31,16 @@ public function performAction(): Model
...
+
+    protected function prepareForValidation(): void
+    {
+        if (($this->data['is_default'] ?? false)
+            && ! resolve_static(VatRate::class, 'query')
+                ->whereKeyNot($this->data['id'] ?? 0)
+                ->where('is_default', true)
+                ->exists()
+        ) {
+            $this->rules['is_default'] .= '|accepted';
```
patrickweh, line 43: could be an array when getting overridden

patrickweh (changes requested):

src/Actions/VatRate/UpdateVatRate.php
SirSplasch: could be string, array, or maybe doesn't even exist. This is a mare general issue and needs to be addressed in a different pull request.

patrickweh (approved):

PR #632, opened 2024-10-26
fix mail message import and auto assigning

patrickweh:


src/Models/Communication.php
```diff
@@ -135,4 +173,52 @@ public function getPrintViews(): array
             'communication' => CommunicationView::class,
         ];
     }
+
+    public function autoAssign(string $type, array|string $matchAgainst): void
+    {
+        $matchAgainst = Arr::wrap($matchAgainst);
+        $typeColumn = $type === 'email' ? 'email_primary' : 'phone';
```
SirSplasch, line 180: better use match ($type), and if not email or phone, then return.

src/Models/Communication.php
```diff
@@ -135,4 +173,52 @@ public function getPrintViews(): array
...
+            $contacts = $addresses->pluck('contact')->unique();
+
+            if ($contacts->isNotEmpty()) {
+                $contacts->each(
+                    fn (Contact $contact) => $contact->communications()->attach($this->id)
+                );
+            }
+        }
+
+        if (is_string($this->subject) && ! empty($this->subject) && $type === 'email') {
```
SirSplasch, line 205: shouldn't this stay in the CreateMailMessage action?

SirSplasch (changes requested):

src/Models/Communication.php
patrickweh: nope, because this applies to phone calls as well as mail messages

src/Models/Communication.php
```diff
@@ -135,4 +173,55 @@ public function getPrintViews(): array
             'communication' => CommunicationView::class,
         ];
     }
+
+    public function autoAssign(string $type, array|string $matchAgainst): void
+    {
+        $matchAgainst = Arr::wrap($matchAgainst);
+        $typeColumn = match ($type) {
+            'phone' => 'phone',
+            default => 'email_primary',
```
SirSplasch, line 182: please fix, so that this is not defaulted to something "useful"

SirSplasch (changes requested):

SirSplasch (approved):

PR #633, opened 2024-10-26
add delivery lock functionality

patrickweh:


src/Livewire/Forms/OrderForm.php
```diff
@@ -164,12 +168,15 @@ public function fill($values): void
```
SirSplasch, line 168: probably in the if statement above $values instanceof Order as well.

src/Rules/Numeric.php
```diff
@@ -2,47 +2,42 @@
...
-class Numeric implements InvokableRule
+class Numeric implements ValidationRule
 {
-    private string|int|null $min;
+    private string|int|float|null $min;
 
-    private string|int|null $max;
+    private string|int|float|null $max;
 
     public function __construct(string|int|null $min = null, string|int|null $max = null)
```
SirSplasch, line 15: also add float as allowed type

SirSplasch (changes requested):

PR #636, opened 2024-10-30
Use class uses recursive in model info

patrickweh:


SirSplasch (changes requested):
I ain't gonna review this, as this is mostly the same code as in dispatchable flux action pr.
Fix it, so that this is actually only the code that is new / has changed in this branch.

SirSplasch (approved):

PR #646, opened 2024-11-08
fix user edit

patrickweh:
add test

### DIFF
--- a/src/Livewire/Settings/UserEdit.php
+++ b/src/Livewire/Settings/UserEdit.php
@@ -86,16 +86,6 @@ class UserEdit extends Component
         );
     }
 
-    public function getRules(): array
-    {
-        $action = ($this->user['id'] ?? false) ? UpdateUser::make($this->user) : CreateUser::make($this->user);
-        $rules = $action->getRules();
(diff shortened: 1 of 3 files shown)


src/Livewire/Settings/UserEdit.php
```diff
@@ -121,22 +122,26 @@ public function show(?int $id = null): void
...
         }
 
         $action = ($this->user['id'] ?? false) ? UpdateUser::class : CreateUser::class;
 
         try {
             $user = $action::make($this->user)
                 ->checkPermission()
-                ->setRules($this->getRules())
+                ->setRulesFromRulesets()
+                ->mergeRules($this->getRules())
```
SirSplasch, line 144: can the getRules() method be removed? As this method is not very helpful.
Also mergeRules with ['password' => ['confirmed']] does override the other password validation rules, and only 'confirmed' remains.

SirSplasch (changes requested):

SirSplasch (changes requested):

SirSplasch (approved):

PR #647, opened 2024-11-08
fix replicate order

patrickweh:
add test


SirSplasch (approved):
The replicate order test does not really check if the replicated order is what is expected. It only checks if it was executed successfully and the set replicateOrder id may exist.

PR #648, opened 2024-11-08
fix order positions

patrickweh:
fix deleting a single position

add validation to prevent modifying locked orders orderpositions

add tests

### DIFF
--- a/resources/views/livewire/order/order-positions.blade.php
+++ b/resources/views/livewire/order/order-positions.blade.php
@@ -100,7 +100,13 @@
             <x-slot:footer>
                 <div class="flex justify-between gap-x-4">
                     <div x-show="$wire.orderPosition.id">
-                        <x-button flat negative :label="__('Delete')" x-on:click="$wire.deleteOrderPosition(); close();" />
+                        <x-button
+                            flat
+                            negative
+                            :label="__('Delete')"
+                            wire:click="deleteOrderPosition(); close();"
+                            wire:flux-confirm.icon.error="{{  __('wire:confirm.delete', ['model' => __('Order Position')]) }}"
+                        />
                     </div>
                     <div class="flex w-full justify-end">
                         <x-button flat :label="__('Cancel')" x-on:click="close" />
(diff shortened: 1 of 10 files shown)


resources/views/livewire/order/order-positions.blade.php
```diff
@@ -100,7 +100,13 @@ class="pb-4"
             <x-slot:footer>
                 <div class="flex justify-between gap-x-4">
                     <div x-show="$wire.orderPosition.id">
-                        <x-button flat negative :label="__('Delete')" x-on:click="$wire.deleteOrderPosition(); close();" />
+                        <x-button
+                            flat
+                            negative
+                            :label="__('Delete')"
+                            wire:click="deleteOrderPosition(); close();"
+                            wire:flux-confirm.icon.error="{{  __('wire:confirm.delete', ['model' => __('Order Position')]) }}"
```
SirSplasch, line 108: double whitespace before __()

resources/views/livewire/order/order-positions.blade.php
```diff
@@ -100,7 +100,13 @@ class="pb-4"
             <x-slot:footer>
                 <div class="flex justify-between gap-x-4">
                     <div x-show="$wire.orderPosition.id">
-                        <x-button flat negative :label="__('Delete')" x-on:click="$wire.deleteOrderPosition(); close();" />
+                        <x-button
+                            flat
+                            negative
+                            :label="__('Delete')"
+                            wire:click="deleteOrderPosition(); close();"
```
SirSplasch, line 107: wait this works? To call close(); afterwards?

src/Livewire/Order/OrderPositions.php
```diff
@@ -425,6 +432,21 @@ public function deleteSelectedOrderPositions(): void
         $this->reset('selected');
     }
 
+    #[Renderless]
+    public function deleteOrderPosition(): void
```
SirSplasch, line 436: why not return boolean, and handle this like every other delete?

SirSplasch (changes requested):

resources/views/livewire/order/order-positions.blade.php
patrickweh: sure, wire:click is basically the same as x-on:click="$wire.doSomeThing", so you can add whatever javascript you want afterwards

SirSplasch (approved):

PR #656, opened 2024-11-16
add tag settings

patrickweh:


src/Livewire/Forms/TagForm.php
```diff
@@ -28,4 +28,13 @@ protected function getActions(): array
             'delete' => DeleteTag::class,
         ];
     }
+
+    public function fill($values): void
+    {
+        $valueArray = is_array($values) ? $values : $values->toArray();
+        $valueArray['name'] = $values->name;
```
SirSplasch, line 35: if $values is an array, this throws an exception.

SirSplasch (changes requested):

PR #657, opened 2024-11-17
fix rendering of order type specific views

patrickweh:


src/Livewire/Order/Order.php
```diff
@@ -46,7 +47,8 @@ class Order extends Component
 {
     use Actions, CreatesDocuments, WithTabs;
 
-    protected string $view = 'flux::livewire.order.order';
+    #[Locked]
+    public string $view = 'flux::livewire.order.order';
```
SirSplasch, line 51: why change visibility to public?

SirSplasch (changes requested):

src/Livewire/Order/Order.php
patrickweh: Because protected is resetted to default After Rendering. Mount gets called only once.

PR #664, opened 2024-11-19
add addCartItem to cart Model

patrickweh:


src/Actions/CartItem/CreateCartItem.php
```diff
@@ -26,34 +30,36 @@ public function performAction(): mixed
...
+        $cartItem = app(CartItem::class, ['attributes' => $this->data]);
+        $cartItem->save();
+
+        return $cartItem->fresh();
+    }
 
-        $cart = resolve_static(Cart::class, 'query')
+    protected function prepareForValidation(): void
+    {
+        $this->cart = resolve_static(Cart::class, 'query')
```
SirSplasch, line 41: if cart does not exist an exception is thrown, property of type Cart cannot be null.

src/Actions/CartItem/CreateCartItem.php
```diff
@@ -26,34 +30,36 @@ public function performAction(): mixed
...
+                ->first(['vat_rate_id', 'name']);
+            $this->data['vat_rate_id'] ??= $this->product?->vat_rate_id;
+            $this->data['name'] ??= $this->product?->name;
+        }
+
         if (
-            $cart->authenticatable instanceof Address
-            && ! data_get($this->data, 'price')
-            && $product
+            $this->cart->authenticatable instanceof Address
```
SirSplasch, line 55: eventually cart can be null

src/Models/Cart.php
```diff
@@ -82,6 +86,48 @@ public function broadcastWith(): array
...
+            if ($cartItem = $this->cartItems()->where('product_id', $productId)->first()) {
+                $data['id'] = $cartItem->id;
+                $data['amount'] = bcadd($cartItem->amount, $data['amount']);
+
+                $action = UpdateCartItem::make($data);
+            } else {
+                $action = CreateCartItem::make($data);
+            }
+
+            $action->validate()->execute();
```
SirSplasch, line 125: why no checkPermission()?

src/Livewire/Cart/Cart.php
```diff
@@ -66,57 +62,17 @@ public function refresh(): void
...
-            try {
-                $action->validate()->execute();
-            } catch (ValidationException $e) {
-                exception_to_notifications($e, $this);
-
-                return;
-            }
+        try {
+            $this->cart()->addItems($products);
+        } catch (ValidationException $e) {
```
SirSplasch, line 67: if permission is checked catch corresponding exception

SirSplasch (changes requested):

src/Models/Cart.php
```diff
@@ -82,6 +86,48 @@ public function broadcastWith(): array
...
+            if ($cartItem = $this->cartItems()->where('product_id', $productId)->first()) {
+                $data['id'] = $cartItem->id;
+                $data['amount'] = bcadd($cartItem->amount, $data['amount']);
+
+                $action = UpdateCartItem::make($data);
+            } else {
+                $action = CreateCartItem::make($data);
+            }
+
+            $action->validate()->checkPermission()->execute();
```
SirSplasch, line 125: first checkPermission() then validate(). There's no need to validate the data if the user does not have the right to do so.

SirSplasch (changes requested):

SirSplasch (approved):

PR #666, opened 2024-11-19
fix install wizard

patrickweh:


tests/Livewire/InstallWizardTest.php
```diff
@@ -2,28 +2,182 @@
...
 use FluxErp\Livewire\InstallWizard;
 use FluxErp\Tests\TestCase;
 use Illuminate\Support\Facades\Config;
+use Illuminate\Support\Str;
 use Livewire\Livewire;
 
 class InstallWizardTest extends TestCase
 {
-    public function test_renders_successfully()
+    public function setUp(): void
```
SirSplasch, line 14: why even bother setting Config key in setup when it is set separately in every method except one?

SirSplasch (changes requested):

SirSplasch (approved):

PR #671, opened 2024-11-19
fix wrong gross net on order position add

patrickweh:


tests/Livewire/Order/OrderPositionsTest.php
```diff
@@ -22,6 +24,8 @@ class OrderPositionsTest extends BaseSetup
 {
     protected string $livewireComponent = OrderPositions::class;
 
+    protected $orderPositionForm;
```
SirSplasch, line 27: why does this not have a type?

tests/Livewire/Order/OrderPositionsTest.php
```diff
@@ -80,24 +82,22 @@ protected function setUp(): void
             ]);
 
         $this->order->calculatePrices()->save();
+
+        $this->orderPositionForm = new OrderForm(Livewire::new(OrderPositions::class), 'order');
```
SirSplasch, line 86: why is this property called orderPositionForm but an OrderForm is set?

tests/Livewire/Order/OrderPositionsTest.php
```diff
@@ -146,4 +140,44 @@ public function test_can_show_related_columns()
...
+            ->set('orderPosition.product_id', $product->id)
+            ->call('quickAdd')
+            ->assertStatus(200)
+            ->assertHasNoErrors()
+            ->assertReturned(true);
+
+        $this->assertEquals($orderPositionCount + 1, $this->order->orderPositions()->count());
+        $newOrderPosition = $this->order->orderPositions()->latest()->first();
+        $this->assertNotEquals(
+            $product->prices->first()->getNet($product->vatRate->rate_percentage),
```
SirSplasch, line 174: why not set a variable to the result of getNet() and getGross()? Then you don't need to call these methods again later.

tests/Livewire/Order/OrderPositionsTest.php
```diff
@@ -146,4 +140,47 @@ public function test_can_show_related_columns()
         $this->assertArrayHasKey('order.uuid', $component->get('data.data.0'));
         $this->assertEquals($this->order->uuid, $component->get('data.data.0')['order.uuid']);
     }
+
+    public function test_quick_add_order_position()
+    {
+        PriceList::query()->update(['is_default' => false]);
```
SirSplasch, line 146: add where constraint to only update price lists which are actually is_default = true.

SirSplasch (changes requested):

tests/Livewire/Order/OrderPositionsTest.php
```diff
@@ -146,4 +140,40 @@ public function test_can_show_related_columns()
         $this->assertArrayHasKey('order.uuid', $component->get('data.data.0'));
         $this->assertEquals($this->order->uuid, $component->get('data.data.0')['order.uuid']);
     }
+
+    public function test_quick_add_order_position()
+    {
+        PriceList::query()->where('is_default', false)->update(['is_default' => false]);
```
SirSplasch, line 146: where('is_default', true)
and wrap it

SirSplasch (changes requested):

SirSplasch (approved):

PR #679, opened 2024-11-22
add calendarable to project

patrickweh:
add badge to extended props on tasks and projects


src/Models/Project.php
```diff
@@ -123,4 +126,57 @@ public function calculateProgress(): void
...
+            'end' => $this->end_date?->endOfDay()->toDateTimeString(),
+            'status' => $this->state::$name,
+            'invited' => [],
+            'description' => $this->description,
+            'extendedProps' => [
+                'appendTitle' => $this->state->badge(),
+            ],
+            'allDay' => true,
+            'is_editable' => true,
+            'is_invited' => null,
```
SirSplasch, line 163: why is this null, this key suggests a boolean

src/Models/Task.php
```diff
@@ -167,7 +167,9 @@ public function toCalendarEvent(): array
             'status' => $this->state::$name,
             'invited' => [],
             'description' => $this->description,
-            'extendedProps' => null,
+            'extendedProps' => [
+                'appendTitle' => $this->state->badge(),
+            ],
             'allDay' => false,
             'is_editable' => true,
             'is_invited' => null,
```
SirSplasch, line 175: same here, why is this value null

SirSplasch (changes requested):

SirSplasch (approved):

PR #683, opened 2024-11-28
Fix InitLanguages Command

Slupi:


src/Console/Commands/Init/InitLanguages.php
```diff
@@ -56,7 +56,9 @@ public function handle(): void
                         ->where('language_code', $jsonLanguage['language_code'])
                         ->firstOrNew();
 
-                    if (! $language->exists) {
+                    if (! resolve_static(Language::class, 'query')
```
SirSplasch, line 59: is_null(data_get($language, 'id')) or does this not work either?

SirSplasch (changes requested):

PR #685, opened 2024-12-02
add discount to order

patrickweh:


database/migrations/2024_11_29_125731_add_columns_to_discounts_table.php
```diff
@@ -0,0 +1,28 @@
...
+use Illuminate\Support\Facades\Schema;
+
+return new class() extends Migration
+{
+    public function up(): void
+    {
+        Schema::table('discounts', function (Blueprint $table) {
+            $table->string('name')->nullable()->after('model_id');
+            $table->decimal('discount_percentage', 40, 10)->nullable()->after('discount');
+            $table->decimal('discount_currency', 40, 10)->nullable()->after('discount_percentage');
```
SirSplasch, line 14: I don't get why this column is named that way. To me this implies the currency of the discount, which does not make any sense to me.

database/migrations/2024_11_29_134608_add_totals_to_orders_table.php
```diff
@@ -0,0 +1,48 @@
...
+                ->nullable()
+                ->after('total_base_gross_price');
+            $table->decimal('total_base_discounted_gross_price', 40, 10)
+                ->nullable()
+                ->after('total_base_discounted_net_price');
+
+            $table->decimal('total_discount_percentage', 40, 10)
+                ->nullable()
+                ->after('total_vats');
+            $table->decimal('total_discount_currency', 40, 10)
```
SirSplasch, line 22: Same here, I don't get the currency.

resources/views/livewire/order/order-positions.blade.php
```diff
@@ -80,6 +80,15 @@ class="pb-4"
                                         :label="__('Vat rate')"
                                         wire:model.live="orderPosition.vat_rate_id"
                                     />
+                                    <template x-for="discount in $wire.orderPosition.discounts">
+                                        <x-input
+                                            prefix="%"
```
SirSplasch, line 85: why does this have the '%' prefix? It can either be a percentage discount or a flat discount.

resources/views/livewire/order/order-positions.blade.php
```diff
@@ -136,6 +145,29 @@ class="pb-4"
             </x-slot:footer>
         </x-card>
     </x-modal>
+    <x-modal name="add-discount">
+        <x-card>
+            <div class="flex flex-col gap-4">
+                <x-input
+                    prefix="%"
```
SirSplasch, line 152: why is it only possible to add a percentage discount this way?

src/Models/Order.php
```diff
@@ -521,6 +540,40 @@ public function calculateTotalNetPrice(): static
...
+            9
+        );
+
+        return $this;
+    }
+
+    public function calculateDiscounts(): static
+    {
+        $previous = $this->total_net_price;
+        foreach ($this->discounts()->ordered()->get() as $discount) {
```
SirSplasch, line 561: why select '*'?

SirSplasch (changes requested):

database/migrations/2024_11_29_125731_add_columns_to_discounts_table.php
patrickweh: do you have a better suggestion? the discount_currency displays the discount in € or $ or whatever the order is.

resources/views/livewire/order/order-positions.blade.php
patrickweh: reverted this, i will create a new PR for the "bulk add"

resources/views/livewire/order/order-positions.blade.php
patrickweh: everted this, i will create a new PR for the "bulk add"

database/migrations/2024_11_29_125731_add_columns_to_discounts_table.php
```diff
@@ -0,0 +1,28 @@
...
+            $table->renameColumn('sort_number', 'order_column');
+        });
+    }
+
+    public function down(): void
+    {
+        Schema::table('discounts', function (Blueprint $table) {
+            $table->renameColumn('order_column', 'sort_number');
+
+            $table->dropColumn(['name', 'order_column', 'discount_percentage', 'discount_flat']);
```
SirSplasch, line 25: wrap array

resources/views/livewire/order/order-positions.blade.php
SirSplasch: If this would be reverted, why is it still showing as a change?

src/Models/Order.php
```diff
@@ -521,6 +540,40 @@ public function calculateTotalNetPrice(): static
...
+            9
+        );
+
+        return $this;
+    }
+
+    public function calculateDiscounts(): static
+    {
+        $previous = $this->total_net_price;
+        foreach ($this->discounts()->ordered()->get(['id', 'discount', 'is_percentage']) as $discount) {
```
SirSplasch, line 561: to be fair this can be solved with ->reduce() as in the end one value is used.

SirSplasch (changes requested):

database/migrations/2024_11_29_125731_add_columns_to_discounts_table.php
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    public function down(): void
+    {
+        Schema::table('discounts', function (Blueprint $table) {
+            $table->renameColumn('order_column', 'sort_number');
+
+            $table->dropColumn([
+                'name',
+                'order_column',
```
SirSplasch, line 31: why would you drop order_column?

SirSplasch (changes requested):

SirSplasch (approved):

PR #688, opened 2024-12-03
Add calendar widget

patrickweh:


src/frontend-helpers.php
```diff
@@ -1,4 +1,5 @@
 <?php
+
```
SirSplasch, line 2: why added this new line

SirSplasch (changes requested):

src/frontend-helpers.php
patrickweh: pint

SirSplasch (approved):

PR #708, opened 2024-12-18
add advertising state to address

patrickweh:


database/migrations/2024_12_17_121217_add_advertising_state_to_addresses_table.php
```diff
@@ -0,0 +1,22 @@
...
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class() extends Migration
+{
+    public function up(): void
+    {
+        Schema::table('addresses', function (Blueprint $table) {
+            $table->string('advertising_state')->default('open')->after('contact_id');
```
SirSplasch, line 12: after 'password'

src/Models/Address.php
```diff
@@ -201,6 +204,7 @@ protected static function booted(): void
     protected function casts(): array
     {
         return [
+            'advertising_state' => AdvertisingState::class,
```
SirSplasch, line 207: after date_of_birth

src/Rulesets/Address/CreateAddressRuleset.php
```diff
@@ -39,6 +41,10 @@ public function rules(): array
                 'nullable',
                 app(ModelExists::class, ['model' => Language::class]),
             ],
+            'advertising_state' => [
+                'string',
```
SirSplasch, line 45: nullable?

src/Rulesets/Address/UpdateAddressRuleset.php
```diff
@@ -39,6 +41,10 @@ public function rules(): array
                 'nullable',
                 app(ModelExists::class, ['model' => Language::class]),
             ],
+            'advertising_state' => [
+                'string',
```
SirSplasch, line 45: nullable?

src/States/Order/OrderState.php
```diff
@@ -85,9 +84,4 @@ public static function getFrontendFormatter(...$args): string|array
                 ->map(fn ($key) => (new $key(''))->color()),
         ];
     }
-
-    public function toArray(): array|string
```
SirSplasch, line 89: does removing this have any effect on the existing code-base. Or was this just added but never used?

src/Livewire/Contact/Addresses.php
```diff
@@ -40,8 +40,20 @@ class Addresses extends Component
...
 
+    public array $availableStates = [];
+
     public function mount(): void
     {
+        $this->availableStates = app(Address::class)
+            ->getStatesFor('advertising_state')
+            ->map(function (string $state) {
+                return [
+                    'label' => __(ucfirst(str_replace('_', ' ', $state))),
```
SirSplasch, line 51: why not __(Str::headline($state))?

SirSplasch (changes requested):

src/Rulesets/Address/CreateAddressRuleset.php
patrickweh: would keep it not nullable as the column isnt nullable

src/States/Order/OrderState.php
patrickweh: the to array is in the flux state base class, as well as the arrayable contract

SirSplasch (approved):

PR #718, opened 2024-12-22
refactor action broadcasting

patrickweh:


src/Traits/Action/HasActionEvents.php
```diff
@@ -0,0 +1,83 @@
+<?php
+
+namespace FluxErp\Traits\Action;
+
+use Illuminate\Contracts\Events\Dispatcher;
+use Illuminate\Events\QueuedClosure;
+
+trait HasActionEvents
+{
+    public static function booting(QueuedClosure|callable|array|string $callback): void
```
SirSplasch, line 10: why are these methods public and not protected?

src/Traits/Action/HasActionEvents.php
SirSplasch, line 1: Why did you get rid of the NullDispatcher?

src/Actions/FluxAction.php
```diff
@@ -289,40 +255,4 @@ protected function convertEmptyStringToNull(array $data): array
...
-                static::$dispatcher = app()->make(Dispatcher::class);
-            } catch (BindingResolutionException) {
-                static::$dispatcher = new NullDispatcher(new \Illuminate\Events\Dispatcher());
-            }
-        } else {
-            static::$dispatcher = new NullDispatcher(static::$dispatcher);
-        }
-    }
-
-    public function __serialize(): array
```
SirSplasch, line 313: Why remove __serialize() and __unserialize() methods? These don't do any harm. If someone wants to serialize an action this is necessary, or when an action is passed to a constructor of an aysnc job it will be serialized.

src/Traits/Action/BroadcastsActionEvents.php
```diff
@@ -0,0 +1,98 @@
...
+
+    public function broadcastQueue(): ?string
+    {
+        //
+    }
+
+    public function broadcastWith(string $event): array
+    {
+        if (! $this->getResult() instanceof Model) {
+            $payload['result'] = $this->result instanceof Serializable
```
SirSplasch, line 84: but what if $this->result is a simple data type, like an integer, boolean, etc. Then this information is lost, as all will be null.

src/Traits/Action/BroadcastsActionEvents.php
SirSplasch, line 1: This trait is currently not used anywhere (as far as I can see). Is this intended?

src/Events/BroadcastableActionEventOccurred.php
```diff
@@ -0,0 +1,90 @@
...
+
+    protected string $event;
+
+    protected array $channels = [];
+
+    public string $connection;
+
+    public string $queue;
+
+    public function __construct(FluxAction $action, string $event)
```
SirSplasch, line 26: needs this all to be done? Cloning the action and serializing these bits one by one?
In my head there must be an easier solution, but I am not that deep into this, I just review this on a little below surface level.

SirSplasch (changes requested):

src/Traits/Action/HasActionEvents.php
patrickweh: to make it possible to register an event from a ServiceProvider eg `CreateContact::booting(fn($asdf) => dd($asdf));`

src/Events/BroadcastableActionEventOccurred.php
patrickweh: the problem is that result is mixed, could be PDO which cant be serialized but still needs the actionevent to trigger

src/Traits/Action/BroadcastsActionEvents.php
patrickweh: thats intended, for now i dont have actions that need broadcasting (i had but solved it in a different way)

src/Traits/Action/HasActionEvents.php
```diff
@@ -0,0 +1,84 @@
...
+        if (isset(static::$dispatcher)) {
+            $name = static::class;
+
+            static::$dispatcher->listen('action.' . $event . ': ' . $name, $callback);
+        }
+    }
+
+    protected function fireActionEvent(string $event, bool $halt = true): mixed
+    {
+        if (! isset(static::$dispatcher)) {
```
SirSplasch, line 57: is_null(static::$dispatcher)

src/Traits/Action/HasActionEvents.php
```diff
@@ -0,0 +1,84 @@
...
+            return null;
+        }
+
+        $function = $halt ? 'until' : 'dispatch';
+
+        return static::$dispatcher->{$function}('action.' . $event . ': ' . static::class, $this);
+    }
+
+    public static function setEventDispatcher(?Dispatcher $dispatcher = null): void
```
SirSplasch, line 66: move the last three methods above registerActionEvent()

src/Traits/Action/HasActionEvents.php
```diff
@@ -0,0 +1,84 @@
...
+    {
+        static::$dispatcher = $dispatcher;
+    }
+
+    final public function withEvents(): static
+    {
+        $this->setEventDispatcher(app('events'));
```
SirSplasch, line 73: shouldn't this method be called like a static method static::setEventDispatcher()?

src/Traits/Action/BroadcastsActionEvents.php
```diff
@@ -0,0 +1,104 @@
...
+
+        return null;
+    }
+
+    public function newBroadcastableEvent(string $event): BroadcastableActionEventOccurred
+    {
+        return new BroadcastableActionEventOccurred($this, $event);
+    }
+
+    public static function getBroadcastChannel(): string
```
SirSplasch, line 61: move below boot

src/Traits/Action/BroadcastsActionEvents.php
SirSplasch, line 1: protected methods after public

src/Traits/Action/BroadcastsActionEvents.php
```diff
@@ -0,0 +1,104 @@
...
+
+    public function broadcastWith(string $event): array
+    {
+        $result = $this->getResult();
+
+        if (! $result instanceof Model) {
+            $payload['result'] = $result instanceof Serializable
+                ? $result
+                : (
+                    is_int($result) || is_float($result) || is_string($result) || is_bool($result)
```
SirSplasch, line 89: Its not perfect but I guess its fine.
Like an array containing only simple data types would normally be totally fine, but ok.

src/Events/BroadcastableActionEventOccurred.php
SirSplasch, line 1: I will be honest, I don't know if this works, if it doesn't it needs to be fixed at a later point.

SirSplasch (changes requested):

SirSplasch (approved):

PR #729, opened 2025-01-11
add model get permissions

patrickweh:


lang/de.json
```diff
@@ -726,6 +730,7 @@
     "From": "Von",
     "General": "Allgemein",
     "German": "Deutsch",
+    "Get": "Öffnen",
```
SirSplasch, line 733: Why translate "Get" with "Öffnen"? This does not seem right to me. Either change the key or the value, otherwise this can get kinda confusing.

src/Console/Commands/Init/InitPermissions.php
```diff
@@ -99,19 +123,63 @@ private function registerActionPermission(): void
...
+                );
                 unset($this->currentPermissions[$permission->id]);
             }
         }
     }
 
+    private function registerModelGetPermission(): void
+    {
+        $this->info('Registering model get permissions…');
+        $roles = resolve_static(Role::class, 'query')->where('guard_name', 'web')->get();
```
SirSplasch, line 142: wrap

src/Console/Commands/Init/InitPermissions.php
```diff
@@ -99,19 +123,63 @@ private function registerActionPermission(): void
...
+        $newPermissions = [];
+
+        foreach (Relation::morphMap() as $alias => $class) {
+            if (in_array(HasModelPermission::class, class_uses_recursive($class)) && $class::hasPermission()) {
```
SirSplasch, line 146: use resolve_static(), don't directly call the method on the given class.

SirSplasch, line 146: wrap &&
wrap ) {

src/Traits/HasModelPermission.php
```diff
@@ -0,0 +1,44 @@
...
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Support\Collection;
+use Spatie\Permission\Guard;
+
+trait HasModelPermission
+{
+    protected static function bootHasModelPermission(): void
+    {
+        if (auth()->user() && ! auth()->user()->hasRole('Super Admin')) {
+            static::addGlobalScope('permissionCheck', function (Builder $builder) {
```
SirSplasch, line 15: why not use an extra class for this global scope?

src/Models/PriceList.php
SirSplasch, line 1: why has this model hasPermission = false?

SirSplasch (changes requested):
Fix failed tests
I'm unsure why you even added this, but ok.

patrickweh:
> Fix failed tests I'm unsure why you even added this, but ok.

i want to add the possibility to completely deny access to models for users or roles if wanted

src/Models/PriceList.php
patrickweh: PriceList is crucial for booting the application. Thats why it doesnt have a permission on it to read…

SirSplasch (approved):

PR #731, opened 2025-01-14
fix mail component

patrickweh:


src/Livewire/Mail/Mail.php
```diff
@@ -144,4 +131,27 @@ private function findFolderIdById($folders, $id): ?array
 
         return null;
     }
+
+    protected function loadFolders(): void
+    {
+        foreach ($this->mailAccounts as $mailAccount) {
+            app(MailFolder::class)
```
SirSplasch, line 138: does this statement even do anything? Or how does this work in a foreach?

SirSplasch (changes requested):

src/Livewire/Mail/Mail.php
patrickweh: adds/overwrites the children global scope. this ensures that children of children of children and so on are loaded depending on the account id

SirSplasch (approved):

PR #732, opened 2025-01-14
add generic work time widget

patrickweh:
add revenue by top customers widget

fix css on widgets


src/Livewire/Widgets/WorkTimes.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+namespace FluxErp\Livewire\Widgets;
+
+use Illuminate\Contracts\View\Factory;
+use Illuminate\Contracts\View\View;
+
+class WorkTimes extends MyWorkTimes
+{
+    public ?int $userId;
```
SirSplasch, line 10: Shouldn't this property be initialized with null?

SirSplasch (approved):

PR #733, opened 2025-01-15
remove translatable from tags

patrickweh:


src/Actions/Tag/UpdateTag.php
```diff
@@ -29,4 +30,23 @@ public function performAction(): Tag
...
+
+    protected function validateData(): void
+    {
+        parent::validateData();
+
+        if (resolve_static(Tag::class, 'query')
+            ->where('name', $this->getData('name'))
+            ->where(
+                'type',
+                resolve_static(Tag::class, 'query')->whereKey($this->getData('id'))->value('type')
```
SirSplasch, line 42: wrap

src/Actions/Tag/UpdateTag.php
```diff
@@ -29,4 +30,23 @@ public function performAction(): Tag
 
         return $tag->withoutRelations()->fresh();
     }
+
+    protected function validateData(): void
+    {
+        parent::validateData();
+
+        if (resolve_static(Tag::class, 'query')
```
SirSplasch, line 38: if name is null / not given, this does not need to be validated

src/Actions/Tag/UpdateTag.php
```diff
@@ -29,4 +30,23 @@ public function performAction(): Tag
...
+            ->where('name', $this->getData('name'))
+            ->where(
+                'type',
+                resolve_static(Tag::class, 'query')->whereKey($this->getData('id'))->value('type')
+            )
+            ->whereKeyNot($this->getData('id'))
```
SirSplasch, line 44: whereKeyNot() as first where condition

SirSplasch (changes requested):

SirSplasch (approved):

PR #734, opened 2025-01-15
minor tweaks

patrickweh:


src/Livewire/Accounting/TransactionList.php
SirSplasch, line 1: Inconsistent how $this->selected is being emptied. Like it works and does its job, but I prefer using the same way every time.
I don't know which one is currently used more often $this->selected = [], or $this->reset('selected').

src/Livewire/Settings/Tags.php
```diff
@@ -20,6 +22,28 @@ class Tags extends TagList
...
     public TagForm $tagForm;
 
+    #[Locked]
+    public bool $isSelectable = true;
+
+    #[Renderless]
+    public function deleteSelected(): void
+    {
+        try {
+            $this->getSelectedModelsQuery()->pluck('id')->each(function (int $id) {
```
SirSplasch, line 32: Inconsistent, I saw foreach and ->each().
Also what if the one tag can not be deleted, are the rest being deleted or does it jump out at the first failure?

SirSplasch (changes requested):

src/Livewire/DataTables/WorkTimeList.php
```diff
@@ -53,6 +51,52 @@ class WorkTimeList extends BaseDataTable
 
     public bool $isSelectable = true;
 
+    #[Renderless]
+    public function toggleIsBillable(bool $isBillable)
```
SirSplasch, line 55: missing return type

src/Livewire/Settings/Tags.php
```diff
@@ -20,6 +22,30 @@ class Tags extends TagList
 
     public TagForm $tagForm;
 
+    #[Locked]
+    public bool $isSelectable = true;
+
+    #[Renderless]
+    public function deleteSelected(): void
+    {
+        $this->getSelectedModelsQuery()
```
SirSplasch, line 31: please use foreach and return if one fails. Or do you think this should be like it is?

SirSplasch (changes requested):

SirSplasch (approved):

PR #738, opened 2025-01-17
add missing tags migration

patrickweh:


database/migrations/2025_01_15_073553_remove_translation_from_tags_table.php
```diff
@@ -0,0 +1,77 @@
...
+{
+    public function up(): void
+    {
+        if (DB::table('information_schema.table_constraints')
+            ->where('table_name', 'tags')
+            ->where('constraint_name', 'tags_chk_1')
+            ->exists()
+        ) {
+            DB::statement('ALTER TABLE tags DROP CHECK tags_chk_1');
+        }
```
SirSplasch, line 18: new line after }

database/migrations/2025_01_15_073553_remove_translation_from_tags_table.php
SirSplasch, line 1: I prefer using double quotes DB::raw(""), that way you don't need to escape all the single quotes and is therefore easier to read.
(If the DB statement gets a little more complex.)

SirSplasch (changes requested):

SirSplasch (approved):

PR #743, opened 2025-01-20
fix tag creation

patrickweh:


src/Models/Tag.php
```diff
@@ -21,6 +23,52 @@ public static function bootHasSlug(): void
         });
     }
 
+    public static function findFromString(string $name, ?string $type = null, ?string $locale = null)
```
SirSplasch, line 26: return type?

src/Models/Tag.php
```diff
@@ -21,6 +23,52 @@ public static function bootHasSlug(): void
...
+                    ->orWhere('slug', $name);
+            })
+            ->first();
+    }
+
+    public static function findFromStringOfAnyType(string $name, ?string $locale = null): Collection
+    {
+        return static::query()
+            ->where('name', $name)
+            ->orWhere('slug', $name)
```
SirSplasch, line 41: please no direkt orWhere(). Can mess with scopes.

SirSplasch (changes requested):

SirSplasch (approved):

PR #747, opened 2025-01-21
add contact discounts

patrickweh:


src/Livewire/Contact/Accounting/Discounts.php
```diff
@@ -0,0 +1,142 @@
...
+                    'wire:click' => 'delete(record.id)',
+                ])
+                ->color('negative'),
+        ];
+    }
+
+    protected function getBuilder(Builder $builder): Builder
+    {
+        return $builder
+            ->with('model')
```
SirSplasch, line 64: is this eager loading even needed?

src/Actions/Contact/CreateContact.php
SirSplasch, line 1: I am generally curious why/if discounts are addable on create

src/Rulesets/Contact/DiscountRuleset.php
```diff
@@ -0,0 +1,29 @@
...
+    public function rules(): array
+    {
+        return [
+            'discounts' => 'nullable|array',
+            'discounts.*.id' => [
+                'nullable',
+                'integer',
+                app(ModelExists::class, ['model' => Discount::class]),
+            ],
+            'discounts.*.sort_number' => 'required_without:id|integer|min:0',
```
SirSplasch, line 21: maybe even exclude these other keys if id is given? Because the purpose is only to sync discounts for given contact, not update them.

src/Actions/Contact/UpdateContact.php
```diff
@@ -34,6 +37,29 @@ public function performAction(): Model
         $contact->fill($this->data);
         $contact->save();
 
+        if (! is_null($discounts)) {
```
SirSplasch, line 40: why is this an attach only, why can't I detach discounts from a contact?

src/Actions/Contact/UpdateContact.php
```diff
@@ -34,6 +37,29 @@ public function performAction(): Model
...
+
+                    continue;
+                }
+
+                try {
+                    $attachDiscounts[] = CreateDiscount::make($discount)
+                        ->checkPermission()
+                        ->validate()
+                        ->execute();
+                } catch (ValidationException|UnauthorizedException) {
```
SirSplasch, line 55: is this intended to just let failed discounts fall through without letting the user know?

src/Actions/Contact/UpdateContact.php
```diff
@@ -34,6 +37,29 @@ public function performAction(): Model
...
+                    $attachDiscounts[] = $discountId;
+
+                    continue;
+                }
+
+                try {
+                    $attachDiscounts[] = CreateDiscount::make($discount)
+                        ->checkPermission()
+                        ->validate()
+                        ->execute();
```
SirSplasch, line 54: shouldn't only the id be added to the array? I am unsure what happens if a combination of ids and models are passed to the attach() method.

src/Actions/Contact/CreateContact.php
```diff
@@ -37,6 +41,29 @@ public function performAction(): Contact
...
+
+                    continue;
+                }
+
+                try {
+                    $attachDiscounts[] = CreateDiscount::make($discount)
+                        ->checkPermission()
+                        ->validate()
+                        ->execute();
+                } catch (ValidationException|UnauthorizedException) {
```
SirSplasch, line 59: if intended to make discounts addable on create, why "ignore" failed discounts and give no feedback to the user?

SirSplasch (changes requested):

src/Livewire/Contact/Accounting/Discounts.php
patrickweh: yes, its a morph and i want to show the name of the category or product that has this discount

src/Actions/Contact/CreateContact.php
patrickweh: for example when creating new contacts while migrating from a previous erp system

src/Rulesets/Contact/DiscountRuleset.php
patrickweh: true

resources/views/livewire/contact/accounting/discounts.blade.php
```diff
@@ -0,0 +1,72 @@
...
+                    ]
+                ]"
+            />
+        </div>
+        <x-toggle
+            :label="__('Is Percentage')"
+            wire:model="discountForm.is_percentage"
+        />
+        <x-inputs.number
+             :label="__('Discount')"
```
SirSplasch, line 58: wait why is there a min and max given. These boundaries are only useful if the discount is percentage based.
Also even if, why should it not be possible to have a negative discount, or 100% discount?

src/Actions/Contact/UpdateContact.php
```diff
@@ -34,6 +36,26 @@ public function performAction(): Model
         $contact->fill($this->data);
         $contact->save();
 
+        if (! is_null($discounts)) {
```
SirSplasch, line 39: if it is a sync empty array should be a valid option.
Also in the current usage previous discounts would be removed every time as only the id of the "saved" discount is passed.

SirSplasch (changes requested):

resources/views/livewire/contact/accounting/discounts.blade.php
patrickweh: added a second input for discount flat.
min makes sense to prevent negative discounts.
100% discount could be possible e.g. for marketing materials for a customer that other customers have to buy for example.

resources/views/livewire/contact/accounting/discounts.blade.php
```diff
@@ -0,0 +1,81 @@
...
+        <x-toggle
+            :label="__('Is Percentage')"
+            wire:model="discountForm.is_percentage"
+        />
+        <div x-cloak x-show="$wire.discountForm.is_percentage">
+            <x-inputs.number
+                 :label="__('Discount Percent')"
+                 wire:model="discountForm.discount"
+                 step="0.01"
+                 min="100"
```
SirSplasch, line 62: now it can only be 100

src/Actions/Contact/CreateContact.php
```diff
@@ -28,15 +29,32 @@ public static function models(): array
...
         $mainAddress = Arr::pull($this->data, 'main_address');
 
-        $this->data['price_list_id'] ??= PriceList::default()?->id;
-        $this->data['payment_type_id'] ??= PaymentType::default()?->id;
-        $this->data['currency_id'] ??= Currency::default()?->id;
-
         $contact = app(Contact::class, ['attributes' => $this->data]);
         $contact->save();
 
+        if (! is_null($discounts)) {
```
SirSplasch, line 38: if ($discounts) an empty array does nothing.

src/Actions/Contact/UpdateContact.php
```diff
@@ -34,13 +36,50 @@ public function performAction(): Model
...
+            $selectedDiscounts = [];
+
+            foreach ($discounts as $discount) {
```
SirSplasch, line 42: no need to create discounts if syncType === 'detach'

SirSplasch (changes requested):

PR #750, opened 2025-01-22
add phone mobile to address

patrickweh:


resources/views/components/address/address.blade.php
```diff
@@ -145,6 +145,27 @@ class="h-full rounded-l-md"
                 </div>
             </div>
         @show
+            @section('contact-channels.phone_mobile')
```
SirSplasch, line 148: remove one tab of indentation on this whole block.

SirSplasch (changes requested):

SirSplasch (approved):

PR #756, opened 2025-01-25
Fix GitHub issues

patrickweh:


src/Livewire/Settings/MailAccounts.php
```diff
@@ -193,6 +196,22 @@ public function testSmtpConnection(): void
         }
     }
 
+    #[Renderless]
+    public function sendTestMail(?string $to = null): void
+    {
+        try {
+            Validator::make(['to' => $to], ['to' => 'required|email'])->validate();
```
SirSplasch, line 203: In the corresponding form the sendTestMail() method accepts null. And $to is set so the authenticated user's email if null. Therefore I ask the following question. Why is 'to' required and not nullable?

src/Livewire/Settings/MailAccounts.php
```diff
@@ -193,6 +196,22 @@ public function testSmtpConnection(): void
...
+            $this->mailAccount->sendTestMail($to);
+
+            $this->notification()->success(__('Test mail sent'));
+        } catch (
+            ValidationException|TransportExceptionInterface $e
+        ) {
+            exception_to_notifications($e, $this);
+        }
+
```
SirSplasch, line 212: remove empty line

src/Livewire/Settings/MailAccounts.php
SirSplasch, line 203: wrap ->validate()

src/Livewire/WorkTime.php
```diff
@@ -130,6 +138,26 @@ public function save(): bool
         return true;
     }
 
+    #[Renderless]
+    public function recordSelected(array $data): void
+    {
+        $this->workTime->fill($data);
```
SirSplasch, line 144: why fill and later set properties extra? I don't really get it.

src/Rulesets/MailAccount/UpdateMailAccountRuleset.php
```diff
@@ -19,13 +19,13 @@ public function rules(): array
                 app(ModelExists::class, ['model' => MailAccount::class]),
             ],
             'protocol' => 'sometimes|required|string|max:255|in:imap,pop3,nntp',
-            'password' => 'nullable|string|max:255',
+            'password' => 'nullable|string|max:255|exclude_if:password,null',
```
SirSplasch, line 22: if it gets excluded if null, then it cannot be nullable, or am I mistaken? Just put the exclude_if at first position.

src/Livewire/Forms/OrderForm.php
```diff
@@ -198,6 +202,9 @@ public function fill($values): void
                         'order_column',
                         'is_percentage',
                     ]),
+                'orderType:id,name,order_type_enum',
+                'priceList:id,name,is_net',
+                'users:id,name',
             ]);
             $values = array_merge(
```
SirSplasch, line 209: new line before, looks pretty clamped.

src/Mail/GenericMail.php
```diff
@@ -25,18 +26,31 @@ class GenericMail extends Mailable
...
+        if ($this->mailMessageForm instanceof CommunicationForm) {
+            $this->client ??= $this->mailMessageForm->communicatable()?->client;
+        }
+
+        $this->client ??= data_get($this->mailMessageForm, 'client_id')
+            ? resolve_static(Client::class, 'query')
+                ->whereKey(data_get($this->mailMessageForm, 'client_id'))
+                ->first()
+            : Client::default();
         $this->client ??= $this->mailMessageForm->communicatable()?->client ?? Client::default();
```
SirSplasch, line 42: why is this done twice? First in the if statement above, and here again?
Also if mailMessageForm is an array, array->communicatable() throws an exception.

SirSplasch (changes requested):

src/Livewire/WorkTime.php
patrickweh: fill doesnt override if the value is null, so a contact without a description would keep the description of the previous model (e.g. description of a task)

src/Mail/GenericMail.php
```diff
@@ -25,18 +26,34 @@ class GenericMail extends Mailable
...
+            $this->client ??= $this->mailMessageForm->communicatable()?->client;
+        }
+
+        $this->client ??= data_get($this->mailMessageForm, 'client_id')
+            ? resolve_static(Client::class, 'query')
+                ->whereKey(data_get($this->mailMessageForm, 'client_id'))
+                ->first()
+            : (
+                $this->mailMessageForm instanceof CommunicationForm
+                    ? $this->mailMessageForm->communicatable()?->client
```
SirSplasch, line 43: But why must this be done twice? Like this will always be null at this point.

SirSplasch (changes requested):

SirSplasch (approved):

PR #760, opened 2025-01-28
fix order position task attachment

patrickweh:


src/Livewire/Order/OrderPositions.php
SirSplasch, line 1: It seems odd to have order_position_id, but also the morph which both lead to the same record.

patrickweh: when tasks are created from an order its important to keep a 1:1 relation which is not given when using the morph, thats why this additional column is needed

SirSplasch (approved):

PR #762, opened 2025-01-28
fix ticket creation from mail

patrickweh:


src/Actions/FluxAction.php
```diff
@@ -208,10 +223,20 @@ final public function execute(): mixed
             return false;
         }
 
+        $current = null;
+        if ($this->getActingAs()) {
```
SirSplasch, line 227: shouldn't there be checked if $this->getActingAs() defers from auth()->user()?
Like if I for Instance want to set $this->actingAs = null.

src/Actions/FluxAction.php
```diff
@@ -126,6 +129,18 @@ public static function description(): ?string
...
     }
 
+    public function actingAs(?Authenticatable $user = null): static
+    {
+        $this->actingAs = $user;
+
+        return $this;
+    }
+
+    public function getActingAs(): ?Authenticatable
```
SirSplasch, line 139: Is there really a need for this method? I currently don't see one, if you can enlighten me, I'll be thankful.

tests/Unit/Action/MailMessage/CreateMailMessageTest.php
```diff
@@ -0,0 +1,118 @@
...
+            ->create()
+            ->addresses()
+            ->first();
+
+        $this->mailAccount = MailAccount::factory()
+            ->has(MailFolder::factory()->state(['can_create_ticket' => true]))
+            ->create();
+    }
+
+    public function test_creates_ticket()
```
SirSplasch, line 43: test_create_ticket_from_mail_message

tests/Unit/Action/MailMessage/CreateMailMessageTest.php
```diff
@@ -0,0 +1,118 @@
...
+        $this->assertDatabaseHas('tickets', [
+            'title' => $subject,
+            'description' => $action->getData('text_body'),
+            'created_by' => $this->address->getMorphClass() . ':' . $this->address->getKey(),
+        ]);
+
+        $this->assertTrue($result->communications()->where('communications.id', $result->id)->exists());
+    }
+
+    public function test_creates_ticket_comment()
```
SirSplasch, line 78: test_add_comment_to_ticket_from_mail_message

tests/Unit/BaseSetup.php
SirSplasch, line 1: I'm not that sure, that this is needed for unit tests, as unit tests simply check the functionality of a single unit.

tests/Unit/Action/MailMessage/CreateMailMessageTest.php
SirSplasch, line 1: Test is vastly incomplete, as it only covers the creation of a ticket and a comment if the CreateMailMessage action was successful. But ok, I know it was not your intention to fully test the CreateMailMessage action.

SirSplasch (changes requested):

src/Actions/FluxAction.php
patrickweh: could be overridden in subclass, you could set the acting as from the validate action for example

PR #764, opened 2025-01-29
add page block on navigate

patrickweh:


SirSplasch (commented):
Looks fine to me, but I have not deeper insight into js.

resources/js/components/navigation-spinner.js
```diff
@@ -0,0 +1,54 @@
...
+            setTimeout(() => {
+                overlay.classList.add('hidden');
+                spinnerVisible = false;
+            }, 200);
+        } else {
+            overlay.classList.add('hidden');
+        }
+    }
+
+    function init() {
```
sajberkurajber90, line 44: maybe to remove function init - since you call it once in alpine.js - just call the content of the function in this file - no need to return init.

resources/js/alpine.js
```diff
@@ -36,7 +39,7 @@ Alpine.data('folder_tree', folders);
 
 document.addEventListener(
```
sajberkurajber90, line 40: document.addEventListener('livewire:navigated', wireNavigation, { once: true });

sajberkurajber90: also wireNavigation() put in separate file

sajberkurajber90 (approved):

PR #765, opened 2025-01-29
fix flare errors

patrickweh:


routes/frontend/web.php
```diff
@@ -123,6 +124,18 @@
                         Route::get('/{id?}', Contact::class)->where('id', '[0-9]+')->name('id?');
                         Route::get('/communications', CommunicationList::class)->name('communications');
                     });
+                Route::get(
```
SirSplasch, line 127: why is this no redirect route?

SirSplasch (changes requested):

routes/frontend/web.php
patrickweh: redirect route accepts only a string as target

resources/views/livewire/folder-tree.blade.php
```diff
@@ -238,6 +238,13 @@ class="flex-1"
                                     >
                                         <x-slot:append>
                                             <div class="absolute inset-y-0 right-0 flex items-center p-0.5">
+                                                <x-button
+                                                    x-on:click="$openDetailModal(selection.original_url)"
```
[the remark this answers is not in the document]
patrickweh: opens the url in a modal with an iframe that loads the parameter

src/Livewire/Forms/TicketForm.php
```diff
@@ -0,0 +1,72 @@
...
+
+    public ?string $created_at = null;
+
+    public ?string $created_by = null;
+
+    public ?string $updated_at = null;
+
+    public ?string $updated_by = null;
+
+    public array $users = [];
```
SirSplasch, line 42: public array $ticket_type is missing. Or at least this property is called in view.

routes/frontend/web.php
SirSplasch: I guess I'll accept it, but I'm not a huge fan, as this is hard to customize as only a callback is attached to this route.

SirSplasch (changes requested):

SirSplasch (approved):

PR #777, opened 2025-02-06
fix ticket additional columns

patrickweh:


src/Livewire/Portal/Ticket/TicketCreate.php
```diff
@@ -160,6 +162,14 @@ public function updatedTicketTypeId(): void
         $this->selectedAdditionalColumns = $this->ticketTypeId ?
             $ticketTypeAdditionalColumns[$this->ticketTypeId] ?? [] : [];
 
+        $this->ticket = Arr::except(
```
SirSplasch, line 165: why do Arr::except(), can just remove the array_merge() above where the $oldAdditionalColumns are being added

SirSplasch (changes requested):

src/Livewire/Portal/Ticket/TicketCreate.php
SirSplasch: nah nevermind, Arr::except() is good, but nonetheless the array_merge() of the $oldAdditionalColumns above should be removed. Or am I misreading this?

PR #781, opened 2025-02-10
drop unique for creditor number

patrickweh:


database/migrations/2025_02_10_095747_drop_unique_for_creditor_number_on_contacts_table.php
```diff
@@ -0,0 +1,22 @@
...
+    {
+        Schema::table('contacts', function (Blueprint $table) {
+            $table->dropUnique('contacts_creditor_number_client_id_unique');
+        });
+    }
+
+    public function down(): void
+    {
+        Schema::table('contacts', function (Blueprint $table) {
+            $table->unique(['creditor_number', 'client_id']);
```
SirSplasch, line 19: As always, adding this on existing records can throw an exception, as the unique constraint cannot be met eventually.

SirSplasch (changes requested):
Also I hope you know what you are doing by making contact bank connection and payment type updateable on locked orders.

SirSplasch (approved):

PR #782, opened 2025-02-10
add tax exemption

patrickweh:


resources/views/livewire/contact/accounting/general.blade.php
```diff
@@ -37,6 +37,15 @@
                 option-label="name"
                 option-value="id"
             />
+            <x-select
```
SirSplasch, line 40: Why can a contact not have tax exemptions on multiple vat rates?

SirSplasch (changes requested):
I don't get the premise why a contact or an order can only have one tax exempted vat rate. Also I find the naming vat_rate_id rather confusing as it suggests, that the contact or order has a vat rate.

SirSplasch (approved):

PR #787, opened 2025-02-12
refactor comments

patrickweh:


resources/views/components/features/comments/comment.blade.php
SirSplasch, line 1: why renamed comment to node?
I get why the name node is used in methods of comments.js, but here I don't.

resources/views/livewire/features/comments/comments.blade.php
SirSplasch, line 1: Here too, I question the variable name node. Why is it not defined as comment?
Why is it x-for=node in comments? and not comment in comments.

SirSplasch (changes requested):
You can overrule me on the naming of node -> comment, if you insist.

src/Actions/Comment/UpdateComment.php
```diff
@@ -30,4 +30,15 @@ public function performAction(): Model
 
         return $comment->withoutRelations()->fresh();
     }
+
+    protected function prepareForValidation(): void
+    {
+        if (! $this->getData('is_sticky')) {
```
SirSplasch, line 36: this way one cannot set is_sticky = false

src/Actions/Comment/UpdateComment.php
```diff
@@ -30,4 +30,15 @@ public function performAction(): Model
...
+            unset($this->data['is_sticky']);
+        }
+
+        if (! $this->getData('is_internal')) {
```
SirSplasch, line 40: this way one cannot set is_internal = false

src/Livewire/Features/Comments/Comments.php
```diff
@@ -75,201 +68,176 @@ public function mount(): void
...
+        ]);
 
         try {
-            UpdateComment::make([
-                'id' => $comment->id,
-                'is_sticky' => ! $comment->is_sticky,
-            ])
-                ->validate()
-                ->execute();
+            $this->commentForm->save();
```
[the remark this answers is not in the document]
SirSplasch: This is not really solved, as unsetting is_sticky/is_internal prevents is_sticky/is_internal = false

SirSplasch (approved):

PR #812, opened 2025-03-11
fix order project

patrickweh:
fix search for scoped models in SearchController.php

### DIFF
--- a/src/Http/Controllers/SearchController.php
+++ b/src/Http/Controllers/SearchController.php
@@ -3,10 +3,12 @@
 namespace FluxErp\Http\Controllers;
 
 use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\SoftDeletingScope;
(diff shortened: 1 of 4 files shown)


src/Http/Controllers/SearchController.php
```diff
@@ -39,10 +41,19 @@ class_uses_recursive(resolve_static($model, 'class'))
                 ? $query->whereIn($optionValue, Arr::wrap($selected))
                 : $query->where($optionValue, $selected);
         } elseif ($request->has('search') && $isSearchable) {
+            /** @var Builder $perPageSearch */
+            $perPageSearch = count(Arr::except(
+                app($model)->getGlobalScopes(),
+                [
+                    SoftDeletingScope::class,
+                    SearchableScope::class,
```
SirSplasch, line 49: exclude UserClientScope

SirSplasch (changes requested):

PR #819, opened 2025-03-14
add auto render for forms

patrickweh:
add trait to handle form interaction for data tables


src/Traits/Livewire/DataTableHasFormEdit.php
```diff
@@ -0,0 +1,185 @@
...
+use Illuminate\Validation\ValidationException;
+use InvalidArgumentException;
+use Livewire\Attributes\Renderless;
+use Spatie\Permission\Exceptions\UnauthorizedException;
+use TeamNiftyGmbH\DataTable\DataTable;
+use TeamNiftyGmbH\DataTable\Htmlables\DataTableButton;
+
+trait DataTableHasFormEdit
+{
+    public function bootSupportsFluxForms(): void
```
SirSplasch, line 16: is this even executed, as these boot methods are run on eloquent models, but otherwise one has to implement this oneself.

src/Traits/Livewire/DataTableHasFormEdit.php
```diff
@@ -0,0 +1,185 @@
...
+    {
+        if (! $this instanceof DataTable) {
+            throw new InvalidArgumentException('This trait can only be used in a DataTable');
+        }
+    }
+
+    #[Renderless]
+    public function delete(string|int $id): bool
+    {
+        $model = resolve_static($this->model, 'query')->whereKey($id)->firstOrFail();
```
SirSplasch, line 26: wrap

src/Traits/Livewire/DataTableHasFormEdit.php
```diff
@@ -0,0 +1,185 @@
...
+        $this->reset('selected');
+    }
+
+    #[Renderless]
+    public function edit(string|int|null $id = null): void
+    {
+        $this->{$this->formAttributeName()}->reset();
+
+        if ($id) {
+            $model = resolve_static($this->model, 'query')->whereKey($id)->firstOrFail();
```
SirSplasch, line 75: wrap

SirSplasch, line 75: what happens if the ModelNotFound exception is thrown?

src/Traits/Livewire/DataTableHasFormEdit.php
```diff
@@ -0,0 +1,185 @@
...
+            $this->{$this->formAttributeName()}->fill($model);
+        }
+
+        $modalName = $this->modalName();
+        $this->js(<<<JS
+            \$modalOpen('$modalName');
+        JS);
+    }
+
+    abstract public function loadData(): void;
```
SirSplasch, line 85: abstract before non-abstract methods

SirSplasch (changes requested):

src/Traits/Livewire/DataTableHasFormEdit.php
patrickweh: implemented in datatables

src/Traits/Livewire/DataTableHasFormEdit.php
patrickweh: 404 shows up in the frontend

src/Livewire/Forms/FluxForm.php
```diff
@@ -19,6 +19,11 @@ abstract class FluxForm extends BaseForm
 
     abstract protected function getActions(): array;
 
+    public function canAction(string $action): bool
+    {
+        return resolve_static(data_get($this->getActions(), $action), 'canPerformAction', [false]);
```
SirSplasch, line 24: if data_get() returns null => exception is thrown as resolve_static expects a string as first parameter

SirSplasch (changes requested):

SirSplasch (approved):

PR #854, opened 2025-04-09
add currency input component

SirSplasch:


resources/views/components/input/currency.blade.php
```diff
@@ -0,0 +1,37 @@
+<div
+    x-data="{
+        input: @entangle($attributes->wire('model')),
```
patrickweh, line 3: Use $wire.$entangle, recommended according the docs

patrickweh (changes requested):

PR #861, opened 2025-04-11
Add printing frontend

patrickweh:


src/Models/Pivots/PrinterUser.php
```diff
@@ -16,6 +16,24 @@ class PrinterUser extends FluxPivot
 
     protected $table = 'printer_user';
 
+    protected static function booted(): void
+    {
+        static::saving(function (PrinterUser $printerUser): void {
+            if ($printerUser->isDirty('is_default') && $printerUser->is_default) {
+                $printerUser->user->printerUsers()
```
SirSplasch, line 23: why take a full turn around? this is just a 
resolve_static(PrinterUser::class, 'query')
->whereKeyNot($printerUser->getKey())
->where('user_id'. $printerUser->user_id')
->where('is_default', true)
->update(['is_default' => false])

src/Models/Printer.php
```diff
@@ -19,6 +19,11 @@ protected function casts(): array
         ];
     }
 
+    public function printerUsers(): HasMany
```
SirSplasch, line 22: what is this relation? This surely does not work, and the correct relation already exists in public function users()

src/Models/User.php
```diff
@@ -185,6 +185,11 @@ public function printers(): BelongsToMany
             ->using(PrinterUser::class);
     }
 
+    public function printerUsers(): HasMany
```
SirSplasch, line 188: Same here, the printers() relation already exists, one might add withPivot() or somthing along those lines, to the existing relation if anything.

resources/views/livewire/settings/tokens.blade.php
```diff
@@ -0,0 +1,59 @@
+<x-modal :id="$tokenForm->modalName()">
```
SirSplasch, line 1: Token abilities is somewhat important, if its possible to create tokens from frontend, then it should be possible to specify its abilities.

src/Traits/Livewire/CreatesDocuments.php
```diff
@@ -79,6 +82,19 @@ public function downloadPreview(): ?StreamedResponse
         );
     }
 
+    public function mountCreatesDocuments(): void
+    {
+        if (
+            $defaultPrinter = auth()->user()
+                ?->printerUsers()
+                ->where('is_default', true)
+                ->with('printer:id,media_sizes')->first()
```
SirSplasch, line 91: wrap ->first()

src/Traits/Livewire/CreatesDocuments.php
```diff
@@ -306,6 +341,20 @@ protected function createDocumentFromItems(
                 ->addMedia($files);
         }
 
+        if ($printIds && $this->printJobForm->printer_id) {
+            foreach ($printIds as $printId) {
```
SirSplasch, line 345: I am unsure if this works as intended, as on form->save() the form will be filled with the values from the action result. And therefore id may be set and causes an update() rather than a create()

SirSplasch:
I also strongly advise replacing the use of the printerUser relation with either printers() or users() relation.

src/Livewire/Forms/PrinterUserForm.php
```diff
@@ -0,0 +1,34 @@
...
+use FluxErp\Actions\PrinterUser\UpdatePrinterUser;
+use Livewire\Attributes\Locked;
+
+class PrinterUserForm extends FluxForm
+{
+    public ?string $default_size = null;
+
+    public bool $is_default = false;
+
+    public ?int $pivot_id = null;
```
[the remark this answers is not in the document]
patrickweh: because its bound by the default printer select

src/Livewire/Settings/UserEdit.php
```diff
@@ -175,8 +182,16 @@ public function save(): void
             exception_to_notifications($e, $this);
         }
 
-        $user->loadMissing(['roles', 'permissions', 'clients:id']);
-        $this->userForm->fill($user);
+        if ($this->printerUserForm->pivot_id) {
```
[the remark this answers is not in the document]
patrickweh: the pivot_id is filled when a default printer is chosen which has to be set as default…

src/Traits/Livewire/CreatesDocuments.php
patrickweh: you are right, needed a reset of the id on each loop

SirSplasch (approved):

PR #896, opened 2025-04-26
enhance LogsActivity trait to filter out state changes from logged at…

patrickweh:
…tributes

### DIFF
--- a/src/Traits/LogsActivity.php
+++ b/src/Traits/LogsActivity.php
@@ -31,7 +31,11 @@ trait LogsActivity
             $attribute = data_get($properties, 'attributes.' . $key);
             $old = data_get($properties, 'old.' . $key);
 
-            if ($old instanceof State && $attribute instanceof State && $old::class === $attribute::class) {
+            if (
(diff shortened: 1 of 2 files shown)


src/Traits/LogsActivity.php
```diff
@@ -4,10 +4,41 @@
...
+                data_get($properties, 'attributes') ?? [],
+                data_get($properties, 'old') ?? []
+            )
+        );
+
+        foreach ($changed as $key) {
+            $attribute = data_get($properties, 'attributes.' . $key);
+            $old = data_get($properties, 'old.' . $key);
+
+            if ($old instanceof State && $attribute instanceof State && $old::class === $attribute::class) {
```
SirSplasch, line 34: this may be unnecessary, but resolve_static($old, 'class') === resolve_static($attribute, 'class')?

SirSplasch, line 34: Wrap conditions

SirSplasch (changes requested):

PR #897, opened 2025-04-27
Add queued export functionality and notifications for data tables

patrickweh:


lang/de.json
```diff
@@ -7,6 +7,7 @@
     ":job_name is finished": ":job_name ist beendet",
     ":job_name is processing": ":job_name wird verarbeitet",
     ":job_name started": ":job_name gestartet",
+    ":model Export Ready": ":model Export Bereit",
```
SirSplasch, line 10: lowercase b?

routes/frontend/web.php
```diff
@@ -121,6 +122,12 @@
         Route::middleware(['auth:web', 'permission'])->group(function (): void {
             Route::get('/', Dashboard::class)->name('dashboard');
 
+            Route::get('/storage/{path}', function (string $path) {
+                return response()->file(Storage::disk('local')->path($path));
```
SirSplasch, line 126: may this not be an issue hard-coding used disk to 'local'?

src/Notifications/ExportReady.php
```diff
@@ -0,0 +1,46 @@
...
+    public function toToastNotification(object $notifiable): ToastNotification
+    {
+        return ToastNotification::make()
+            ->notifiable($notifiable)
+            ->title(__(':model Export Ready', ['model' => __(Str::headline($this->model))]))
+            ->description(__('Your export is ready for download.'))
+            ->accept(
+                NotificationAction::make()
+                    ->label(__('Download'))
+                    ->url(Storage::disk('local')->url($this->filePath))
```
SirSplasch, line 33: same here you assume the disk to be 'local', but this mustn't be the case.

src/Jobs/ExportDataTableJob.php
```diff
@@ -0,0 +1,48 @@
...
+        protected array $columns,
+        protected string $userMorph
+    ) {
+        $this->columns = array_filter($columns);
+    }
+
+    public function handle(): void
+    {
+        /** @var Model $model */
+        $model = new $this->modelClass();
```
SirSplasch, line 31: why use new? This disregards customization.

SirSplasch (changes requested):

routes/frontend/web.php
patrickweh: the disk is not given in the url that is generated by Storage::disk()->url()

SirSplasch (approved):

PR #899, opened 2025-04-28
fix: ensure correct order_id filtering in originPosition query

patrickweh:


src/Actions/OrderPosition/CreateOrderPosition.php
```diff
@@ -193,10 +193,11 @@ protected function validateData(): void
 
                 $originPosition = resolve_static(OrderPosition::class, 'query')
                     ->whereKey($originPositionId)
+                    ->where('order_id', $order->parent_id)
                     ->withSum('descendants as descendantsAmount', 'amount')
                     ->first();
                 $maxAmount = bcsub(
-                    $originPosition->amount,
+                    $originPosition?->amount ?? 0,
```
SirSplasch, line 200: no need for ?->amount, as the existence is checked beforehand

SirSplasch (changes requested):

src/Actions/OrderPosition/CreateOrderPosition.php
patrickweh: its checked for existance but not exited so this line is called even if the model doesnt exist

SirSplasch (approved):

PR #902, opened 2025-04-29
fix: handle missing main address in Address.php and add contact delet…

patrickweh:
…ion test

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,10 +21,9 @@
         "@tailwindcss/typography": "^0.5.10",
         "@tiptap/core": "^2.3.1",
         "@tiptap/extension-mention": "^2.5.7",
-        "@tiptap/extension-text-style": "2.11.5",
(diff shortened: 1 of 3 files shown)


SirSplasch (approved):

package.json
```diff
@@ -53,6 +52,7 @@
         "tailwindcss": "^3.4.0",
         "tippy.js": "^6.3.7",
         "uuid": "^11.1.0",
-        "vite": "^6.0.3"
+        "vite": "^6.0.3",
+        "@tiptap/extension-text-style": "2.11.5"
```
SirSplasch, line 56: why changed the ordering of the packages?

SirSplasch (changes requested):

package.json
patrickweh: Prettier

PR #950, opened 2025-05-27
add my leads widget

patrickweh:


src/Models/LeadState.php
```diff
@@ -33,7 +33,11 @@ protected function casts(): array
     public function getImageAttribute(): string
     {
         return route('avatar', [
-            'text' => $this->name,
+            'text' => Str::of($this->name)
+                ->replaceMatches('/[^A-Z]/', '')
+                ->trim()
+                ->limit(2, '')
```
SirSplasch, line 39: why return a string with length 2?

SirSplasch (changes requested):

PR #972, opened 2025-06-04
refactor: update FolderTree usage and improve data retrieval in Attachments

patrickweh:
## Summary by Sourcery

Refactor and centralize the folder-tree attachments feature by moving the base component into a Support namespace, generating specialized media components per model, and standardizing Blade views with data_get() and model ID bindings.

New Features:
- Introduce specialized Livewire media components for each model (e.g., portal.serial-number.media, ticket.media) to replace the generic folder-tree usage

Enhancements:
- Move and rename the base FolderTree component into a Support namespace and convert it into an abstract class with a protected modelType property
- Refactor existing attachments components to extend the new abstract FolderTree and declare their modelType instead of using Modelable attributes
- Replace generic <livewire:folder-tree> directives in Blade templates with the new model-specific media components and remove explicit model-type attributes
- Standardize data access in Blade views by switching to the data_get() helper for safe nested value retrieval
- Update Livewire tab configurations for Order and Task to bind attachments to the model’s id property

Tests:
- Update FolderTreeTest to import the renamed Support\FolderTree namespace

### DIFF
--- a/src/Livewire/Task/Media.php
+++ b/src/Livewire/Task/Media.php
@@ -1,0 +1,10 @@
+<?php
+
+namespace FluxErp\Livewire\Task;
+
+use FluxErp\Livewire\Support\FolderTree;
+
+class Media extends FolderTree
+{
+    protected string $modelType = \FluxErp\Models\Task::class;
+}
--- a/src/Livewire/Task/Media.php
+++ b/src/Livewire/Task/Media.php
@@ -3,8 +3,9 @@
 namespace FluxErp\Livewire\Task;
 
 use FluxErp\Livewire\Support\FolderTree;
+use FluxErp\Models\Task;
 
 class Media extends FolderTree
 {
-    protected string $modelType = \FluxErp\Models\Task::class;
+    protected string $modelType = Task::class;
 }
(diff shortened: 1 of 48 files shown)


src/Livewire/Task/Media.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace FluxErp\Livewire\Task;
+
+use FluxErp\Livewire\Support\FolderTree;
+
+class Media extends FolderTree
+{
+    protected string $modelType = \FluxErp\Models\Task::class;
```
SirSplasch, line 9: import class

tests/Livewire/Widgets/MyLeadsTest.php
```diff
@@ -0,0 +1,25 @@
...
+use FluxErp\Models\LeadState;
+use FluxErp\Tests\Livewire\BaseSetup;
+use Livewire\Livewire;
+
+class MyLeadsTest extends BaseSetup
+{
+    public function test_renders_successfully(): void
+    {
+        $lead = Lead::factory()
+            ->for(LeadState::factory()->state(['is_won' => false]))
```
SirSplasch, line 16: wrap, and also set is_lost = false

SirSplasch (changes requested):

PR #993, opened 2025-06-15
adding a widget displaying leads by referral source

ktownmaxi:


src/Livewire/Widgets/LeadsByReferralSource.php
```diff
@@ -0,0 +1,104 @@
...
+            ->setEndingDate($this->getEnd())
+            ->setStartingDate($this->getStart())
+            ->setLabelKey('addressRecommendedBy.name')
+            ->count('recommended_by_address_id', 'id');
+
+        $this->series = $metrics->getData();
+        $this->labels = $metrics->getLabels();
+    }
+
+    public function getPlotOptions(): array
```
patrickweh, line 57: I think this should be protected

lang/de.json
```diff
@@ -997,6 +997,8 @@
     "Latitude": "Breitengrad",
     "Lay out user": "Auslage Benutzer",
     "Lay Out User Id": "Auslage-Benutzer-ID",
+    "Leads recommended by :referralSourceName": "Leads empfohlen von :referralSourceName",
```
patrickweh, line 1000: Kebab case for the variable

patrickweh (changes requested):

PR #1000, opened 2025-06-23
add widget configuration and event handling methods

patrickweh:


src/Traits/Widgetable.php
```diff
@@ -4,9 +4,13 @@
 
 use Illuminate\Contracts\View\View;
 use Illuminate\Support\Str;
+use Livewire\Attributes\Locked;
 
 trait Widgetable
 {
+    #[Locked]
+    public ?string $dashboardComponent = null;
```
SirSplasch, line 12: is this property needed, or cant the dashboard component name be get by dashboardComponent method?

src/Traits/Livewire/SupportsWidgetConfig.php
```diff
@@ -0,0 +1,33 @@
...
+    public ?array $config = null;
+
+    protected function storeConfig(string $path, array|bool|string|int|float|Arrayable|null $value): void
+    {
+        $widget = auth()
+            ->user()
+            ->widgets()
+            ->where('dashboard_component', app(ComponentRegistry::class)->getClass($this->dashboardComponent))
+            ->where('component_name', $this->getName())
+            ->first();
```
SirSplasch, line 24: Can $widget be null? If so, handle that case.

resources/views/components/dashboard/grid.blade.php
```diff
@@ -30,6 +30,8 @@ class="w-full"
                     @livewire(
                         $widget['component_name'] ?? $widget['class'],
                         array_merge($this->getWidgetAttributes(), [
+                            'config' => data_get($widget, 'config'),
+                            'dashboardComponent' => $this->getName(),
```
SirSplasch, line 34: is dashboardComponent $this->getName()? Just want to be sure.

src/Traits/Livewire/SupportsWidgetConfig.php
```diff
@@ -0,0 +1,33 @@
...
+
+    #[Locked]
+    public ?array $config = null;
+
+    protected function storeConfig(string $path, array|bool|string|int|float|Arrayable|null $value): void
+    {
+        $widget = auth()
+            ->user()
+            ->widgets()
+            ->where('dashboard_component', app(ComponentRegistry::class)->getClass($this->dashboardComponent))
```
SirSplasch, line 22: Maximum line length exceeded

SirSplasch (changes requested):

src/Traits/Widgetable.php
patrickweh: no because the dashboardComponent could contain an array. in that case i dont know the context where the widget is currently used and configured in

resources/views/components/dashboard/grid.blade.php
patrickweh: $this->getName() is from livewire and returns the alias of the component

PR #1035, opened 2025-07-20
add Active Leads by Sales Representative widget and expand LeadTableS…

ktownmaxi:
…eeder

## Summary by Sourcery

Add a new Active Leads by Sales Representative dashboard widget and enhance lead seeding to populate additional fields.

New Features:
- Introduce Active Leads by Sales Representative widget with configurable time frames and interactive filtering
- Enable redirect from widget to a filtered leads list by sales representative

Enhancements:
- Extend LeadTableSeeder to assign random lead_state, record_origin, and recommended_by_address to seeded leads


src/Livewire/Widgets/ActiveLeadsBySalesRepresentative.php
```diff
@@ -0,0 +1,149 @@
...
+    public function show(array $params): void
+    {
+        $start = $this->getStart()->toDateString();
+        $end = $this->getEnd()->toDateString();
+
+        SessionFilter::make(
+            Livewire::new(resolve_static(LeadList::class, 'class'))->getCacheKey(),
+            fn (Builder $query) => $query
+                ->where('user_id', data_get($params, 'id'))
+                ->whereBetween('end', [$start, $end])
```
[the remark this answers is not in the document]
DarkGhostHunter: Depends on what the widget does. If you have two columns of the same type (like timestamps), `whereValueBetween($value, ['starts_at', 'ends_at'])` will work fine.

Otherwise, you will need to use `DB::raw` on one column or both columns to normalise them.

```php
$query->whereValueBetween(
    $value, [DB:raw('DATE(starts_at)'), DB::raw('DATE('ends_at')]
);
```

Depending on your indexes, this may need a full table sweep unless you limit the data with `limit()`.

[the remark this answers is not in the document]
DarkGhostHunter: Then it should be fine. I ensured that PR allowed raw statements on both values and columns.

Also, consider that is uses `BETWEEN`, which is inclusive and will return `false` if one of the column values is `null`. If you need support of half-closed comparison, you will need to do a manual `whereNull`. For example, let's imagine you need to find records where the `ends_at` is null but the `$value` is the same or greater than `started_at`:

```php
$query->where(function ($query) use ($value) {
    $query->where('starts_at', '>=', $value)
          ->whereNull('ends_at');
})->orWhereValueBetween($value, ['starts_at', 'ends_at']);
```

database/seeders/LeadTableSeeder.php
```diff
@@ -12,10 +14,15 @@ class LeadTableSeeder extends Seeder
     public function run(): void
     {
         $addresses = Address::all(['id']);
+        $leadStates = LeadState::all(['id']);
+        $leadRecordOrigins = RecordOrigin::all()->where('model_type', morph_alias(Lead::class))->get(['id']);
```
SirSplasch, line 18: wrap

SirSplasch, line 18: Do not use all()->where() as this queries the collection. Better query the database query()->where()

src/Livewire/Widgets/ActiveLeadsBySalesRepresentative.php
```diff
@@ -0,0 +1,140 @@
...
+        $this->series = $leadCounts
+            ->map(function (User $user): array {
+                return [
+                    'id' => $user->getKey(),
+                    'name' => $user->getLabel(),
+                    'color' => $user->color ?: ChartColorEnum::forKey($user->getKey())->value,
+                    'data' => [$user->total],
+                ];
+            })
+            ->take(25)
```
SirSplasch, line 87: Why limit this to the first 25 users? Am I missing something or do the later users never get shown?

src/Livewire/Widgets/ActiveLeadsBySalesRepresentative.php
```diff
@@ -0,0 +1,140 @@
...
+            fn (Builder $query) => $query
+                ->where('user_id', data_get($params, 'id'))
+                ->whereValueBetween(now(), ['start', 'end'])
+                ->whereHas(
+                    'leadState',
+                    fn (Builder $query) => $query
+                        ->where('is_won', false)
+                        ->where('is_lost', false)
+                ),
+            __('Active leads by :user', ['user' => data_get($params, 'name')]) . ' ' .
```
SirSplasch, line 133: the concatination dot at the start of the next line, not at the end of the current one.

src/Livewire/Widgets/ActiveLeadsBySalesRepresentative.php
```diff
@@ -0,0 +1,140 @@
...
+    public function show(array $params): void
+    {
+        $start = $this->getStart()->toDateString();
+        $end = $this->getEnd()->toDateString();
+
+        SessionFilter::make(
+            Livewire::new(resolve_static(LeadList::class, 'class'))->getCacheKey(),
+            fn (Builder $query) => $query
+                ->where('user_id', data_get($params, 'id'))
+                ->whereValueBetween(now(), ['start', 'end'])
```
SirSplasch, line 126: This does not take $start and $end into account and therefore does not lead to the results the user would expect.
The two timeframes must intersect.

SirSplasch (approved):

PR #1056, opened 2025-07-30
add-order-type-selection-on-child-orders

patrickweh:
## Summary by Sourcery

Enable choosing a specific order type when creating child orders and add a percentage-driven bulk inclusion of order positions

New Features:
- Add an order type selection dropdown to the child order creation form
- Introduce a percentage-based "Take All" action to include order positions proportionally

Enhancements:
- Populate available order types dynamically and auto-select when only one option is available
- Extend takeOrderPositions to compute and round amounts based on the entered percentage and reset the percentage input after use
- Adjust the save button to require both a selected order type and at least one position


resources/views/livewire/order/create-child-order.blade.php
```diff
@@ -47,6 +47,41 @@ class="text-sm font-medium text-gray-500 dark:text-gray-400"
         </div>
 
         <x-card>
+            <div class="mb-6">
+                <x-select.styled
```
SirSplasch, line 51: please keep ordering like every other x-select.styled

src/Livewire/Order/CreateChildOrder.php
```diff
@@ -125,18 +133,26 @@ public function save(): void
...
-            $this->selectedPositions = [];
-
-            return;
+        // Check if percentage is set for "take all" functionality
+        $takeAllWithPercentage = $this->percentage && $this->percentage > 0;
+
+        if ($takeAllWithPercentage) {
+            // Take all positions with percentage
+            $this->replicateOrder->order_positions = [];
+            $positionIds = [];
```
SirSplasch, line 142: no need to set positionIds if it is not used on $takeAllWithPercentage

SirSplasch (changes requested):

SirSplasch (approved):

PR #1071, opened 2025-08-06
add ChartColorEnum for color management in charts

ktownmaxi:
All colors and hex values are taken from: https://tailwindcss.com/docs/colors

Color after Slate were omitted because of low distinctiveness.

### DIFF
--- a/src/Enums/ChartColorEnum.php
+++ b/src/Enums/ChartColorEnum.php
@@ -1,0 +1,85 @@
+<?php
+
+namespace FluxErp\Enums;
+
+use FluxErp\Enums\Traits\EnumTrait;
+
+enum ChartColorEnum: string
+{
+    use EnumTrait;
+
+    public static function forIndex(int $index): self
+    {
+        return self::cases()[$index % count(self::cases())];
+    }
+
+    public static function forKey(string|int $key): self
+    {
+        return self::forIndex(crc32((string) $key));
+    }
+
+    public function hex(): string
+    {
+        return match ($this) {
+            self::Red => '#ef4444',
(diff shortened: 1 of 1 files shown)


src/Enums/ChartColorEnum.php
```diff
@@ -0,0 +1,56 @@
...
+enum ChartColorEnum: string
+{
+    use EnumTrait;
+
+    public static function forIndex(int $index): self
+    {
+        return self::cases()[$index % count(self::cases())];
+    }
+
+    public static function forKey(string|int $key): self
```
SirSplasch, line 16: What is this method used for? And why does it use the crc32() function? I would not recommend using this function due to the warnings given in the php documentation.

src/Enums/ChartColorEnum.php
SirSplasch, line 1: Why add an enum that is not used? Maybe use it in a widget so that others have an example on how it is intended to be used?

SirSplasch (changes requested):

src/Enums/ChartColorEnum.php
ktownmaxi: CRC32() is a hashing method. This method is intended to output consistent colors throughout all the widgets for the same input values. 

For example if you input the name "Leo" a deterministic color is outputted. This makes the output appear random but its identical for the same input. That means the name "Leo" has in every widget where this function is used the same color.

ktownmaxi: Do you recommend to use another hasing method or remove the method entirely?

SirSplasch (approved):

PR #1074, opened 2025-08-06
Prepare chat integration flux core

ktownmaxi:
## Summary by Sourcery

New Features:
- Automatically join the 'presence_channel' using Echo on DOMContentLoaded for authenticated web users in the main layout.

### DIFF
--- a/resources/views/components/layouts/head/head.blade.php
+++ b/resources/views/components/layouts/head/head.blade.php
@@ -29,4 +29,15 @@
 <tallstackui:script />
 {{ \FluxErp\Facades\Asset::toHtml() }}
 @stack('scripts')
+@auth('web')
+    <script>
+        document.addEventListener(
+            'livewire:navigated',
+            () => {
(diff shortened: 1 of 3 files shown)


routes/channels.php
```diff
@@ -48,3 +48,9 @@ class_to_broadcast_channel(morphed_model('user')),
 });
 
 Broadcast::channel('action.*', fn () => true);
+
+Broadcast::channel('presence', function (FluxErp\Models\User $user) {
```
patrickweh, line 52: Import the model

resources/views/components/layouts/head/head.blade.php
```diff
@@ -29,4 +29,16 @@
 <tallstackui:script />
 {{ \FluxErp\Facades\Asset::toHtml() }}
 @stack('scripts')
+@auth('web')
```
patrickweh, line 32: Move This Inside the script Stack

PR #1080, opened 2025-08-11
fix: add mutators and recalculation methods for lead model attributes

ktownmaxi:
Add mutators to automaticly calculate weighted_gross_profit and weighted_revenue every time a change to relevant values is made

## Summary by Sourcery

Add mutators to recalculate weighted revenue and gross profit whenever probability, expected revenue, or expected gross profit attributes change, and update the Lead factory to rely on these automatic calculations.

New Features:
- Automatically recalculate weighted_revenue when expected_revenue or probability_percentage is set
- Automatically recalculate weighted_gross_profit when expected_gross_profit or probability_percentage is set

Enhancements:
- Extract weighted calculation logic into recalculateWeightedRevenue and recalculateWeightedGrossProfit methods
- Remove precomputed weighted values from the Lead factory since they are now derived automatically

### DIFF
--- a/src/Models/Lead.php
+++ b/src/Models/Lead.php
@@ -259,7 +259,7 @@ class Lead extends FluxModel implements Calendarable, HasMedia, InteractsWithDat
         $revenue = data_get($this->attributes, 'expected_revenue');
 
         $this->attributes['weighted_revenue'] = ($prob !== null && $revenue !== null)
-            ? bcmul($prob, $revenue, 2)
+            ? bcmul($prob, $revenue)
             : null;
     }
 }
--- a/src/Models/Lead.php
+++ b/src/Models/Lead.php
@@ -130,8 +130,13 @@ class Lead extends FluxModel implements Calendarable, HasMedia, InteractsWithDat
                 $lead->probability_percentage = $probability;
             }
 
-            $lead->recalculateWeightedGrossProfit();
-            $lead->recalculateWeightedRevenue();
+            if ($lead->isDirty('probability_percentage') || $lead->isDirty('expected_gross_profit')) {
+                $lead->recalculateWeightedGrossProfit();
+            }
+
+            if ($lead->isDirty('probability_percentage') || $lead->isDirty('expected_revenue')) {
+                $lead->recalculateWeightedRevenue();
+            }
         });
     }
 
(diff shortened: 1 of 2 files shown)


src/Models/Lead.php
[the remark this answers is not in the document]
ktownmaxi: I used the old system because when calling Attribute::make() the cast pipeline gets triggered. This cast triggers an update of  the value which in turn calls the mutator again. The old setAttribute system doesn't trigger the cast pipline and prevents infinite loops with the money cast

src/Models/Lead.php
```diff
@@ -223,4 +242,28 @@ public function user(): BelongsTo
...
     }
+
+    protected function recalculateWeightedGrossProfit(): void
+    {
+        if (! is_null($this->probability_percentage) && ! is_null($this->expected_gross_profit)) {
+            $this->weighted_gross_profit = bcmul(
+                $this->probability_percentage,
+                $this->expected_gross_profit
+            );
+        } else {
```
SirSplasch, line 253: Why else statement, do you really want to set the weighted_gross_profit to null and therefore override the existing value?

src/Models/Lead.php
```diff
@@ -223,4 +242,28 @@ public function user(): BelongsTo
...
+    }
+
+    protected function recalculateWeightedRevenue(): void
+    {
+        if (! is_null($this->probability_percentage) && ! is_null($this->expected_revenue)) {
+            $this->weighted_revenue = bcmul(
+                $this->probability_percentage,
+                $this->expected_revenue
+            );
+        } else {
```
SirSplasch, line 265: Same here, why set to null?

src/Models/Lead.php
```diff
@@ -200,6 +200,25 @@ public function scopeInTimeframe(
...
+        $this->attributes['expected_revenue'] = $value;
+        $this->recalculateWeightedRevenue();
+    }
+
+    public function setProbabilityPercentageAttribute(float $value): void
+    {
+        $this->attributes['probability_percentage'] = $value;
+        $this->recalculateWeightedRevenue();
+        $this->recalculateWeightedGrossProfit();
+    }
```
SirSplasch, line 220: Why do you even use mutators for the goal you want to achieve? Just do the recalculations in the static::saving() function. Then those recalculation methods must be public as you otherwise can not access them.
Or do you really want to update those values on the fly even if you do not want the model to be saved?

SirSplasch (changes requested):

PR #1085, opened 2025-08-22
implement shared targets

ktownmaxi:


resources/views/livewire/settings/targets.blade.php
```diff
@@ -42,15 +42,55 @@
             multiple
             wire:model="target.users"
             select="label:label|value:id"
+            x-on:select="$wire.updateUsers()"
```
SirSplasch, line 45: I would update the user_shares array fully in js, not passing this to php. So that there is no re-rendering needed.
Then you also need x-on:remove

resources/views/livewire/settings/targets.blade.php
```diff
@@ -42,15 +42,55 @@
...
         />
+
+        <x-toggle
+            :label="__('Is Group Target')"
+            wire:model="target.is_group_target"
+        />
+
+        <div x-cloak x-show="$wire.target.is_group_target">
+            <div class="space-y-2">
+                @foreach (\FluxErp\Models\User::whereIn('id', $target->users)->get() as $user)
```
SirSplasch, line 63: This is not good. Do not query the Flux user model directly as this hinders customization.
Also I would prefer using <template x-for=""> So that no re-rendering is needed.

resources/views/livewire/settings/targets.blade.php
```diff
@@ -42,15 +42,55 @@
...
+                    <x-card class="flex items-center">
+                        <div class="flex flex-1 items-center">
+                            <img
```
SirSplasch, line 66: Do you really need to display the user avatar?

resources/views/livewire/settings/targets.blade.php
```diff
@@ -42,15 +42,55 @@
...
+                                class="h-8 w-8 rounded-full bg-gray-200 object-cover object-center"
+                                src="{{ $user->getAvatarUrl() }}"
+                            />
+                            <span class="ml-2">{{ $user->name }}</span>
+                        </div>
+
+                        <div class="flex gap-2" style="flex: 0 0 55%">
```
SirSplasch, line 73: I would add a toggle for whether the share is absolute or a percentage value.
And then disable or set readonly the other value, so that the user can only modify either one. (If you want to keep showing the user both values)

resources/views/livewire/settings/targets.blade.php
```diff
@@ -42,15 +42,55 @@
...
+                            <x-number
+                                class="flex-1"
+                                :label="__('Relative Target Share (%)')"
+                                step="1"
+                                min="0"
+                                max="100"
+                                wire:model.defer="target.user_shares.{{ $user->id }}.relative"
```
SirSplasch, line 80: no need to ever set .defer as this is the default behaviour

src/Livewire/Forms/TargetForm.php
```diff
@@ -38,15 +40,49 @@ class TargetForm extends FluxForm
...
-    public ?array $users = null;
+    public array $user_shares = [];
+
+    public array $users = [];
 
     public function fill($values): void
     {
         if ($values instanceof Target) {
-            $values->loadMissing('users:id');
+            $model = $values->loadMissing('users');
```
SirSplasch, line 50: Why did you change this? You create an exact copy of $values and you are selecting all of the users columns but you only need the id.

src/Livewire/Forms/TargetForm.php
```diff
@@ -38,15 +40,49 @@ class TargetForm extends FluxForm
...
+
+            $arr = $model->toArray();
+            $arr['users'] = $model->users->pluck('id')->all();
+            $arr['user_shares'] = $userShares;
+
+            parent::fill($arr);
 
-            $values = $values->toArray();
-            $values['users'] = array_column($values['users'] ?? [], 'id');
+            return;
```
SirSplasch, line 85: There was no reason to change this, just set the user_shares and you're good.

src/Livewire/Forms/TargetForm.php
```diff
@@ -38,15 +40,49 @@ class TargetForm extends FluxForm
...
     public function fill($values): void
     {
         if ($values instanceof Target) {
-            $values->loadMissing('users:id');
+            $model = $values->loadMissing('users');
+
+            $userShares = [];
+            $targetValue = ($model->target_value ?? 0);
+
+            foreach ($model->users as $user) {
```
SirSplasch, line 55: no need to loop through this. Something along those lines is sufficient.

            $values = $values->toArray();
            $values['user_shares'] = array_map(
                fn (array $user) => [
                    'id' => data_get($user, 'id'),
                    'target_share' => data_get($user, 'pivot.target_share'),
                ],
                $values['users'] ?? []
            );
            $values['users'] = array_column($values['users'] ?? [], 'id');

src/Livewire/Settings/Targets.php
```diff
@@ -92,6 +94,23 @@ public function updateSelectableColumns(?string $modelType = null): void
         $this->forceRender();
     }
 
+    public function updateUsers(): void
```
SirSplasch, line 97: Can be removed if all of this is done in js.

src/Actions/Target/CreateTarget.php
SirSplasch, line 1: I would not validate if sum of the absolutes = target_value or sum of the percentages = 100%. As why shouldn't it be possible to set the target for a group as an "overachieved" target.

src/Actions/Target/CreateTarget.php
```diff
@@ -22,16 +22,29 @@ protected function getRulesets(): string|array
...
+            }
+
+            $target->users()->attach($pivotData);
 
-            foreach ($target->children()->get('id') as $child) {
-                $child->users()->attach($users);
+            foreach ($target->children()->get() as $child) {
+                $child->users()->attach($pivotData);
             }
         }
```
SirSplasch, line 49: No need to adjust this in any shape or form if the $users array is set up properly. Also a mutator on the target_share might be needed that a percentage value of 0.5 => 50% and 50 => 0.5

E.g: 

        $users = [
            [
                'id' => 1,
                'target_share' => 100,
                'is_percentage' => true,
            ],
            [
                'id' => 2,
                'target_share' => 12,
                'is_percentage' => false,
            ],
        ];

SirSplasch (approved):

PR #1088, opened 2025-08-26
fix: fix x-axis formatting

ktownmaxi:
## Summary by Sourcery

Bug Fixes:
- Bypass JS Date formatting for four-digit year values in xAxisFormatter to display raw year labels


src/Traits/HasTemporalXAxisFormatter.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+namespace FluxErp\Traits;
```
patrickweh, line 3: Should be in traits/Livewire

patrickweh (changes requested):

PR #1090, opened 2025-08-27
added end state property to all endable States

ktownmaxi:


src/States/EndableState.php
SirSplasch, line 1: Why did you use abstract class instead of a trait? I just want to hear your train of thought.

SirSplasch (approved):

src/States/EndableState.php
ktownmaxi: Abstract classes are used to build hierarchies. If I used a trait instead, I’d lose the benefits of a unified hierarchy and wouldn’t be able to enforce structure as strictly as with an abstract class. Traits are great for mixing in utilities or behaviors (code reuse that is not bound by the actual class), but here I wanted an enforceable relationship between the classes.

PR #1098, opened 2025-09-04
fix margin calculation

ktownmaxi:
## Summary by Sourcery

Bug Fixes:
- Fix margin calculation to sum the ‘margin’ field of non-alternative order positions instead of using net price minus total purchase price

### DIFF
--- a/src/Models/Order.php
+++ b/src/Models/Order.php
@@ -414,13 +414,10 @@ class Order extends FluxModel implements HasMedia, InteractsWithDataTables, Offe
 
     public function calculateMargin(): static
     {
-        $this->total_purchase_price = $this->orderPositions()
-            ->where('is_alternative', false)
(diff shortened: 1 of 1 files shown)


src/Models/Order.php
```diff
@@ -414,13 +414,10 @@ function (string|float|int $previous, Discount $discount): string|float|int {
 
     public function calculateMargin(): static
     {
-        $this->total_purchase_price = $this->orderPositions()
-            ->where('is_alternative', false)
-            ->sum('purchase_price');
-
         $this->margin = Rounding::round(
```
patrickweh, line 417: You Removed the 2 for the rounding

patrickweh (changes requested):

patrickweh (approved):

PR #1112, opened 2025-09-10
add order transaction modal and refactor accounting logic

patrickweh:
## Summary by Sourcery

Add a modal-based flow for creating order transactions and refactor the accounting Livewire component to use separate form objects, streamline save logic, and improve data handling and sorting

New Features:
- Introduce a modal dialogue for adding order transactions directly from the order accounting page
- Add an “Add” button in the transactions table to launch the new order-transaction modal
- Implement DataTableHasFormEdit support with a custom modelKeyName for inline editing in data tables

Enhancements:
- Refactor the Accounting component to extend OrderTransactionList, split OrderTransactionForm and TransactionForm, and adjust the saveTransaction workflow
- Update query builders to filter by order_id, eager load transaction relations, and include bank connections in view data
- Add a currency relationship to Transaction models and include currency ISO code in table output
- Adjust the Order model boot logic to recalculate balance only when an invoice_number is present
- Sort transactions by booking_date in the TransactionAssignments component

### DIFF
--- a/src/Livewire/Order/Accounting.php
+++ b/src/Livewire/Order/Accounting.php
@@ -79,16 +79,6 @@ class Accounting extends OrderTransactionList
         }
     }
 
-    public function getReturnKeys(): array
-    {
-        return array_merge(
-            parent::getReturnKeys(),
-            [
-                'currency_iso',
-            ]
-        );
-    }
-
     #[Renderless]
     public function saveTransaction(): bool
     {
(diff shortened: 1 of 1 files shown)


src/Livewire/Order/Accounting.php
```diff
@@ -3,86 +3,157 @@
...
+        $this->js(<<<'JS'
+            $modalOpen('transaction-details-modal');
+        JS);
+
         if (! $transaction) {
             $this->transactionForm->amount = $this->order->balance;
         }
     }
 
+    public function getReturnKeys(): array
```
SirSplasch, line 82: protected

SirSplasch (changes requested):

PR #1115, opened 2025-09-12
add closed_at column to leads and update related queries

ktownmaxi:
## Summary by Sourcery

Introduce a closed_at timestamp on leads to record when they are won or lost, auto-populate it on state changes, and update related dashboard queries to filter by closed_at instead of created_at

New Features:
- Add nullable closed_at column to the leads table via migration
- Automatically set or clear the closed_at timestamp when a lead’s state changes to won or lost

Enhancements:
- Modify Livewire widgets to use closed_at for filtering and aggregating lead won/lost charts and ratios

Build:
- Add database migration to add and drop the closed_at column

### DIFF
--- a/src/Models/Lead.php
+++ b/src/Models/Lead.php
@@ -157,13 +157,6 @@ class Lead extends FluxModel implements Calendarable, HasMedia, InteractsWithDat
             if ($lead->isDirty('probability_percentage') || $lead->isDirty('expected_revenue')) {
                 $lead->recalculateWeightedRevenue();
             }
-
-            if ($lead->isDirty('lead_state_id')) {
-                $isClosed = $lead->leadState
-                    && ($lead->leadState->is_won || $lead->leadState->is_lost);
-
-                $lead->closed_at = $isClosed ? now() : null;
-            }
         });
     }
 


database/migrations/2025_09_12_062633_add_closed_at_column_to_leads_table.php
```diff
@@ -0,0 +1,22 @@
...
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class() extends Migration
+{
+    public function up(): void
+    {
+        Schema::table('leads', function (Blueprint $table): void {
+            $table->timestamp('closed_at')->nullable()->after('weighted_revenue');
```
SirSplasch, line 12: I'm curious whether a closed_by column could be useful as well.

src/Models/Lead.php
```diff
@@ -138,6 +139,13 @@ protected static function booted(): void
             if ($lead->isDirty('probability_percentage') || $lead->isDirty('expected_revenue')) {
                 $lead->recalculateWeightedRevenue();
             }
+
+            if ($lead->isDirty('lead_state_id')) {
+                $isClosed = $lead->leadState
```
SirSplasch, line 144: $isClosed = $lead->leadState()->where()->exists();

src/Models/Lead.php
```diff
@@ -138,6 +139,13 @@ protected static function booted(): void
...
+                    && ($lead->leadState->is_won || $lead->leadState->is_lost);
+
+                $lead->closed_at = $isClosed ? now() : null;
```
SirSplasch, line 147: If $lead->closed_at is not null beforehand and $isClosed = true, is it intended to update the value to a (most likely) newer timestamp?

SirSplasch (changes requested):

database/migrations/2025_09_12_062633_add_closed_at_column_to_leads_table.php
ktownmaxi: I like that. I'll find out if a closed_by column is something we'd use.

PR #1117, opened 2025-09-12
fix search in permissions

ktownmaxi:
## Summary by Sourcery

Bug Fixes:
- Use Permission::search with a query callback to exclude the ‘address’ guard and restore proper pagination for permissions lookup

### DIFF
--- a/src/Livewire/Settings/UserEdit.php
+++ b/src/Livewire/Settings/UserEdit.php
@@ -62,7 +62,6 @@ class UserEdit extends Component
                 )
                     ->query(fn ($query) => $query->where('guard_name', '!=', 'address'))
                     ->paginate(pageName: 'permissionsPage'),
-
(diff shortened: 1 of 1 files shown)


src/Livewire/Settings/UserEdit.php
```diff
@@ -53,10 +53,9 @@ public function render(): View|Factory|Application
     {
         return view('flux::livewire.settings.user-edit',
             [
-                'permissions' => resolve_static(Permission::class, 'query')
-                    ->where('guard_name', '!=', 'address')
-                    ->when($this->searchPermission, fn ($query) => $query->search($this->searchPermission))
-                    ->paginate(pageName: 'permissionsPage'),
+                'permissions' => Permission::search($this->searchPermission)
```
SirSplasch, line 56: use resolve_static() to allow customization

SirSplasch (changes requested):

PR #1126, opened 2025-09-25
Add "Assign to me" functionality and update task form with authenticated user

patrickweh:
## Summary by Sourcery

Implement self-assignment support by adding an "Assign to me" action on tickets and pre-populating task forms with the current user, while streamlining default assignment logic and updating relevant tests.

New Features:
- Add "Assign to me" button in the ticket UI for users to self-assign tickets
- Automatically prefill the task creation form with the authenticated user as responsible and in the user list

Enhancements:
- Consolidate default user assignment logic into TaskForm reset and remove redundant code from TaskList
- Update binding directive in ticket component from wire:model.live to wire:model

Tests:
- Expand ProjectTaskList test to cover new task creation with default authenticated user assignment and verify database persistence

### DIFF
--- a/resources/views/livewire/ticket/ticket.blade.php
+++ b/resources/views/livewire/ticket/ticket.blade.php
@@ -232,7 +232,8 @@
                             ]"
                         />
                         <x-button
-                            x-bind:class="$wire.ticket.users?.includes({{ auth()->id() }}) && 'hidden'"
+                            x-cloak
+                            x-show="! $wire.ticket.users?.includes({{ auth()->id() }})"
                             class="w-full"
                             color="secondary"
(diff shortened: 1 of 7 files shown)


resources/views/livewire/ticket/ticket.blade.php
```diff
@@ -231,6 +231,13 @@ class="flex items-center justify-between border-b px-4 py-2.5 dark:border-0"
                                 ],
                             ]"
                         />
+                        <x-button
```
SirSplasch, line 234: Maybe hide button if already assigned to oneself?

SirSplasch (changes requested):

PR #1143, opened 2025-10-08
Enhance communication handling by adding mail address retrieval and improving form interactions

patrickweh:


src/Livewire/Contact/Communication.php
```diff
@@ -15,4 +17,48 @@ public function getCacheKey(): string
...
+                            ->where('type', 'email')
+                            ->whereNotNull('value'),
+                    ]),
+            ])
+            ->first([
+                'id',
+                'main_address_id',
+            ])
+            ->mainAddress;
+
```
SirSplasch, line 40: remove line to keep it consistent with every other getMailAddress() and getPostalAddress() method, or add this new line to every other method.

SirSplasch (changes requested):

SirSplasch (changes requested):
Add null-safety checks to ever getMailAddress and getPostalAddress that does not query the address directly

PR #1408, opened 2026-02-13
Prefer order VAT rate over product default on position changes

patrickweh:
## Summary
- When changing the product on an order position, the order-level VAT rate now takes precedence over the product's default VAT rate
- Allow explicit manual VAT rate overrides on position creation even when the order has a VAT rate set
- Strip `vat_rate_id` from replicated position data so duplicated orders correctly inherit the new order's VAT rate

## Summary by Sourcery

Align order position VAT handling with order-level VAT configuration and support explicit overrides.

New Features:
- Allow explicit VAT rate overrides when creating order positions even if the parent order has a VAT rate set.

Bug Fixes:
- Ensure changing the product on an order position preserves the order-level VAT rate instead of reverting to the product's default VAT rate.
- Ensure replicated orders correctly inherit the new order's VAT rate on their positions instead of carrying over the original position VAT rate.

Tests:
- Add feature tests covering product change VAT precedence and explicit VAT override behavior for order positions.

### DIFF
--- a/tests/Feature/Actions/OrderPosition/UpdateOrderPositionTest.php
+++ b/tests/Feature/Actions/OrderPosition/UpdateOrderPositionTest.php
@@ -1,0 +1,176 @@
+<?php
+
+use FluxErp\Actions\OrderPosition\CreateOrderPosition;
+use FluxErp\Actions\OrderPosition\UpdateOrderPosition;
+use FluxErp\Enums\OrderTypeEnum;
+use FluxErp\Models\Address;
+use FluxErp\Models\Contact;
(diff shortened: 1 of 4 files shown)


tests/Feature/Actions/OrderPosition/UpdateOrderPositionTest.php
SirSplasch: test() instead of it(). Maybe you can go through all tests to fix this, if you want to, but no need in this specific pull request.

SirSplasch (changes requested):

PR #1495, opened 2026-03-20
Add execution preview to schedule modal

patrickweh:
## Summary

- Adds a reactive execution date preview to the schedule edit modal showing the next 5 planned execution dates
- Renames "Due At" / "Fällig am" to "Next Execution" / "Nächste Ausführung" for clarity
- Shows a warning when the next execution date is in the past (will execute immediately)
- Makes the Repeat frequency select searchable
- Preview updates reactively when any schedule parameter changes (cron config, due_at, ends_at, recurrences, end mode radio)
- Respects the active end mode radio (never/ends_at/recurrences) for preview calculation
- Includes due_at itself as the first execution date when it is in the future
- All changes applied to both subscription and purchase-subscription views

## Summary by Sourcery

Add a reactive execution preview and improved scheduling UX to the subscription scheduling modals.

New Features:
- Display the next planned execution dates for a schedule based on the selected frequency and end conditions.
- Show a warning in the schedule modal when the next execution date is in the past, indicating immediate execution.
- Make the repeat frequency selector searchable in the schedule configuration.
- Rename the schedule date label from "Due At" to "Next Execution" in subscription and purchase-subscription views.

Enhancements:
- Recalculate and update the execution preview whenever schedule parameters change, including cron settings, start date, end mode, end date, and recurrences.


src/Livewire/Forms/ScheduleForm.php
```diff
@@ -131,6 +141,115 @@ public function save(): void
...
+                    if ($endsAt && $next->greaterThan($endsAt)) {
+                        break;
+                    }
+
+                    $dates[] = $next->toDateTimeString();
+                    $current = $next;
+                }
+            }
+        } catch (Throwable) {
+            return [];
```
SirSplasch: Why not return $dates?

SirSplasch (changes requested):

src/Livewire/Forms/ScheduleForm.php
```diff
@@ -131,6 +141,116 @@ public function save(): void
...
+        $endsAt = $this->end_radio === 'ends_at' && $this->ends_at
+            ? Carbon::parse($this->ends_at)
+            : null;
+        $remainingRecurrences = $this->end_radio === 'recurrences' && $this->recurrences
+            ? max(0, $this->recurrences - ($this->current_recurrence ?? 0))
+            : null;
+
+        if ($remainingRecurrences === 0) {
+            return [];
+        }
```
SirSplasch: Move this up, directly below the first exit condition, as this is the second one and needs nothing of the statements that are currently above it.

src/Livewire/Forms/ScheduleForm.php
```diff
@@ -131,6 +141,116 @@ public function save(): void
...
+            }
+        }
+
+        $maxDates = $remainingRecurrences !== null
+            ? min($count, $remainingRecurrences - count($dates))
+            : $count;
+
+        if ($maxDates <= 0) {
+            return $dates;
+        }
```
SirSplasch: Move everything up that is needed for this as the third exit condition. But only what's needed, no something like $from.

src/Livewire/Forms/ScheduleForm.php
```diff
@@ -131,6 +141,116 @@ public function save(): void
...
+        );
+
+        $schedule = app(Schedule::class);
+        $event = $schedule->call(fn () => null);
+
+        if ($parameters !== []) {
+            $event = $event->{$method}(...$parameters);
+        } else {
+            $event = $event->{$method}();
+        }
```
SirSplasch, line 209: All of this is only needed right before the try-catch, so move it accordingly. No need to do this this early if there are possible exit conditions that can be met without this.

SirSplasch (changes requested):
Also resolve merge conflicts.

SirSplasch (approved):

PR #1567, opened 2026-04-10
Add SalutationEnum cast to Address model

patrickweh:
## Summary
- Add missing SalutationEnum cast to Address model salutation column
- Enables automatic translation of salutation values in DataTables
- Enables select dropdown filter instead of free text for salutation column
- Depends on Team-Nifty-GmbH/tall-datatables#210

## Summary by Sourcery

Enhancements:
- Cast the Address model salutation attribute to SalutationEnum for consistent typed access and downstream integrations.

### DIFF
--- a/src/Models/Address.php
+++ b/src/Models/Address.php
@@ -342,6 +342,7 @@ class Address extends FluxAuthenticatable implements Calendarable, HasLocalePref
     {
         return [
             'date_of_birth' => 'date',
+            'salutation' => SalutationEnum::class,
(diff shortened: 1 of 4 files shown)


src/Models/Address.php
```diff
@@ -342,6 +342,7 @@ protected function casts(): array
     {
         return [
```
SirSplasch, line 343: order keys according to database table

SirSplasch (changes requested):

SirSplasch (approved):

PR #1574, opened 2026-04-13
Fix email attachments using cached documents when force is active

patrickweh:
## Summary
- When a print layout has force-recreate enabled (like balance-statement), `collectMailMessages()` now generates a fresh attachment instead of reusing the cached media from the model
- Previously, the email was assembled with the old PDF before `CreateDocumentsJob` regenerated it, causing stale documents to be sent via email

## Summary by Sourcery

Ensure email attachments use freshly generated documents when force-recreate layouts are selected instead of cached media.

Bug Fixes:
- Prevent sending stale PDF attachments by regenerating documents for force-enabled print layouts during email collection.

Tests:
- Add a feature test verifying that email attachments for force-enabled layouts are generated fresh and not taken from cached media.

### DIFF
--- a/tests/Feature/Traits/CreatesDocumentsForceEmailTest.php
+++ b/tests/Feature/Traits/CreatesDocumentsForceEmailTest.php
@@ -5,7 +5,7 @@ use FluxErp\Models\Media;
 use FluxErp\Traits\Livewire\CreatesDocuments;
 use Livewire\Component;
 
-it('regenerates email attachment when force is active instead of using cached media', function (): void {
+test('regenerates email attachment when force is active instead of using cached media', function (): void {
     $contact = Contact::factory()->create();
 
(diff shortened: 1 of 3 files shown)


tests/Feature/Traits/CreatesDocumentsForceEmailTest.php
```diff
@@ -0,0 +1,76 @@
+<?php
+
+use FluxErp\Models\Contact;
+use FluxErp\Models\Media;
+use FluxErp\Traits\Livewire\CreatesDocuments;
+use Livewire\Component;
+
+it('regenerates email attachment when force is active instead of using cached media', function (): void {
```
SirSplasch: test()

SirSplasch (changes requested):

PR #1581, opened 2026-04-14
Fix categories DataTable showing no data

patrickweh:
## Summary
- `getBuilder()` was replacing the DataTable's configured builder with a new query from `familyTree()`, losing select/join/sort configuration
- `getResultFromQuery()` returned a flat array instead of the paginated format the DataTable frontend expects
- Formatter was not applied to child categories

Fix: use filtered query IDs to load the tree via `familyTree()`, run each model through `itemToArray()` for proper formatting, wrap result in expected format

Depends on Team-Nifty-GmbH/tall-datatables#219 for leftAppend rendering

## Summary by Sourcery

Fix category DataTable to use the existing filtered query as the root set and return results in the paginated format expected by the frontend.

Bug Fixes:
- Preserve the DataTable's configured query builder while restricting categories to root items.
- Ensure category trees are loaded via the familyTree relationship while still respecting user-applied filters.
- Apply the standard item-to-array formatting to all categories, including nested children, and return results with the expected data/total structure.

### DIFF
--- a/src/Livewire/DataTables/CategoryList.php
+++ b/src/Livewire/DataTables/CategoryList.php
@@ -5,6 +5,7 @@ namespace FluxErp\Livewire\DataTables;
 use FluxErp\Models\Category;
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Support\Arr;
+use Illuminate\Support\Collection;
 
 class CategoryList extends BaseDataTable
 {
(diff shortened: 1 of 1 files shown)


src/Livewire/DataTables/CategoryList.php
```diff
@@ -31,23 +30,55 @@ protected function getLeftAppends(): array
...
+            ->get();
 
-        foreach ($tree as &$item) {
-            $item = Arr::only(Arr::dot($item), $returnKeys);
-            $item['indentation'] = '';
+        $tree = to_flat_tree($categories->toArray());
+
+        // Collect ALL models recursively (parents + all nested children)
+        $allModels = collect();
+        $collectRecursive = function ($items) use (&$collectRecursive, &$allModels): void {
```
SirSplasch: Why define this as an closure and not as a normal method that can be called (can be protected)?

src/Livewire/DataTables/CategoryList.php
```diff
@@ -31,23 +30,55 @@ protected function getLeftAppends(): array
...
+                $allModels->push($item);
+
+                if ($item->relationLoaded('children')) {
+                    $collectRecursive($item->children);
+                }
+            }
+        };
+
+        $collectRecursive($categories);
+        $modelsById = $allModels->keyBy(fn ($m) => $m->getKey());
```
SirSplasch: no need to make the keyBy(), you can just create the collection or array with the corresponding keys.

SirSplasch (changes requested):

src/Livewire/DataTables/CategoryList.php
patrickweh: Extracted to `collectModelsById()` protected method.

src/Livewire/DataTables/CategoryList.php
patrickweh: Removed — the method now builds the keyed array directly during recursion.

SirSplasch (approved):

PR #1583, opened 2026-04-14
Add Spatie MailSettings to override default mail configuration

patrickweh:
## Summary
- Adds a `MailSettings` Spatie Settings class with nullable properties for all MAIL_* config values (mailer, host, port, username, password, encryption, from_address, from_name)
- Non-null settings values override `config('mail.*')` in the ServiceProvider boot phase
- Includes settings migration, Livewire form with auto-render, and settings component accessible under General settings group

## Summary by Sourcery

Introduce configurable mail settings that can override default Laravel mail configuration via Spatie settings and expose them through the settings UI.

New Features:
- Add a MailSettings Spatie settings class for configurable mail transport and sender options.
- Expose a MailSettings Livewire settings page and form under the General settings group for managing mail configuration.

Enhancements:
- Apply stored MailSettings at application boot to override mail configuration values when present.
- Register the new mail settings page in the frontend routes and menu structure.

### DIFF
--- a/src/FluxServiceProvider.php
+++ b/src/FluxServiceProvider.php
@@ -320,32 +320,37 @@ class FluxServiceProvider extends ServiceProvider
     {
         try {
             $settings = app(Settings\MailSettings::class);
+            $previousConfig = [];
 
             if (! is_null($settings->mailer)) {
+                $previousConfig['mail.default'] = config('mail.default');
                 config()->set('mail.default', $settings->mailer);
             }
 
             $defaultMailer = config('mail.default');
 
-            $mailerMap = [
+            $configMap = [
                 'host' => "mail.mailers.{$defaultMailer}.host",
                 'port' => "mail.mailers.{$defaultMailer}.port",
                 'username' => "mail.mailers.{$defaultMailer}.username",
                 'password' => "mail.mailers.{$defaultMailer}.password",
                 'encryption' => "mail.mailers.{$defaultMailer}.encryption",
-            ];
-
-            $globalMap = [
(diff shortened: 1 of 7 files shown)


src/FluxServiceProvider.php
```diff
@@ -313,4 +315,37 @@ protected function updateAuthUserProvider(): void
...
+                'password' => "mail.mailers.{$defaultMailer}.password",
+                'encryption' => "mail.mailers.{$defaultMailer}.encryption",
+            ];
+
+            $globalMap = [
+                'from_address' => 'mail.from.address',
+                'from_name' => 'mail.from.name',
+            ];
+
+            foreach (array_merge($mailerMap, $globalMap) as $property => $configKey) {
```
SirSplasch: Why do you even need to create $mailerMap and $globalMap as two separate variables, if their only purpose is to get merged together.

src/FluxServiceProvider.php
```diff
@@ -313,4 +315,37 @@ protected function updateAuthUserProvider(): void
...
+                if (! is_null($settings->{$property})) {
+                    config()->set($configKey, $settings->{$property});
+                }
+            }
+        } catch (Throwable) {
```
SirSplasch: If anything throws an exception, shouldn't the previous state recovered? As all changes to the mail config should be reverted if anything fails?

SirSplasch: Also consider reporting the exception as this silently suppresses potential issues.

SirSplasch (changes requested):

SirSplasch (approved):

PR #1586, opened 2026-04-14
Fix calendar week display: unknown option and title accumulation

patrickweh:
## Summary
- Exclude `calendarWeekAbbreviation` from FullCalendar config spread to prevent `Unknown option` console warning
- Use a separate `<span>` element for the CW suffix instead of modifying `textContent` directly, preventing title accumulation when navigating between months (e.g. "April 2026Mai 2026Juni 2026...")

## Summary by Sourcery

Fix calendar week display and configuration handling in the calendar component.

Bug Fixes:
- Prevent FullCalendar from receiving the unsupported calendarWeekAbbreviation option to avoid console warnings.
- Ensure the calendar title’s calendar-week suffix is rendered in a dedicated span element to prevent duplicated or accumulated month titles when navigating.

### DIFF
--- a/resources/js/components/calendar.js
+++ b/resources/js/components/calendar.js
@@ -594,8 +594,12 @@ const calendar = () => {
                 },
             };
 
-            const { activeCalendars, showCalendars, calendarWeekAbbreviation, ...filteredConfig } =
-                this.config;
+            const {
+                activeCalendars,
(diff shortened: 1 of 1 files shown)


SirSplasch (approved):
run prettier, so that the tests run

PR #1597, opened 2026-04-15
Fix overtime reduction (Überstundenabbau) not deducting from overtime…

Slupi:
… balance.

The CloseEmployeeDay function was incorrectly adding overtime_used back to totalOvertime. This neutralized the deficit from not working, treating it the same as sick leave or vacation. As a result, taking a day of Überstundenabbau had no effect on the overtime balance.

## Summary by Sourcery

Correct overtime balance calculation so that overtime reduction days decrease the employee's overtime instead of being neutralized.

Bug Fixes:
- Stop adding used overtime back into the daily overtime balance when closing an employee day, ensuring Überstundenabbau properly reduces total overtime.

Tests:
- Update employee day closing tests to assert that overtime reduction days and partial overtime deductions result in a negative overtime balance equal to the missed work hours.

### DIFF
--- a/src/Actions/EmployeeDay/CloseEmployeeDay.php
+++ b/src/Actions/EmployeeDay/CloseEmployeeDay.php
@@ -151,7 +151,7 @@ class CloseEmployeeDay extends FluxAction
         }
 
         // ((sick hours + vacation hours + absence hours) - target hours) + actual hours = overtime
-        $totalOvertime = bcadd(
+        $plusMinusOvertimeHours = bcadd(
             bcsub(
                 bcadd(
                     $data['sick_hours_used'],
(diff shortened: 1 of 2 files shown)


src/Actions/EmployeeDay/CloseEmployeeDay.php
```diff
@@ -174,7 +170,7 @@ function (?string $carry, AbsenceRequest $item): ?string {
             $data
         );
 
-        $plusMinusOvertimeHours = bcadd($totalOvertime, Arr::pull($data, 'overtime_used'));
+        $plusMinusOvertimeHours = $totalOvertime;
```
SirSplasch: This can be removed, just rename the $totalOvertime variable to $plusMinusOvertimeHours

SirSplasch (changes requested):

src/Actions/EmployeeDay/CloseEmployeeDay.php
Slupi: done

SirSplasch (approved):

PR #1603, opened 2026-04-16
Add Accept-Language header support and API locale fallback

patrickweh:
## Summary
- Extend Localization middleware with Accept-Language header parsing and base-language matching (e.g. `de-DE` → `de`), cached via `Cache::memo()`
- Add locale-based fallback in `HasAttributeTranslations` for API contexts where no session exists, with cached language ID lookups
- Fix `getAttributeTranslation` return type to `?string`

## Summary by Sourcery

Add header-based locale resolution and locale-based translation fallback for models when no session language is set.

New Features:
- Resolve application locale from content-language or accept-language headers, including base-language matching and user/default fallbacks.
- Allow models using HasAttributeTranslations to infer the active language from the application locale when no session language ID is available.

Bug Fixes:
- Adjust getAttributeTranslation to return null when no translation exists instead of always returning a string.

Enhancements:
- Cache language code and ID lookups to avoid repeated database queries during localization.
- Ensure number and date localization consistently follow the resolved application locale.

Tests:
- Add unit tests for HasAttributeTranslations locale fallback behavior without a session language.
- Add feature tests for Localization middleware covering header precedence, base-language matching, default fallback, and middleware chaining.

### DIFF
--- a/src/Traits/Model/HasAttributeTranslations.php
+++ b/src/Traits/Model/HasAttributeTranslations.php
@@ -25,33 +25,6 @@ trait HasAttributeTranslations
         return app(static::class)->translatableAttributes();
     }
 
-    protected static function resolveLanguageId(): ?int
-    {
-        $languageId = Session::get('selectedLanguageId');
-        $default = resolve_static(Language::class, 'default');
-
-        if (is_null($languageId)) {
-            $locale = app()->getLocale();
-
-            if ($locale === $default?->language_code) {
-                return null;
-            }
-
-            $languageId = Cache::memo()->rememberForever(
-                'language_id_' . $locale,
-                fn () => resolve_static(Language::class, 'query')
-                    ->where('language_code', $locale)
-                    ->value('id')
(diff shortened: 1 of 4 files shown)


src/Traits/Model/HasAttributeTranslations.php
```diff
@@ -24,27 +25,49 @@ public static function getTranslatableAttributes(): array
         return app(static::class)->translatableAttributes();
     }
 
+    protected static function resolveLanguageId(): ?int
```
SirSplasch: after boot method

SirSplasch (changes requested):

src/Traits/Model/HasAttributeTranslations.php
patrickweh: Fixed in b3194384d — moved resolveLanguageId() after bootHasAttributeTranslations().

SirSplasch (approved):

PR #1609, opened 2026-04-17
Bypass global scopes when resolving selected records in search controller

patrickweh:
## Summary
- Async selects fetch the initially selected record via the search controller, but global scopes (SoftDeletes, etc.) prevented soft-deleted or scope-filtered records from being resolved — showing an empty select despite a stored value
- Added `withoutGlobalScopes()` and early return for the `selected` branch so stored values are always resolvable
- Extracted `formatAndDispatch()` to deduplicate the InteractsWithDataTables mapping logic

## Summary by Sourcery

Ensure the search controller can resolve and return selected records regardless of global scopes while centralizing result formatting and dispatching logic.

Bug Fixes:
- Allow async select requests to resolve and return soft-deleted or globally-scoped-out records when they are provided as selected values.

Enhancements:
- Bypass global scopes when resolving selected records in the search controller and return early once selected records are loaded.
- Extract and reuse a dedicated formatter method for InteractsWithDataTables search results before event dispatch.

Tests:
- Add a feature test verifying that the search controller returns a soft-deleted record when it is passed as a selected value.

### DIFF
--- a/src/Http/Controllers/SearchController.php
+++ b/src/Http/Controllers/SearchController.php
@@ -38,7 +38,7 @@ class SearchController extends Controller
             $optionValue = $request->input('option-value') ?: (app($model))->getKeyName();
 
             $query = resolve_static($model, 'query')
-                ->withoutGlobalScopes();
+                ->withoutGlobalScopes([SoftDeletingScope::class]);
 
             is_array($selected)
                 ? $query->whereIn($optionValue, Arr::wrap($selected))
--- a/src/Http/Controllers/SearchController.php
+++ b/src/Http/Controllers/SearchController.php
@@ -37,10 +37,14 @@ class SearchController extends Controller
             $selected = $request->input('selected');
             $optionValue = $request->input('option-value') ?: (app($model))->getKeyName();
 
-            $query = resolve_static($model, 'query');
+            $query = resolve_static($model, 'query')
+                ->withoutGlobalScopes();
+
             is_array($selected)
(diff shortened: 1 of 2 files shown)


src/Http/Controllers/SearchController.php
```diff
@@ -37,10 +37,14 @@ class_uses_recursive(resolve_static($model, 'class'))
...
 
-            $query = resolve_static($model, 'query');
+            $query = resolve_static($model, 'query')
+                ->withoutGlobalScopes([SoftDeletingScope::class]);
+
             is_array($selected)
                 ? $query->whereIn($optionValue, Arr::wrap($selected))
                 : $query->where($optionValue, $selected);
+
+            return $this->formatAndDispatch($query->get(), $model, $request);
```
SirSplasch, line 48: How does this work with multi-select? Does this even work with multi-select?

SirSplasch (changes requested):

src/Http/Controllers/SearchController.php
patrickweh: Works via `is_array($selected)` → `whereIn` for arrays, `where` for single values (line 43-45).

TallStackUI sends `selected: [id1, id2, ...]` as array in POST body during value resolution. The early return avoids the default `limit(10)` that the original code would have applied, which actually fixes multi-select for >10 items.

Added a dedicated multi-select test (active + soft-deleted record) in 937b954.

SirSplasch (approved):

PR #1639, opened 2026-04-24
Fix null attribute_changes crash in Activities component

patrickweh:
## Summary

- Adds null-safe operator to `attribute_changes` access in `Activities::loadData()` - the column is nullable but `->toArray()` was called without null check
- Adds type hints to closure parameters in the same method

Reported from production (sbm-verlag.de) where an activity log entry had `attribute_changes = NULL`.

## Summary by Sourcery

Handle activities with nullable attribute_changes and strengthen type safety in the Activities component while adding regression coverage.

Bug Fixes:
- Prevent a crash in the Activities Livewire component when an activity log entry has a null attribute_changes column.

Enhancements:
- Add explicit Activity model and closure parameter type hints in the Activities component for better type safety.

Tests:
- Add a Livewire test ensuring Activities handles activity_log entries with null attribute_changes without errors.

### DIFF
--- a/src/Livewire/Support/Activities.php
+++ b/src/Livewire/Support/Activities.php
@@ -12,6 +12,7 @@ use Livewire\Attributes\Modelable;
 use Livewire\Attributes\Renderless;
 use Livewire\Component;
 use Livewire\WithPagination;
+use Spatie\Activitylog\Models\Activity;
 use TeamNiftyGmbH\DataTable\Helpers\Icon;
 
 abstract class Activities extends Component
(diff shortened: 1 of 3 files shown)


src/Livewire/Support/Activities.php
```diff
@@ -12,6 +12,7 @@
 use Livewire\Attributes\Renderless;
 use Livewire\Component;
 use Livewire\WithPagination;
+use Spatie\Activitylog\Models\Activity;
```
SirSplasch, line 15: Are you intentionally not using the flux Activity model?

SirSplasch (changes requested):

src/Livewire/Support/Activities.php
patrickweh: yes

SirSplasch (approved):

PR #1650, opened 2026-04-28
Use address language for email salutation translation

patrickweh:
## Summary

Email salutation was rendered using the current application locale instead of the recipient address's language. A contact with English language received "Sehr geehrte Frau Lopez" while the rest of the template body was in English.

## Fix

`SalutationEnum::salutation()` now reads `language.language_code` from the passed address and forwards it as the locale to `__()`. When the address has no language relation, it falls back to the current application locale (existing behaviour).

## Tests

- `SalutationEnumTest`: covers the locale extraction from a plain address array with a nested `language` key.
- `AddressSalutationTest`: integration test creating an `Address` with an English `Language` relation and asserting `Dear Mrs. Lopez` is returned even when the app locale is `de`. Also asserts the German fallback when no language is set.

## Summary by Sourcery

Use the address language to determine the locale for email salutation translations while preserving the existing fallback to the application locale.

Bug Fixes:
- Ensure email salutations are translated using the recipient address language instead of always relying on the current application locale.

Tests:
- Add unit tests to verify salutation locale selection from an address array and to confirm correct English and German salutations based on address language and app locale.
- Add integration tests for the Address model to assert salutation respects the related Language and falls back to the app locale when no language is set.

### DIFF
--- a/src/Enums/SalutationEnum.php
+++ b/src/Enums/SalutationEnum.php
@@ -37,21 +37,23 @@ class SalutationEnum extends FluxEnum
             'company' => data_get($address, 'company'),
         ];
 
+        $locale = data_get($address, 'language.language_code');
+
         if (data_get($address, 'has_formal_salutation')) {
             return match ($case) {
-                SalutationEnum::Mrs => __('salutation.formal.mrs', $parameter),
-                SalutationEnum::Mr => __('salutation.formal.mr', $parameter),
-                SalutationEnum::Company => __('salutation.formal.company', $parameter),
-                SalutationEnum::Family => __('salutation.formal.family', $parameter),
-                default => __('salutation.formal.no_salutation', $parameter),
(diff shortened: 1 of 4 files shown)


src/Enums/SalutationEnum.php
```diff
@@ -29,31 +29,24 @@ public static function gender(string $case): string
...
+        $form = data_get($address, 'has_formal_salutation') ? 'formal' : 'informal';
+        $suffix = match ($case) {
+            SalutationEnum::Mrs => 'mrs',
+            SalutationEnum::Mr => 'mr',
+            SalutationEnum::Company => 'company',
+            SalutationEnum::Family => 'family',
+            default => 'no_salutation',
+        };
+
+        return __("salutation.{$form}.{$suffix}", $parameter, $locale);
```
SirSplasch: use single quotes for string

SirSplasch (changes requested):

SirSplasch (approved):

PR #1658, opened 2026-04-29
Use DataTableExport::store() instead of Excel::store

patrickweh:
## Summary
- `tall-datatables` 2.4.10 drops the `maatwebsite/excel` dependency and exposes its own `store()` method on `DataTableExport`.
- Switches `ExportDataTableJob` to call `$export->store($filePath)` directly instead of going through the Excel facade.
- Bumps the `team-nifty-gmbh/tall-datatables` constraint to `^2.4.10` so the new API is required.
- Removes the `Maatwebsite\Excel\ExcelServiceProvider` registration from `tests/TestCase.php` since the package is no longer transitively available.

## Summary by Sourcery

Switch data table export job to use the DataTableExport store API and align dependencies with the updated tall-datatables package.

Enhancements:
- Update ExportDataTableJob to invoke DataTableExport::store() directly instead of using the Excel facade.
- Remove the Excel service provider registration from the test application bootstrap as the Excel package is no longer required.

Build:
- Bump the team-nifty-gmbh/tall-datatables dependency to ^2.4.10 to require the new export API.

### DIFF
--- a/src/Jobs/ExportDataTableJob.php
+++ b/src/Jobs/ExportDataTableJob.php
@@ -43,12 +43,14 @@ class ExportDataTableJob implements ShouldQueue
         $folder = 'exports/' . str_replace(':', '_', $this->userMorph) . '/';
         $filePath = $folder . str_replace(['<', '>', ':', '"', '/', '\\', '|', '?', '*'], '_', $fileName);
 
-        $export = app(DataTableExport::class, [
-            'builder' => $query,
-            'exportColumns' => $this->columns,
-        ]);
-
-        $export->store($filePath);
+        app(
+            DataTableExport::class,
+            [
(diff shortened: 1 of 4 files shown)


src/Jobs/ExportDataTableJob.php
```diff
@@ -44,16 +43,13 @@ public function handle(): void
...
-            ),
-            $filePath
-        );
+        app(
+            DataTableExport::class,
+            [
+                'builder' => $query,
+                'exportColumns' => $this->columns,
+            ]
+        )->store($filePath);
```
SirSplasch: wrap
->store()

SirSplasch (changes requested):

SirSplasch (approved):

PR #1659, opened 2026-04-29
fix(MergeRecords): skip relations that throw QueryException

patrickweh:
## Summary
Spatie Permission 7.4 added a public `teams()` accessor on `HasRoles` that returns a `BelongsToMany` even when `permission.teams` is disabled. `MergeRecords` iterates over relations and runs pivot queries against the related table — for the disabled teams relation that table column (`model_has_roles.team_id`) doesn't exist, so the whole merge crashes.

Treat `QueryException` the same way `ReflectionException` is already treated: skip that relation and continue. Real DB errors on legitimately-broken relations now get skipped instead of crashing the merge — that's the trade-off.

## Summary by Sourcery

Bug Fixes:
- Skip relations that throw a QueryException while merging records so merges no longer crash on invalid or unavailable pivot tables or columns.

### DIFF
--- a/src/Actions/Record/MergeRecords.php
+++ b/src/Actions/Record/MergeRecords.php
@@ -218,7 +218,15 @@ class MergeRecords extends FluxAction
                             );
                         break;
                 }
-            } catch (ReflectionException|QueryException) {
+            } catch (ReflectionException) {
+                continue;
+            } catch (QueryException $e) {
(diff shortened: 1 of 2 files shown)


patrickweh:
Cherry-picked into #1658 — keeping all related fixes in one PR.

PR #1665, opened 2026-05-01
Add SubscriptionOrderFailedNotification and split schedule hooks

patrickweh:
## Summary

- Split `ScheduleRunCommand`'s `after()` hook into `onSuccess()` and `onFailure()` so each outcome has a dedicated branch (no behavior change for non-subscription schedules).
- Add `SubscriptionOrderFailedEvent` dispatched from `ProcessSubscriptionOrder`'s catch block whenever the subscription replication fails. The event holds serializable primitives (`exceptionClass`, `exceptionMessage`, `validationErrors`) so it stays queue-safe.
- Add `SubscriptionOrderFailedNotification` (`ShouldQueue`) listening to the event. The order's creator is auto-subscribed via `subscribeChannel()` so they receive the notification without needing prior opt-in. Title is prefixed with ⚠️.

This surfaces subscription processing failures (e.g., stale address snapshots, mis-configured order-type tenants) to the user who set up the order, instead of silently advancing `due_at` to the next cron occurrence and only logging an Activity entry.

## Summary by Sourcery

Introduce failure-specific handling for scheduled tasks and subscription order processing, including user-facing notifications for subscription failures.

New Features:
- Dispatch a SubscriptionOrderFailedEvent when subscription order replication fails, carrying serialized exception details.
- Send a queued SubscriptionOrderFailedNotification to the subscription order creator when processing fails, subscribing them automatically via the event channel.

Enhancements:
- Split ScheduleRunCommand hooks into explicit onSuccess and onFailure branches so success and failure outcomes are handled separately while still advancing due_at appropriately.

Tests:
- Add coverage to ensure failing invokable schedules advance due_at and update run metadata correctly.
- Add unit tests verifying SubscriptionOrderFailedEvent is dispatched with the expected payload on subscription processing failures.
- Add feature tests ensuring subscription order failures notify the order creator once and that no notification is sent when the order has no creator.

### DIFF
--- a/src/Console/Commands/ScheduleRunCommand.php
+++ b/src/Console/Commands/ScheduleRunCommand.php
@@ -77,22 +77,20 @@ class ScheduleRunCommand extends BaseScheduleRunCommand
             // Re-fetch from the DB and skip when last_run already covers the cron's
             // most recent past occurrence.
             $event->skip(function () use ($repeatable): bool {
-                $fresh = $repeatable->fresh();
-
-                if (! $fresh || ! $fresh->last_run || ! $fresh->cron_expression) {
+                if (! $repeatable->refresh()->last_run || ! $repeatable->cron_expression) {
                     return false;
                 }
 
                 try {
                     $previousOccurrence = Carbon::instance(
-                        (new CronExpression($fresh->cron_expression))
+                        (new CronExpression($repeatable->cron_expression))
                             ->getPreviousRunDate(now()->toDateTime(), 0, true)
                     );
                 } catch (Throwable) {
                     return false;
                 }
 
-                return $fresh->last_run->greaterThanOrEqualTo($previousOccurrence);
+                return $repeatable->last_run->greaterThanOrEqualTo($previousOccurrence);
             });
 
             foreach ($repeatable->cron['methods'] as $key => $method) {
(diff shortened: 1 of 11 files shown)


src/Invokable/ProcessSubscriptionOrder.php
```diff
@@ -93,6 +94,15 @@ public function __invoke(
 
             $activity->log(class_basename($e));
 
+            event(
+                SubscriptionOrderFailedEvent::make(
+                    $order,
+                    $e::class,
+                    $e->getMessage(),
+                    $e instanceof ValidationException ? $e->errors() : [],
+                )->subscribeChannel(array_filter([$order->getCreatedBy()]))
```
SirSplasch: wrap
->subscribeChannel()

src/Console/Commands/ScheduleRunCommand.php
```diff
@@ -68,6 +70,31 @@ public function handle(Schedule $schedule, Dispatcher $dispatcher, Cache $cache,
...
             }
 
+            // Guard against duplicate execution by an overlapping schedule:run process.
+            // withoutOverlapping only protects DURING execution; once the per-event
+            // mutex is released a second scheduler tick (whose in-memory $repeatable
+            // still has the past due_at) can pick the same row up and run it again.
+            // Re-fetch from the DB and skip when last_run already covers the cron's
+            // most recent past occurrence.
+            $event->skip(function () use ($repeatable): bool {
+                $fresh = $repeatable->fresh();
```
SirSplasch: refresh() instead of fresh() to ensure all values are present and being updated to the db table record value.

src/Console/Commands/ScheduleRunCommand.php
```diff
@@ -95,12 +122,19 @@ public function handle(Schedule $schedule, Dispatcher $dispatcher, Cache $cache,
...
-                    $nextRunDate = $nextRunDate->copy()->addMonthNoOverflow()->endOfMonth()
-                        ->setTime((int) $parts[1], (int) $parts[0]);
-                } else {
-                    $nextRunDate = (new CronExpression($event->expression))->getNextRunDate($nextRunDate);
+                // Only skip past the natural next occurrence when it falls on the same
+                // day - otherwise the schedule would double-fire today (e.g. yearlyOn at
+                // 06:00 when due_at was midnight). For schedules whose next occurrence
+                // is on a different day (monthly, lastDayOfMonth, quarterly, ...), that
+                // occurrence belongs to the NEXT cycle and must not be skipped.
+                if (Carbon::instance($nextRunDate)->isSameDay(now())) {
```
SirSplasch: What about schedules that run multiple times per day?

SirSplasch (changes requested):

src/Notifications/Order/SubscriptionOrderFailedNotification.php
```diff
@@ -0,0 +1,44 @@
...
+    public function subscribe(): array
+    {
+        return [
+            resolve_static(SubscriptionOrderFailedEvent::class, 'class') => 'sendNotification',
+        ];
+    }
+
+    protected function getDescription(): ?string
+    {
+        return $this->event->exceptionMessage;
```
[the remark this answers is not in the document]
patrickweh: Description now sanitizes: ValidationException renders the validation error messages, anything else falls back to a generic message so SQL/stack details don't reach the user. Tests in 7d1702cc2.

src/Invokable/ProcessSubscriptionOrder.php
patrickweh: Wrapped.

src/Console/Commands/ScheduleRunCommand.php
patrickweh: Switched to refresh().

src/Console/Commands/ScheduleRunCommand.php
patrickweh: Good catch — the same-day advance was skipping legitimate intra-day occurrences (e.g. hourly with due_at=06:00 missed at 06:30 ended up on 08:00 instead of 07:00). Now only advancing past the same-day next tick when due_at itself was not a valid cron occurrence (legacy/migration case like yearlyOn at 06:00 with due_at at midnight). Added a hourly-overdue regression test.

SirSplasch (approved):

PR #1668, opened 2026-05-04
Skip negative overtime accumulation for hourly employees

patrickweh:
## Summary

- Cap `plus_minus_overtime_hours` at zero from below in `CloseEmployeeDay::calculateDayData()` when `employees.salary_type = 'hourly'`. Positive overtime, sick/vacation accounting, and all other salary types remain untouched.
- Render the previously hidden `salary_type` field on the Employee admin form's "Employment Information" card with a helper hint, so the value can be configured from the UI instead of via the database.

## Background

The `employees.salary_type` column accepts `hourly | monthly | annual` and was validated by the rulesets, but its value was never consulted by any calculation. Hourly employees therefore accumulated negative overtime balances whenever they worked fewer hours than their nominal contractual target — which is conceptually wrong: an hourly worker is paid for the hours actually worked and cannot owe time. Positive overtime continues to be tracked as before; how to compensate it remains the company's call, the same as for other salary types.

No migration of historical `employee_days` is performed — only future day closes are affected. Manual correction of historical balances, if desired, is out of scope.

## Summary by Sourcery

Prevent hourly employees from accumulating negative overtime and expose salary type configuration in the employee admin UI.

Bug Fixes:
- Stop recording negative plus/minus overtime balances for hourly salary type when actual hours are below target, including during vacations and sick leave.

Enhancements:
- Ensure positive overtime continues to be tracked for hourly employees while capping negative balances at zero.
- Add a selectable salary type field to the employee general information form with contextual guidance.
- Extend employee day closing logic tests to cover hourly salary type behavior across work, vacation, and sick leave scenarios.
- Update Livewire employee component tests to assert visibility of the new salary type field.


src/Models/Employee.php
```diff
@@ -67,6 +68,7 @@ protected static function booted(): void
     protected function casts(): array
     {
         return [
+            'salary_type' => SalaryTypeEnum::class,
```
SirSplasch: put this in table order, so after 'residence_permit_until'

SirSplasch (changes requested):

SirSplasch (approved):

PR #1684, opened 2026-05-05
Show confirm dialog when dragging recurring calendar events

patrickweh:
## Summary
- Dragging a recurring event in the calendar fired `calendar-event-change`, which then saved the event directly via `editEvent()` without ever prompting the user.
- The confirm dialog with the `this event` / `this and following` / `all events` radios is rendered in `calendar-event-edit.blade.php` but only opened from the Save button inside the edit modal — never from a drag-drop action.
- After a drag, `confirm_option` defaulted to `'this'` and the form was saved straight away (introduced by `1d65d4bf` "fix drag n drop on repeatable calendar events"), so the user had no way to apply the change to following or all occurrences.

## Changes
- `editEvent()` now branches on `was_repeatable` for `event-change`: repeatable events open the confirm-dialog instead of saving directly, while non-repeatable events keep the existing direct-save path.
- A new `set-confirm-dialog-type` window event lets the dispatched JS set the dialog type so the Save button is visible alongside the radios.
- Added a browser test that drags a recurring event via `calendar-event-change` and asserts the confirm-dialog plus the future/all radios appear and the source event is not silently mutated.

## Summary by Sourcery

Handle drag-and-drop of recurring calendar events by opening the confirmation dialog instead of saving immediately.

Enhancements:
- Update calendar event editing logic to route drag-change actions on recurring events through the confirmation dialog flow while preserving direct-save behavior for non-recurring events.
- Allow the calendar event edit view to receive a confirm-dialog type via a window event so the appropriate dialog mode is set when opened from drag-drop.

Tests:
- Add a browser test to verify that dragging a recurring calendar event opens the confirmation dialog, shows future/all occurrence options, and does not persist changes before user confirmation.


patrickweh:
Both points checked, declining both:

**Point 1 (test arrow wrapper not invoked):** The `() => { ... }` wrapper is the Pest browser convention — every existing browser test in this file uses it (e.g. the `editing a calendar event opens modal` and `clicking different events` tests dispatch Livewire events through the same shape). Pest's `script()` invokes the arrow function. Concrete evidence: this exact test was red before the fix (`#confirm-dialog` not found) and green after. If the dispatch did not fire, both states would be the same.

**Point 2 (`was_repeatable` branch not trigger-guarded):** It already is. Calendar.php:160 sits inside the `if ($trigger === 'event-change')` block opened on Calendar.php:156. The early `return` after the dispatch keeps the branch from leaking into the non-repeatable save path. Other edit flows (`event-click`, etc.) take the `elseif`/`else` branches further down and never enter this block.

SirSplasch (approved):

PR #1693, opened 2026-05-06
Stop dehydrating media downloads through Livewire's response payload

patrickweh:
## Summary

`Livewire\Features\SupportFileDownloads` intercepts every `BinaryFileResponse` / `StreamedResponse` returned from a component, calls `sendContent()` inside `ob_start()` to capture the full body, then base64-encodes it into the response payload. For a 70 MB attachment that costs ~93 MB of base64 on top of the 70 MB output buffer, which is enough to exhaust the customer's 128 MB `memory_limit` on `/projects/{id}` (Comments tab) and other pages that load media via Livewire actions.

## Changes

Every Livewire component download method that previously returned a file response now triggers a client-side redirect to the existing streamed media route. `window.location.href = signed-url` issues a separate HTTP request served by Spatie's `StreamedResponse`, which is chunked and never buffers the body.

Affected entry points:

- `FluxErp\Livewire\DataTables\MediaList::downloadMedia`
- `FluxErp\Livewire\DataTables\PurchaseInvoiceList::downloadMedia`
- `FluxErp\Livewire\EditMail::downloadAttachment`
- `FluxErp\Livewire\Mail\Mail::download`
- `FluxErp\Traits\Livewire\WithFileUploads::download` (used by Comments, MediaUploadForm, FolderTree, EmailTemplates and every other consumer of the trait)

`WithFileUploads::download` reuses the existing `DownloadMedia` action by passing `as => 'url'` so the URL generation (signed `media.private` for non-public disks, public URL otherwise) keeps living in one place.

A regression test on `MediaList::downloadMedia` asserts the response carries an `xjs` effect (the redirect) and never the `download` effect that `SupportFileDownloads` injects when it dehydrates a file response.

## Summary by Sourcery

Avoid returning file responses from Livewire components by triggering client-side redirects to media URLs instead, preventing Livewire from dehydrating large downloads into the response payload.

Bug Fixes:
- Prevent out-of-memory errors caused by Livewire dehydrating large media downloads into base64-encoded response payloads.

Enhancements:
- Standardize media downloads across Livewire components to use JS redirects to existing streamed media routes, reusing the central DownloadMedia action for URL generation.

Tests:
- Add a regression test for MediaList::downloadMedia to ensure downloads trigger a client-side redirect and no Livewire file-download effect is generated.

### DIFF
--- a/src/Livewire/DataTables/MediaList.php
+++ b/src/Livewire/DataTables/MediaList.php
@@ -11,7 +11,6 @@ use Illuminate\Support\Str;
 use Illuminate\Validation\ValidationException;
 use Livewire\Attributes\Renderless;
 use Spatie\Permission\Exceptions\UnauthorizedException;
-use Symfony\Component\HttpFoundation\BinaryFileResponse;
 use TeamNiftyGmbH\DataTable\Htmlables\DataTableButton;
 
 class MediaList extends BaseDataTable
@@ -64,7 +63,8 @@ class MediaList extends BaseDataTable
         ];
     }
 
-    public function downloadMedia(Media $media): false|BinaryFileResponse
+    #[Renderless]
+    public function downloadMedia(Media $media): bool
     {
         if (! file_exists($media->getPath())) {
             $this->toast()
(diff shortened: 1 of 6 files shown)


patrickweh:
Closing — the diagnosis was wrong. The actual failing call on /projects/1875 is `downloadCollection` (zipping 20 attachments) in `WithFileUploads`, not the four single-file download methods this PR patched. Need to step back and check whether downloads are even the dominant cause of memory exhaustion across the customer instance before picking a fix.

PR #1694, opened 2026-05-06
Stream media downloads via signed routes instead of dehydrating through Livewire

patrickweh:
## Summary

`Livewire\Features\SupportFileDownloads` intercepts every `BinaryFileResponse` / `StreamedResponse` returned from a component, captures the body inside `ob_start()` and base64-encodes it into the JSON response payload. For an on-the-fly zip of attachments that triples memory usage on top of the buffer itself — production hit `"Allowed memory size of 134217728 bytes exhausted"` five different ways on `/projects/1875` alone, all rooted in `WithFileUploads::downloadCollection`.

### Trait split

`WithFileUploads` becomes a thin composer of two focused traits, so the ten existing consumers keep working unchanged:

- `SupportsFileUploads` — Livewire's upload base, `prepareForMediaLibrary`, `removeFileUpload`, `saveFileUploadsToMediaLibrary`.
- `SupportsFileDownloads` — `download(Media)` and `downloadCollection(...)`, both of which now generate a signed URL and hand control back to the browser via `$this->redirect()`. No bytes ever pass through Livewire's payload.

### Streamed zip route

For the on-the-fly zip case, Livewire pre-validates the `DownloadMultipleMedia` action data, encrypts `{data, name}` into a 5-minute signed URL to a new route `/media-collection-download/{token}`, and redirects. The route streams the zip via `response()->streamDownload()` and ZipStream-PHP's `getZipStream()` — chunked output, never buffered.

### Component cleanup

The four components that hand-rolled their own download helpers drop the duplicates:

- `MediaList::downloadMedia` → removed; component now `use SupportsFileDownloads`. Row action calls `$wire.download(record.id)`.
- `PurchaseInvoiceList::downloadMedia` → removed; trait inherited via existing `WithFileUploads`.
- `EditMail::downloadAttachment` → removed; trait inherited via existing `WithFileUploads`. Blade switches to `$wire.download(file.id)`.
- `Mail::download` → removed; component now `use SupportsFileDownloads`. Same method name, no blade change.

## Tests

`MediaListTest` asserts:
- `download` produces a Livewire redirect effect and **no** `download` effect (the latter would mean `SupportFileDownloads` kicked in).
- `downloadCollection` redirects to the new `media-collection-download` URL with the same guarantee.

## Breaking

`$wire.downloadMedia(...)` and `$wire.downloadAttachment(...)` no longer exist on `MediaList` / `EditMail`. Consumer blades that called them (sbm-flux, customer-specific extensions) must rename their calls to `$wire.download(...)`.

## Summary by Sourcery

Stream media downloads from Livewire components via redirects to signed routes instead of returning file responses through Livewire, and split file upload/download responsibilities into dedicated traits.

New Features:
- Add a dedicated SupportsFileDownloads trait that redirects media and media-collection downloads to streamed routes using signed URLs.
- Introduce a SupportsFileUploads trait and compose it with SupportsFileDownloads in the WithFileUploads trait.
- Add a signed media-collection download route that streams ZIP archives directly to the client.

Enhancements:
- Update Mail, MediaList, PurchaseInvoiceList, and EditMail components to reuse the shared SupportsFileDownloads trait instead of custom download methods.
- Adjust EditMail Blade and MediaList row actions to invoke the shared download method.
- Pre-validate multi-media download requests in Livewire before generating signed download URLs to avoid redirecting into authorization or validation errors.

Tests:
- Extend MediaList Livewire tests to assert that media and collection downloads now trigger redirects and avoid Livewire's built-in download effect.


routes/frontend/web.php
```diff
@@ -373,5 +376,21 @@
...
                 return $media;
             })->name('media.private');
+
+            // The signed URL itself is the authorization — Livewire validated
+            // permission and data shape before generating it. Same threat
+            // model as media.private above.
+            Route::get('/media-collection-download/{token}', function (string $token) {
+                $payload = Crypt::decrypt($token);
+
+                $stream = DownloadMultipleMedia::make(data_get($payload, 'data', []))->execute();
```
SirSplasch: Why no checkPermission() and validate()? Just to prevent unwanted access.
I know you are validating this before you redirect here, but I my opinion you have to view this as a standalone route.

SirSplasch: wrap
->execute()

routes/frontend/web.php
```diff
@@ -373,5 +376,21 @@
...
+
+                return response()->streamDownload(
+                    fn () => $stream->getZipStream(),
+                    Str::finish((string) data_get($payload, 'name', 'media'), '.zip'),
```
SirSplasch: ?? 'media' instead of passing it as a fallback, just to be sure it cannot be null.

SirSplasch (changes requested):

SirSplasch:
also check ai annotations

SirSplasch (approved):

PR #1696, opened 2026-05-06
Init Alpine on tiptap toolbar after appending button commands

patrickweh:
## Summary
- Cloned `<template>` button commands inside the tiptap controlPanel were no longer auto-initialized by Alpine after the bump to alpinejs 3.15.12, leaving dropdown buttons (Table, Headings, etc.) inert with `x-data` and `x-on:click.prevent="onClick"` unwired.
- Call `Alpine.initTree(controlPanel)` right after appending the cloned commands, mirroring the existing pattern used on the floating tooltip variant a few lines down.

## Summary by Sourcery

Bug Fixes:
- Restore Alpine.js initialization for cloned tiptap toolbar button commands so dropdown controls respond to clicks as expected.

### DIFF
--- a/resources/js/components/tiptap.js
+++ b/resources/js/components/tiptap.js
@@ -252,6 +252,10 @@ export default function (
 
                 this.proxy = Alpine.raw(_editor);
 
+                if (!showTooltipDropdown) {
+                    Alpine.initTree(controlPanel);
+                }
+
                 element.dataset.tiptapInitialized = 'true';
 
                 this.$watch('editable', (editable) => {


patrickweh:
Both suggestions don't apply here:

1. **Init only the cloned node** — `commands` is a `<template>`, so `commands.content` is a `DocumentFragment`. After `appendChild(fragment)` the fragment's children are *moved* into the parent and the fragment itself becomes empty (per DOM spec). `Alpine.initTree(clone)` after the append would no-op on an empty fragment. The existing pattern in this same file at L190–194 does the equivalent on the floating tooltip variant (`floatingElement.appendChild(actions); ... Alpine.initTree(floatingElement)`); this fix mirrors it. Also: `controlPanel.innerHTML = ''` runs a few lines earlier, so there are no pre-existing children to risk re-binding.

2. **Guard with `if (window.Alpine?.initTree)`** — `Alpine.raw()` / `Alpine.initTree()` are used unguarded elsewhere in this file (L31, 194, 253), and the whole editor setup lives inside an Alpine `x-data`. If Alpine isn't present, nothing here runs anyway. A guard would be inconsistent and YAGNI.

PR #1699, opened 2026-05-07
Load OpenCV via classic script tag for mobile WebView compatibility

patrickweh:
## Summary

- `loadOpenCv()` now injects a classic `<script src="/flux/opencv.js">` tag and waits for `cv.onRuntimeInitialized` instead of `await import('@techstark/opencv-js')`. Dynamic import of the 10 MB UMD bundle hangs indefinitely on iOS WKWebView (Capacitor mobile app) — fetch completes, parse/instantiate never resolves and never throws. The HTML-parser path streams large scripts reliably across mobile WebViews.
- `npm run build` now copies `node_modules/@techstark/opencv-js/dist/opencv.js` to `dist/opencv.js` so `FrontendAssets`' `/flux/{file}` route can serve it.
- `FrontendAssets::pretendResponseIsFile()` appends `no-transform` to the cache-control header. FrankenPHP / Caddy's `encode zstd br gzip` mis-frames brotli responses for PHP-generated bodies (HEAD returns `content-length: 1`, GET drops it entirely) which iOS WKWebView then refuses to consume. `no-transform` tells the encoder to leave the body alone.
- `processImage` now wraps `cv.imread` / `findPaperContour` in `try/catch/finally` and resets `isProcessing` on `img.onerror`, so a detection failure or load error no longer freezes the spinner forever.

## Summary by Sourcery

Load OpenCV via a classic script tag and hard-copied bundle to improve mobile WebView compatibility and ensure scanner UI recovers gracefully from loading and detection failures.

New Features:
- Serve a bundled opencv.js file via the existing /flux asset route and load it through a classic script tag instead of dynamic import.

Bug Fixes:
- Prevent iOS WKWebView from hanging on large OpenCV bundles by avoiding dynamic import and relying on script tag loading with runtime initialization.
- Ensure document scanning failures or image load errors reset processing state so the UI spinner does not remain stuck.
- Avoid broken brotli-encoded responses in FrankenPHP / Caddy by appending no-transform to cache-control for pseudo-file responses, preventing WebView hangs.

Build:
- Extend the build script to copy the OpenCV UMD bundle into the dist directory for serving by the frontend assets mechanism.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
     "type": "module",
     "scripts": {
         "dev": "vite",
-        "build": "vite build && cp node_modules/@techstark/opencv-js/dist/opencv.js dist/opencv.js",
+        "build": "vite build && node scripts/copy-opencv.mjs",
         "format": "prettier --write ."
     },
     "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
     "type": "module",
     "scripts": {
         "dev": "vite",
-        "build": "vite build",
+        "build": "vite build && cp node_modules/@techstark/opencv-js/dist/opencv.js dist/opencv.js",
         "format": "prettier --write ."
     },
     "devDependencies": {
--- a/resources/js/components/document-scanner.js
+++ b/resources/js/components/document-scanner.js
@@ -222,13 +222,55 @@ export default ($wire) => ({
             return;
         }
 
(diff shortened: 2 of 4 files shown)


SirSplasch (approved):
Check AI annotations if they are valid. I don't have enough understanding of this for an adequate review.

PR #1700, opened 2026-05-07
Add unassigned purchase invoices widget with document scanner

patrickweh:
## Summary

- New dashboard widget `UnassignedPurchaseInvoices` lists every purchase invoice without an `order_id` using TallStackUI's `<x-list>` with client-side search and a per-row ⋮ menu. Caption shows supplier · invoice date · formatted amount · invoice number, falling back to the attached file name (or `#id`) when supplier and invoice number are both blank.
- The menu's "Edit" entry calls `goToPurchaseInvoice($id)` on the widget, which stores the id in the session and navigates to `/accounting/purchase-invoices`. `PurchaseInvoiceList::mount()` pulls the id, fills the edit form via the new shared `fillEditFormFromPurchaseInvoice()` method, and flips a public `autoOpenEditModal` flag.
- An Alpine `x-init` in the existing `include-before` partial reads `$wire.autoOpenEditModal` and opens the modal on the next tick. `$this->js()` / `$this->dispatch()` from `mount()` don't reliably reach the client across `wire:navigate`, so the flag-driven approach replaces them only for the navigate path — the regular `wire:click="edit(...)"` flow on the list page still uses `$this->js()` as before.
- Always-visible Scan button in the widget header opens a scoped scanner modal that reuses the `documentScanner` Alpine component and the `WithDocumentScanning` trait, dispatching `CreatePurchaseInvoice` for each scan.

## Summary by Sourcery

Introduce a dashboard widget for unassigned purchase invoices and enable deep-linking into the purchase invoice edit modal when navigating from the widget.

New Features:
- Add an UnassignedPurchaseInvoices dashboard widget listing purchase invoices without an associated order and providing quick navigation to edit them.
- Add a scanner modal to the unassigned purchase invoices widget to capture and upload documents for creating purchase invoices.

Enhancements:
- Refactor purchase invoice editing to share form-population logic via a dedicated helper method used by both in-place edits and navigation-triggered edits.
- Add an auto-open flag on the purchase invoice list to reliably trigger the edit modal after wire:navigate using Alpine initialization.


resources/views/livewire/accounting/purchase-invoice-list/include-before.blade.php
```diff
@@ -1,4 +1,16 @@
 <div>
+    <div
+        x-data
```
SirSplasch, line 3: What does the x-data do? Can this be removed?

src/Livewire/Widgets/UnassignedPurchaseInvoices.php
```diff
@@ -0,0 +1,117 @@
...
+
+        $captionParts = array_filter([
+            $invoice->invoice_date?->format('d.m.Y'),
+            ! is_null($invoice->total_gross_price)
+                ? number_format(
+                    (float) $invoice->total_gross_price,
+                    2,
+                    ',',
+                    '.',
+                ) . ' ' . (data_get($invoice, 'currency.iso') ?? '')
```
SirSplasch: Why not use Number::currency()?

src/Livewire/Widgets/UnassignedPurchaseInvoices.php
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    protected function getScannedDocumentAction(): string
+    {
+        return CreatePurchaseInvoice::class;
+    }
+
+    protected function toListItem(PurchaseInvoice $invoice): array
+    {
+        $supplier = data_get($invoice, 'contact.invoiceAddress.name');
```
SirSplasch: Intentionally getting name, instead of $invoice->contact->invoiceAddress->getLabel()?

resources/views/livewire/widgets/unassigned-purchase-invoices.blade.php
```diff
@@ -0,0 +1,332 @@
...
+        >
+            <div
+                x-data="documentScanner($wire)"
+                x-on:keydown.escape.window="
+                    if (isEditing) {
+                        closeEditor();
+                        $event.stopPropagation();
+                    }
+                "
+            >
```
SirSplasch, line 53: I feel like I have seen this code already somewhere. Is there no component for the document scanner, or does this have to be implemented every time separately?

SirSplasch (changes requested):

resources/views/livewire/accounting/purchase-invoice-list/include-before.blade.php
patrickweh: The x-data is required to bootstrap an Alpine context — without it, x-init doesn't fire. The empty form is the documented Alpine pattern when you only need lifecycle (no reactive state of your own). Removing it would silently break the auto-open behaviour after wire:navigate.

resources/views/components/document-scanner/controls.blade.php
SirSplasch, line 1: Why does this component not have a singular root element?
Is it intentional not to put all of this in a `<div>` element?

SirSplasch (changes requested):

SirSplasch (approved):

PR #1702, opened 2026-05-08
Add backend foundation for product variant inheritance

patrickweh:
## Summary

Variants now inherit field values and pivot-relation rows from their parent product through an explicit `overridden_fields` mechanism. Solves the recurring pain that adding/correcting a field at the parent (e.g. a price) required manual re-entry on every variant.

## What's in this PR

- New columns on `products`: `overridden_fields` (JSON, list of fields the variant explicitly owns) and `was_parent` (sticky flag once a product has had any child).
- New tenant toggle `product_variant_inheritance_enabled` (default false for existing tenants, true for new ones via creating-hook).
- `InheritsFromParent` trait on `Product` with field/relation resolvers, write-tracking, reset helpers, and bulk-reset across variants.
- Inheritable scope today: `name`, `description`, dimensions, units, vat, etc. (26 fields) and relations `prices`, `categories`, `productProperties`, `suppliers`. Tags + crossSellings + media are deferred (Spatie coupling / collection-level semantics).
- Relation rename `prices()` → `ownPrices()` (and analogous for `categories`, `productProperties`, `suppliers`); `$product->prices` accessor returns the resolved Collection.
- Validation rules: `ProductHierarchyDepth` (max 2 levels, no variants-of-variants) and `ProductIsOrderable` (parents with active children, or `was_parent=true`, are not orderable). Wired into `CreateOrderPosition`, `UpdateOrderPosition`, `CreateStockPosting`.
- Orphaned-parent state: `was_parent` stays sticky after the last variant is removed; `PromoteParentToStandalone` action explicitly clears it once the user has decided.
- Reset Actions: `ResetProductField`, `ResetProductRelation`, `ResetFieldOnAllVariants`, `ResetRelationOnAllVariants`.
- One-shot data migration: `MigrateProductVariantInheritanceJob` auto-diffs existing variants vs parents and writes `overridden_fields`. CLI: `flux:product-variants:migrate-inheritance [--parent-id=…]`.

## Out of scope (follow-up branch)

- All UI surfaces — variant edit form indicators, parent bulk-reset buttons, orphaned-parent banner, audit reporter.

## Summary by Sourcery

Introduce product variant inheritance with parent-driven field and relation resolution, controlled by a tenant feature flag, plus supporting migrations, actions, rules, and a migration job/command for backfilling existing data.

New Features:
- Add InheritsFromParent trait and configuration on Product to allow variants to inherit a whitelisted set of fields and relations from their parent.
- Expose effective inherited collections for prices, categories, product properties, and suppliers while keeping dedicated own* relations for direct assignments.
- Add tenant-level product_variant_inheritance_enabled toggle with model defaults and caching behavior.
- Introduce actions to reset variant fields/relations (per-variant and bulk) and to promote former parent products back to standalone.
- Add console command and queued job to migrate existing variants by computing overridden_fields and pruning redundant pivot data.
- Add validation rules to restrict product hierarchy depth and prevent ordering or stock posting against parent/orphaned products.

Enhancements:
- Refine product creation, update, pricing, and variant-generation flows to respect inheritance behavior and avoid duplicating parent data when enabled.
- Track and persist a sticky was_parent flag on products to reflect historical parent status and gate orderability and promotion workflows.

Tests:
- Add extensive unit and feature tests covering inheritance behavior, reset actions, tenant toggle semantics, new validation rules, migrations, and console command paths.
- Update existing tests and factories to use own* relations where appropriate and verify legacy behavior when inheritance is disabled.


database/migrations/2026_05_07_120000_add_inheritance_columns_to_products_table.php
SirSplasch, line 1: Better not check for hasColumn() as this package is the authority over this table. If these columns already exist it is better to fail, as to not fail, because these existing columns may be used differently and can have a different column type.

database/migrations/2026_05_07_120000_add_inheritance_columns_to_products_table.php
```diff
@@ -0,0 +1,36 @@
...
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class() extends Migration
+{
+    public function up(): void
+    {
+        Schema::table('products', function (Blueprint $table): void {
+            if (! Schema::hasColumn('products', 'overridden_fields')) {
+                $table->json('overridden_fields')->nullable()->after('parent_id');
```
SirSplasch: Maybe after 'warning_stock_amount'? Surely not after 'parent_id'.

database/migrations/2026_05_07_120000_add_inheritance_columns_to_products_table.php
```diff
@@ -0,0 +1,36 @@
...
+            }
+
+            if (! Schema::hasColumn('products', 'was_parent')) {
+                $table->boolean('was_parent')->default(false)->after('overridden_fields');
```
SirSplasch: after 'is_shipping_free'

database/migrations/2026_05_07_120000_add_inheritance_columns_to_products_table.php
```diff
@@ -0,0 +1,36 @@
...
+                $table->index('was_parent');
```
SirSplasch: can add index in column creation, no need to add it in an extra statement.

database/migrations/2026_05_07_120100_add_product_variant_inheritance_enabled_to_tenants_table.php
SirSplasch, line 1: This package is the authority over this table, better not check for hasColumn().

database/migrations/2026_05_07_120100_add_product_variant_inheritance_enabled_to_tenants_table.php
```diff
@@ -0,0 +1,30 @@
...
+return new class() extends Migration
+{
+    public function up(): void
+    {
+        if (Schema::hasColumn('tenants', 'product_variant_inheritance_enabled')) {
+            return;
+        }
+
+        Schema::table('tenants', function (Blueprint $table): void {
+            $table->boolean('product_variant_inheritance_enabled')
```
SirSplasch: I don't know where this is going currently, but to me this boolean does not make much sense, as a product is not fixed to a tenant, it can have multiple tenants.

src/Models/Product.php
```diff
@@ -52,13 +54,57 @@ class Product extends FluxModel implements HasMedia, HasMediaForeignKey, Interac
 
     protected ?string $detailRouteName = 'products.id';
 
+    protected array $inheritableFields = [
```
SirSplasch, line 59: order values according to table

src/Models/Product.php
```diff
@@ -52,13 +54,57 @@ class Product extends FluxModel implements HasMedia, HasMediaForeignKey, Interac
...
+        'has_serial_numbers',
+        'is_active_export_to_web_shop',
+        'selling_unit',
+        'basic_unit',
+        'purchase_unit_id',
+        'reference_unit_id',
+        'cover_media_id',
+    ];
+
+    protected array $inheritableRelations = [
```
SirSplasch, line 88: order values alphabetically

src/Models/Product.php
SirSplasch, line 196: order casts according to table

src/Models/Tenant.php
SirSplasch, line 1: No fan of adding this boolean to the tenant as a product can have multiple tenants. Better add it as a setting.

src/Rules/ProductHierarchyDepth.php
SirSplasch, line 1: Can't you add a whereNull('parent_id') to the ModelExists in the ruleset?
Or did you want special user feedback? Even if that is what you want to achieve, I dislike the idea of querying the same table multiple times just to put extra user feedback. There must be a better way, a way where the table will only be queried once.

src/Rules/ProductIsOrderable.php
SirSplasch, line 1: Same as in the ProductHierachyDepth Rule. I dislike the fact that the table will be queried an extra time just to present extra user feedback.

database/migrations/2026_05_07_120050_backfill_was_parent_on_existing_products.php
```diff
@@ -0,0 +1,32 @@
...
+            WHERE id IN (
+                SELECT DISTINCT parent_id
+                FROM (SELECT parent_id FROM products WHERE parent_id IS NOT NULL) AS p
+            )
+        ');
+    }
+
+    public function down(): void
+    {
+        // No-op: schema down() in the prior migration drops the column entirely.
```
SirSplasch: no need for this comment

database/migrations/2026_05_07_120050_backfill_was_parent_on_existing_products.php
```diff
@@ -0,0 +1,32 @@
...
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Schema;
+
+return new class() extends Migration
+{
+    public function up(): void
+    {
+        if (! Schema::hasColumn('products', 'was_parent')) {
```
SirSplasch: no need to check this, this package is the authority, if this does not exist, this should fail.

database/migrations/2026_05_07_120050_backfill_was_parent_on_existing_products.php
```diff
@@ -0,0 +1,32 @@
...
+
+        // One-shot backfill. On installs with 100k+ products this single UPDATE
+        // holds row and gap locks for its full duration; large operators may prefer
+        // to run the equivalent SQL manually in batches before deploying this migration.
+        DB::statement('
+            UPDATE products
+            SET was_parent = 1
+            WHERE id IN (
+                SELECT DISTINCT parent_id
+                FROM (SELECT parent_id FROM products WHERE parent_id IS NOT NULL) AS p
```
SirSplasch: Am I missing something, or can this whole subquery be condensed to
```
SELECT DISTINCT parent_id
FROM products
WHERE parent_id IS NOT NULL
```

SirSplasch (changes requested):
I have only checked around a quarter of files. But I want you to have some feedback already.

database/migrations/2026_05_07_120050_backfill_was_parent_on_existing_products.php
patrickweh: MySQL forbids self-reference in `UPDATE … WHERE id IN (SELECT … FROM products)`. that's what the derived-table wrap was bypassing. switched to `INNER JOIN (SELECT DISTINCT parent_id …) AS parents ON products.id = parents.id`, flatter and avoids the restriction.

database/migrations/2026_07_06_120000_add_is_inherited_to_variant_relations.php
SirSplasch, line 1: $table instead of $t

database/migrations/2026_07_06_120000_add_is_inherited_to_variant_relations.php
```diff
@@ -0,0 +1,27 @@
...
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class() extends Migration
+{
+    public function up(): void
+    {
+        foreach (['prices', 'categorizable', 'product_supplier', 'product_product_property'] as $table) {
```
SirSplasch: I don't think a foreach is the solution here, as you have to put the boolean at specific positions in the table.
For example in the prices table it is after('price'), but in the categorizable table it is after('categorizable_id')

database/migrations/2026_05_07_120050_backfill_was_parent_on_existing_products.php
SirSplasch, line 1: This can be done in the migration which the was_parent column was added.

database/migrations/2026_05_07_120000_add_inheritance_columns_to_products_table.php
```diff
@@ -0,0 +1,24 @@
...
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class() extends Migration
+{
+    public function up(): void
+    {
+        Schema::table('products', function (Blueprint $table): void {
+            $table->json('overridden_fields')->nullable()->after('warning_stock_amount');
+            $table->boolean('was_parent')->default(false)->after('is_shipping_free')->index();
```
SirSplasch: Why is this column called "was_parent"? Is it no parent anymore? I thought was_parent = true if the product has children.

If I understand this correctly this boolean indicates if a product is a root/baseline for its children aka variants.
So why not name it what it actually represents?

src/Settings/ProductSettings.php
```diff
@@ -0,0 +1,30 @@
...
+        return 'product';
+    }
+
+    /**
+     * Resilient gate for the variant inheritance feature. Falls back to the
+     * create_product_settings migration default (true) when the setting has
+     * not been migrated yet, so product writes never crash on an app (or a
+     * schema-dumped test database) where the row is absent.
+     */
+    public static function variantInheritanceEnabled(): bool
```
SirSplasch: I don't think this method is needed, just add a default value to the $variant_inheritance_enabled property.

src/Support/VariantInheritance/InheritanceSync.php
SirSplasch, line 1: This file can be removed, this method is only ever called once in the Product model, so this one-liner can be done there.

src/Actions/Product/PromoteParentToStandalone.php
SirSplasch, line 1: Does this action exist to have a permission for that? Otherwise this is kinda useless, as it only sets a boolean to false and therefore can this functionality be covered with the UpdateProduct action.

src/Rulesets/Product/PromoteParentToStandaloneRuleset.php
```diff
@@ -0,0 +1,43 @@
...
+                    if (! $product->was_parent) {
+                        $fail(__('This product is not flagged as a former parent and does not need to be promoted to standalone.'));
+
+                        return;
+                    }
+
+                    if ($product->children()->where('is_active', true)->exists()) {
+                        $fail(__('Cannot promote: active variants still exist.'));
+                    }
+                },
```
SirSplasch: This is essentially only a ->where('was_parent', true)->whereDoesntHaveRelation('children', 'is_active', true) on the ModelExists rule.

src/Actions/Product/ResetFieldOnAllVariants.php
SirSplasch, line 1: Why is this action limited to only one field at a time, make multiple fields possible.

src/Actions/Product/ResetProductField.php
SirSplasch, line 1: Make multiple fields possible.

src/Actions/Product/ResetProductRelation.php
SirSplasch, line 1: Make multiple relations possible.

src/Actions/Product/ResetRelationOnAllVariants.php
SirSplasch, line 1: Make multiple relations possible.

SirSplasch, line 1: In my mind there is not that big of a difference between this action and the ResetProductRelation action, the only difference is, that here it affects all variants and in the other a single one. Can't they be merged? And what about if I want so reset stuff on multiple variants, but not all? Do I really have to call ResetProductRelation on all of them separately?

src/Actions/Product/ResetProductRelation.php
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    public function performAction(): Product
+    {
+        $variant = resolve_static(Product::class, 'query')
+            ->whereKey($this->getData('id'))
+            ->firstOrFail();
+
+        $relation = $this->getData('relation');
+        $key = $this->getData('key');
```
SirSplasch: What does key even mean, this can be anything. I know I can look deep into the implementation to find out what it is used for, but if I only see this action, or even only the rules, this is totally unclear.

src/Actions/Product/ResetFieldOnAllVariants.php
SirSplasch, line 1: In my mind there is not that big of a difference between this action and the ResetProductField action, the only difference is, that here it affects all variants and in the other a single one. Can't they be merged? And what about if I want so reset stuff on multiple variants, but not all? Do I really have to call ResetProductField on all of them separately?

database/migrations/2026_07_07_120000_backfill_materialized_variant_values.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+use FluxErp\Jobs\MigrateProductVariantInheritanceJob;
+use Illuminate\Database\Migrations\Migration;
+
+return new class() extends Migration
+{
+    public function up(): void
+    {
+        (new MigrateProductVariantInheritanceJob())->handle();
```
SirSplasch: This is a very bad idea. Migrations run once, and should not be altered over time, so that they run exactly the same every single time. And with this, the migration can totally change, even fail if this class does not exist anymore.

src/Actions/Product/CreateProduct.php
```diff
@@ -99,7 +99,13 @@ public function prepareForValidation(): void
             $this->data['bundle_type_enum'] = null;
         }
 
-        if (! data_get($this->data, 'prices') && data_get($this->data, 'parent_id')) {
+        $inheritanceEnabled = app(Product::class)->inheritanceEnabled();
```
SirSplasch: Why do a round trip to find out if inheritance is enabled?
Just call app(ProductSettings;;ckass)->variant_inheritance_enabled

src/Console/Commands/MigrateProductVariantInheritanceCommand.php
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    public static function withoutOverlapping(): bool
+    {
+        return true;
+    }
+
+    public function handle(): int
+    {
+        if (! ProductSettings::variantInheritanceEnabled()) {
```
SirSplasch: app(ProductSettings::class)->variant_inheritance_enabled

src/Console/Commands/MigrateProductVariantInheritanceCommand.php
```diff
@@ -0,0 +1,66 @@
...
+
+            return self::FAILURE;
+        }
+
+        $parentId = $this->option('parent-id') !== null
+            ? (int) $this->option('parent-id')
+            : null;
+
+        $this->info('Starting variant-inheritance migration job (sync) ...');
+        (new MigrateProductVariantInheritanceJob($parentId))->handle();
```
SirSplasch: app()

src/Console/Commands/MigrateProductVariantInheritanceCommand.php
SirSplasch, line 1: Why is this command needed? Like through what can the variants be broken? Shouldn't the system be resilient to such things?

src/Rulesets/Product/ResetFieldOnAllVariantsRuleset.php
```diff
@@ -0,0 +1,30 @@
...
+    protected static ?string $model = Product::class;
+
+    public function rules(): array
+    {
+        return [
+            'parent_id' => [
+                'required',
+                'integer',
+                app(ModelExists::class, ['model' => Product::class]),
+                Rule::exists('products', 'id')->whereNull('parent_id'),
```
SirSplasch: add the where() to the ModelExists rule.

src/Rulesets/Product/ResetProductFieldRuleset.php
```diff
@@ -0,0 +1,30 @@
...
+    protected static ?string $model = Product::class;
+
+    public function rules(): array
+    {
+        return [
+            'id' => [
+                'required',
+                'integer',
+                app(ModelExists::class, ['model' => Product::class]),
+                Rule::exists('products', 'id')->whereNotNull('parent_id'),
```
SirSplasch: add the where() to the ModelExists rule.

src/Rulesets/Product/ResetProductRelationRuleset.php
```diff
@@ -0,0 +1,31 @@
...
+    protected static ?string $model = Product::class;
+
+    public function rules(): array
+    {
+        return [
+            'id' => [
+                'required',
+                'integer',
+                app(ModelExists::class, ['model' => Product::class]),
+                Rule::exists('products', 'id')->whereNotNull('parent_id'),
```
SirSplasch: add the where() to the ModelExists rule.

src/Rulesets/Product/ResetRelationOnAllVariantsRuleset.php
```diff
@@ -0,0 +1,31 @@
...
+    protected static ?string $model = Product::class;
+
+    public function rules(): array
+    {
+        return [
+            'parent_id' => [
+                'required',
+                'integer',
+                app(ModelExists::class, ['model' => Product::class]),
+                Rule::exists('products', 'id')->whereNull('parent_id'),
```
SirSplasch: add the where() to the ModelExists rule.

src/Traits/Model/InheritsFromParent.php
```diff
@@ -0,0 +1,318 @@
...
+trait InheritsFromParent
+{
+    public function isVariant(): bool
+    {
+        return ! is_null($this->{$this->getParentKeyAttribute()});
+    }
+
+    public function inheritanceEnabled(): bool
+    {
+        return ProductSettings::variantInheritanceEnabled();
```
SirSplasch: app(ProductSettings::class)->variant_inheritance_enabled

src/Traits/Model/InheritsFromParent.php
SirSplasch, line 1: I don't know why this trait holds so much logic. As far as I can see, some of these methods are only called from an action, so why is this logic not handled there?

src/Actions/Product/ResetProductField.php
SirSplasch, line 1: These actions are mere skeletons, effectively not handling any logic themselves. I don't think this is how actions should be used, as in my understanding it is the actions which are holding it together and that it is them that should be called, and not the other way round.

tests/Feature/Actions/Product/Variant/CreateVariantsTest.php
SirSplasch, line 1: The return of the 'it' test methods. Adjust ALL tests

SirSplasch, line 1: Why do you need to fill the product settings, if the database record for this setting is generated in the setup?
Affects multiple tests. (Not just this test class)

tests/Unit/Models/ProductInheritanceColumnsTest.php
SirSplasch, line 1: What is this test, check if a migration has been run "correctly"? In my mind this can be removed.

tests/Unit/Models/VariantRelationColumnsTest.php
SirSplasch, line 1: Same here, check if a migration has run "correctly"?

tests/Unit/Support/VariantInheritance/InheritanceSyncTest.php
SirSplasch, line 1: remove

tests/Feature/Actions/Product/Variant/CreateVariantsTest.php
SirSplasch, line 1: Check what tests are "outdated" or need to be adjusted, as I suppose they were created the very first time, but some changes were afterwards.

src/Models/Product.php
```diff
@@ -262,6 +339,42 @@ public function registerMediaCollections(): void
     }
 
     // Attributes
+    public function getCategoriesAttribute(): Collection
```
SirSplasch: I don't know about these attributes, I would prefer a relation, that can be queried rather than a collection.

src/Jobs/SyncVariantInheritanceJob.php
```diff
@@ -0,0 +1,143 @@
...
+    public function __construct(public int $parentId, public array $fields = [], public ?array $variantIds = null) {}
+
+    public function middleware(): array
+    {
+        return [new WithoutOverlapping('variant-sync:' . $this->parentId)];
+    }
+
+    public function handle(): void
+    {
+        if (! ProductSettings::variantInheritanceEnabled()) {
```
SirSplasch: app(ProductSettings::class)->variant_inheritance_enabled

src/Jobs/SyncVariantInheritanceJob.php
```diff
@@ -0,0 +1,143 @@
...
+            return;
+        }
+
+        /** @var Product|null $parent */
+        $parent = resolve_static(Product::class, 'query')->whereKey($this->parentId)->first();
```
SirSplasch: wrap

src/Jobs/SyncVariantInheritanceJob.php
SirSplasch, line 1: No single character variable names.

src/Jobs/SyncVariantInheritanceJob.php
```diff
@@ -0,0 +1,143 @@
...
+        if (! $parent) {
+            return;
+        }
+
+        $fields = $this->fields ?: $parent->getInheritableFields();
+
+        foreach ($fields as $field) {
+            resolve_static(Product::class, 'query')
+                ->where('parent_id', $this->parentId)
+                ->when($this->variantIds, fn ($q) => $q->whereIn('id', $this->variantIds))
```
SirSplasch: add Builder type to callback variable (affects multiple callbacks)

src/Jobs/SyncVariantInheritanceJob.php
```diff
@@ -0,0 +1,143 @@
...
+        // Query-builder updates bypass Scout's ModelObserver — reindex explicitly.
+        resolve_static(Product::class, 'query')
+            ->where('parent_id', $this->parentId)
+            ->when($this->variantIds, fn ($q) => $q->whereIn('id', $this->variantIds))
+            ->searchable();
+    }
+
+    protected function syncTranslations(Product $parent, array $fields): void
+    {
+        $fields = array_intersect($fields, Product::getTranslatableAttributes());
```
SirSplasch: resolve_static

src/Jobs/SyncVariantInheritanceJob.php
SirSplasch: whereIntegerInRaw()

src/Jobs/SyncVariantInheritanceJob.php
```diff
@@ -0,0 +1,143 @@
...
+        if (! $fields) {
+            return;
+        }
+
+        $modelType = $parent->getMorphClass();
+
+        foreach ($fields as $field) {
+            $childIds = resolve_static(Product::class, 'query')
+                ->where('parent_id', $parent->getKey())
+                ->when($this->variantIds, fn ($q) => $q->whereIn('id', $this->variantIds))
```
SirSplasch: whereIntegerInRaw()

src/Jobs/SyncVariantInheritanceJob.php
```diff
@@ -0,0 +1,143 @@
...
+                ->where('model_type', $modelType)
+                ->where('model_id', $parent->getKey())
+                ->where('attribute', $field)
+                ->whereNull('deleted_at')
+                ->pluck('value', 'language_id');
+
+            // Drop child rows for locales the parent no longer has (all of them, if none are left).
+            DB::table('attribute_translations')
+                ->where('model_type', $modelType)
+                ->whereIn('model_id', $childIds)
```
SirSplasch: whereIntegerInRaw()

src/Jobs/SyncVariantInheritanceJob.php
```diff
@@ -0,0 +1,143 @@
...
+                ->where('attribute', $field)
+                ->whereNull('deleted_at')
+                ->when(
+                    $parentTranslations->isNotEmpty(),
+                    fn ($query) => $query->whereNotIn('language_id', $parentTranslations->keys()),
```
SirSplasch: whereIntegerNotInRaw()

src/Jobs/SyncVariantInheritanceJob.php
```diff
@@ -0,0 +1,143 @@
...
+
+            $parentTranslations = DB::table('attribute_translations')
+                ->where('model_type', $modelType)
+                ->where('model_id', $parent->getKey())
+                ->where('attribute', $field)
+                ->whereNull('deleted_at')
+                ->pluck('value', 'language_id');
+
+            // Drop child rows for locales the parent no longer has (all of them, if none are left).
+            DB::table('attribute_translations')
```
SirSplasch: Why raw query the table instead of utilizing the AttributeTranslation model?
Affects all of the following queries.

src/Jobs/SyncVariantInheritanceJob.php
```diff
@@ -0,0 +1,143 @@
...
+                    fn ($query) => $query->whereNotIn('language_id', $parentTranslations->keys()),
+                )
+                ->update(['deleted_at' => now()]);
+
+            foreach ($parentTranslations as $languageId => $value) {
+                $existingChildIds = DB::table('attribute_translations')
+                    ->where('model_type', $modelType)
+                    ->whereIn('model_id', $childIds)
+                    ->where('attribute', $field)
+                    ->where('language_id', $languageId)
```
SirSplasch: language_id before model_type to keep table ordering

src/Jobs/SyncVariantInheritanceJob.php
```diff
@@ -0,0 +1,143 @@
...
+                    ->whereNull('deleted_at')
+                    ->pluck('model_id');
+
+                if ($existingChildIds->isNotEmpty()) {
+                    DB::table('attribute_translations')
+                        ->where('model_type', $modelType)
+                        ->whereIn('model_id', $existingChildIds)
+                        ->where('attribute', $field)
+                        ->where('language_id', $languageId)
```
SirSplasch: language_id before model_type to keep table ordering

src/Jobs/SyncVariantInheritanceJob.php
```diff
@@ -0,0 +1,143 @@
...
+                ->when(
+                    $parentTranslations->isNotEmpty(),
+                    fn ($query) => $query->whereNotIn('language_id', $parentTranslations->keys()),
+                )
+                ->update(['deleted_at' => now()]);
+
+            foreach ($parentTranslations as $languageId => $value) {
+                $existingChildIds = DB::table('attribute_translations')
+                    ->where('model_type', $modelType)
+                    ->whereIn('model_id', $childIds)
```
SirSplasch: whereIntegerInRaw()

src/Jobs/SyncVariantInheritanceJob.php
```diff
@@ -0,0 +1,143 @@
...
+                    ->where('attribute', $field)
+                    ->where('language_id', $languageId)
+                    ->whereNull('deleted_at')
+                    ->pluck('model_id');
+
+                if ($existingChildIds->isNotEmpty()) {
+                    DB::table('attribute_translations')
+                        ->where('model_type', $modelType)
+                        ->whereIn('model_id', $existingChildIds)
```
SirSplasch: whereIntegerInRaw()

src/Jobs/MigrateProductVariantInheritanceJob.php
```diff
@@ -0,0 +1,78 @@
...
+    use Queueable;
+    use SerializesModels;
+
+    public int $batchSize = 500;
+
+    public function __construct(public ?int $parentId = null) {}
+
+    public function handle(): void
+    {
+        if (! ProductSettings::variantInheritanceEnabled()) {
```
SirSplasch: app(ProductSettings::class)->variant_inheritance_enabled

src/Jobs/MigrateProductVariantInheritanceJob.php
```diff
@@ -0,0 +1,78 @@
...
+            'parent_id' => $this->parentId,
+            'processed' => $processed,
+        ]);
+    }
+
+    protected function materialize(Product $parent): void
+    {
+        // Copies parent field values (+ translations) into non-overriding children and
+        // reindexes them in Scout.
+        (new SyncVariantInheritanceJob($parent->getKey()))->handle();
```
SirSplasch: app()

src/Jobs/MigrateProductVariantInheritanceJob.php
```diff
@@ -0,0 +1,78 @@
...
+
+        // Seeds/refreshes is_inherited=true pivot rows for non-owning children.
+        PivotInheritanceSync::propagateToChildren($parent);
```
SirSplasch: resolve_static()

src/Actions/Product/UpdateProduct.php
```diff
@@ -160,6 +193,12 @@ protected function validateData(): void
                     'parent_id' => ['Cycle detected'],
                 ])->errorBag('updateProduct');
             }
+
+            if ($product?->children()->exists()) {
+                throw ValidationException::withMessages([
+                    'parent_id' => [__('A product with existing variants cannot itself become a variant.')],
+                ])->errorBag('updateProduct');
```
SirSplasch: wrap

src/Actions/Product/UpdateProduct.php
```diff
@@ -160,6 +193,12 @@ protected function validateData(): void
                     'parent_id' => ['Cycle detected'],
                 ])->errorBag('updateProduct');
             }
+
+            if ($product?->children()->exists()) {
+                throw ValidationException::withMessages([
+                    'parent_id' => [__('A product with existing variants cannot itself become a variant.')],
```
SirSplasch: do not localize this string here, but add the key to de.json

src/Actions/Product/UpdateProduct.php
SirSplasch, line 1: Call PivotInheritanceSync via resolve_static(). This class currently is never called via resolve_static() which is a customization issue, adjust all other calls, not just in this class.

src/Models/Price.php
```diff
@@ -55,6 +55,113 @@ class Price extends FluxModel
...
+
+            if (empty($childIds)) {
+                return;
+            }
+
+            $childIds = collect($childIds);
+
+            $rawPrice = $price->getAttributes()['price'];
+
+            $owningChildIds = DB::table('prices')
```
SirSplasch: Why are the prices queried with raw DB queries rather than via the model?

src/Support/VariantInheritance/PivotInheritanceSync.php
```diff
@@ -0,0 +1,187 @@
...
+ * (categories, suppliers, productProperties) as is_inherited=true pivot rows.
+ *
+ * Pivot sync()/attach() does not reliably fire per-row pivot model events, so this can't
+ * hook into pivot model booted() the way Price does. Instead it must be invoked explicitly
+ * from the action choke point (UpdateProduct::performAction) right after the parent's own
+ * pivots are synced.
+ */
+class PivotInheritanceSync
+{
+    protected const RELATIONS = ['categories', 'suppliers', 'productProperties'];
```
SirSplasch: I think this is a huge issue, as these relations are "hardcoded", but the inheritable relations in the product can be adjusted and then do not match with these.

src/Support/VariantInheritance/PivotInheritanceSync.php
```diff
@@ -0,0 +1,187 @@
...
+    public static function syncOwned(BelongsToMany $relation, array $desired, bool $takesOwnership): void
+    {
+        if (! $takesOwnership) {
+            $relation->sync($desired);
+
+            return;
+        }
+
+        $relation->sync(
+            collect($desired)->map(fn (array $attributes) => $attributes + ['is_inherited' => false])->all(),
```
SirSplasch: wrap

SirSplasch: $attributes + [] can be an issue if the key already exists in $attributes, so better use array_merge()

src/Support/VariantInheritance/PivotInheritanceSync.php
```diff
@@ -0,0 +1,187 @@
...
+            return;
+        }
+
+        $childIds = $parent->children()->pluck('id');
+
+        if ($childIds->isEmpty()) {
+            return;
+        }
+
+        foreach (self::RELATIONS as $relation) {
```
SirSplasch: Do not use self:: better static::, but this property in itself has issues as I mentioned beforehand.

src/Support/VariantInheritance/PivotInheritanceSync.php
```diff
@@ -0,0 +1,187 @@
...
+            }
+        }
+    }
+
+    protected static function propagateRelation(Product $parent, string $relation, Collection $childIds): void
+    {
+        $rel = $parent->{'own' . ucfirst($relation)}();
+
+        if (! $rel instanceof BelongsToMany) {
+            throw new LogicException("Unsupported pivot relation type for propagation: [$relation].");
```
SirSplasch: use single quotes

src/Support/VariantInheritance/PivotInheritanceSync.php
```diff
@@ -0,0 +1,187 @@
...
+        foreach (self::RELATIONS as $relation) {
+            if ($parent->isInheritableRelation($relation)) {
+                static::propagateRelation($parent, $relation, $childIds);
+            }
+        }
+    }
+
+    protected static function propagateRelation(Product $parent, string $relation, Collection $childIds): void
+    {
+        $rel = $parent->{'own' . ucfirst($relation)}();
```
SirSplasch: $relation

src/Support/VariantInheritance/PivotInheritanceSync.php
```diff
@@ -0,0 +1,187 @@
...
+/**
+ * Materializes parent -> child propagation for Product's inheritable pivot relations
+ * (categories, suppliers, productProperties) as is_inherited=true pivot rows.
+ *
+ * Pivot sync()/attach() does not reliably fire per-row pivot model events, so this can't
+ * hook into pivot model booted() the way Price does. Instead it must be invoked explicitly
+ * from the action choke point (UpdateProduct::performAction) right after the parent's own
+ * pivots are synced.
+ */
+class PivotInheritanceSync
```
SirSplasch, line 21: Please double check yourself, if all these methods work for MorphToMany relations, as these are also BelongsToMany relations.

src/Support/VariantInheritance/PivotInheritanceSync.php
```diff
@@ -0,0 +1,187 @@
...
+                        foreach ($extraColumns as $column) {
+                            $row[$column] = $parentRow->{$column};
+                        }
+
+                        if ($isMorph) {
+                            $row[$morphType] = $morphClass;
+                        }
+
+                        return $row;
+                    })->all()
```
SirSplasch: wrap ->all()

src/Support/VariantInheritance/PivotInheritanceSync.php
```diff
@@ -0,0 +1,187 @@
...
+                    ->where($relatedPivotKey, $key)
+                    ->where('is_inherited', true)
+                    ->update(collect($extraColumns)
+                        ->mapWithKeys(fn ($column) => [$column => $parentRow->{$column}])
+                        ->all());
+            }
+
+            if ($toInsertIds->isNotEmpty()) {
+                DB::table($pivotTable)->insert(
+                    $toInsertIds->map(function ($childId) use (
```
SirSplasch: wrap ->map()

src/Actions/Product/Variant/CreateVariants.php
```diff
@@ -38,12 +39,34 @@ public function performAction(): Collection
...
+            $product['own_prices'],
         );
         $product['parent_id'] = $parentProduct->id;
         $product['tenants'] = $parentProduct->tenants->pluck('id')->toArray();
-        $product['categories'] = $parentProduct->categories?->pluck('id')->toArray();
+        // Tags have no is_inherited concept (no such column on the taggables table), so
+        // they are always plainly copied as owned, regardless of the inheritance toggle.
         $product['tags'] = $parentProduct->tags?->pluck('id')->toArray();
 
+        $inheritanceEnabled = app(Product::class)->inheritanceEnabled();
```
SirSplasch: app(ProductSettings::class)->variant_inheritance_enabled

src/Actions/Product/ResetFieldOnAllVariants.php
```diff
@@ -0,0 +1,39 @@
...
+            ->whereKey($this->getData('parent_id'))
+            ->firstOrFail();
+
+        $field = $this->getData('field');
+        $touched = $parent->resetFieldOnAllVariants($field);
+
+        if ($touched > 0) {
+            // Re-copy the parent's current value (and translations) onto the now
+            // un-overridden variants via the same set-based sync the job already uses.
+            SyncVariantInheritanceJob::dispatchSync($parent->getKey(), [$field]);
```
SirSplasch: app()

src/Jobs/SyncVariantInheritanceJob.php
SirSplasch, line 1: Was this a conscious decision to have the logic in this job, rather than in an action? An action can be dispatched as well. Like in the ResetField... action.
If you haven't gotten it yet, I'm kinda unhappy on where parts of the logic is situated.

src/Actions/Product/ResetProductField.php
```diff
@@ -0,0 +1,39 @@
...
+            ->firstOrFail();
+
+        $field = $this->getData('field');
+
+        $variant->resetField($field);
+        $variant->save();
+
+        // Re-copy the parent's current value (and translations) onto the variant now
+        // that it's no longer flagged as overriding — otherwise the column stays stale.
+        SyncVariantInheritanceJob::dispatchSync($variant->parent_id, [$field], [$variant->getKey()]);
```
SirSplasch: app()

src/Models/Product.php
```diff
@@ -52,13 +56,77 @@ class Product extends FluxModel implements HasMedia, HasMediaForeignKey, Interac
...
+                return;
+            }
+
+            $fields = InheritanceSync::changedInheritableFields($product);
+
+            if ($fields === [] || ! $product->children()->exists()) {
+                return;
+            }
+
+            SyncVariantInheritanceJob::dispatch($product->getKey(), $fields)->afterCommit();
```
SirSplasch: app()

SirSplasch (changes requested):
I will be frank, I have not gotten to review all of the logic, like the stuff that actually does things. I will review the rest eventually (maybe after these annotations are resolved?).

src/Models/Price.php
```diff
@@ -55,6 +55,113 @@ class Price extends FluxModel
...
+        });
+    }
+
+    /**
+     * Resolve the child product ids that should receive inherited-price propagation
+     * for the given parent price, or an empty array if any guard fails.
+     */
+    protected static function inheritableChildIdsForParentPrice(Price $price): array
+    {
+        if ($price->is_inherited || ! ProductSettings::variantInheritanceEnabled()) {
```
SirSplasch: app(ProductSettings::class)->variant_inheritance_enabled

src/Models/Price.php
```diff
@@ -55,6 +55,113 @@ class Price extends FluxModel
...
+            return [];
+        }
+
+        $product = resolve_static(Product::class, 'query')
+            ->whereKey($price->product_id)
```
SirSplasch: Add (keep table column ordering in mind)
 ```
->whereNull('parent_id')
->whereHas('children')
```
Then you can immediately after
`return $product?->children()->pluck('id) ?? [];`

File: src/Models/Price.php [hunk 57]
```diff
@@ -55,6 +55,113 @@ class Price extends FluxModel
...
+            ->first();
```
SirSplasch: Only select needed columns, not everything.

src/Models/Price.php
SirSplasch: All these queries keep table column ordering for where clauses

src/Models/Price.php
```diff
@@ -55,6 +55,113 @@ class Price extends FluxModel
...
+
+            $childIds = collect($childIds);
+
+            $rawPrice = $price->getAttributes()['price'];
+
+            $owningChildIds = DB::table('prices')
+                ->where('price_list_id', $price->price_list_id)
+                ->where('is_inherited', false)
+                ->whereNull('deleted_at')
+                ->whereIn('product_id', $childIds)
```
SirSplasch: update all whereIn('product_id') to whereIntegerInRaw('product_id')

src/Models/Price.php
```diff
@@ -55,6 +55,113 @@ class Price extends FluxModel
...
+                DB::table('prices')->insert(
+                    $missingChildIds->map(fn ($childId) => [
+                        'uuid' => Str::uuid()->toString(),
+                        'product_id' => $childId,
+                        'price_list_id' => $price->price_list_id,
+                        'price' => $rawPrice,
+                        'is_inherited' => true,
+                        'created_at' => $now,
+                        'updated_at' => $now,
+                    ])->all()
```
SirSplasch: wrap

src/Models/Price.php
```diff
@@ -55,6 +55,113 @@ class Price extends FluxModel
...
+                ->whereIn('product_id', $childIds)
+                ->update(['deleted_at' => now()]);
+        });
+    }
+
+    /**
+     * Resolve the child product ids that should receive inherited-price propagation
+     * for the given parent price, or an empty array if any guard fails.
+     */
+    protected static function inheritableChildIdsForParentPrice(Price $price): array
```
SirSplasch: Is this static by choice, so that it can be protected?
If the visibility of this method is not an issue this can be changed to a public function as this is only called on a price model.

SirSplasch (changes requested):

SirSplasch:
I have not specifically looked onto this, therefore I just add a comment on this topic.
Check the queries that retrieve something, that they only select, what is really needed, not just everything.
It may never be an issue, expect the one time I actually annotated it, but please keep an eye on it during your adjustments.

src/Rulesets/Product/ResetProductFieldsRuleset.php
```diff
@@ -0,0 +1,36 @@
...
+{
+    protected static ?string $model = Product::class;
+
+    public function rules(): array
+    {
+        return [
+            'parent_id' => [
+                'required',
+                'integer',
+                app(ModelExists::class, ['model' => Product::class])->whereNull('parent_id'),
```
SirSplasch: wrap ->whereNull()

src/Rulesets/Product/ResetProductFieldsRuleset.php
```diff
@@ -0,0 +1,36 @@
...
+            ],
+            'fields' => 'required|array|min:1',
+            'fields.*' => [
+                'required',
+                'string',
+                Rule::in(app(Product::class)->getInheritableFields()),
+            ],
+            'variant_ids' => 'array|nullable',
+            'variant_ids.*' => [
```
SirSplasch, line 30: Please validate if the variants match the parent, currently they are silently ignored.
Either directly in the Ruleset, or in the corresponding action via the validateData method.

src/Rulesets/Product/ResetProductFieldsRuleset.php
```diff
@@ -0,0 +1,36 @@
...
+                'required',
+                'integer',
+                app(ModelExists::class, ['model' => Product::class])->whereNotNull('parent_id'),
```
SirSplasch: wrap
->whereNotNull()

src/Rulesets/Product/ResetProductRelationsRuleset.php
```diff
@@ -0,0 +1,37 @@
...
+            ],
+            'relations' => 'required|array|min:1',
+            'relations.*.relation' => [
+                'required',
+                'string',
+                Rule::in(app(Product::class)->getInheritableRelations()),
+            ],
+            'relations.*.related_id' => 'integer|nullable',
+            'variant_ids' => 'array|nullable',
+            'variant_ids.*' => [
```
SirSplasch, line 32: Please validate if the variants match the parent, currently they are silently ignored.
Either directly in the Ruleset, or in the corresponding action via the validateData method.

src/Rulesets/Product/ResetProductRelationsRuleset.php
```diff
@@ -0,0 +1,37 @@
...
+{
+    protected static ?string $model = Product::class;
+
+    public function rules(): array
+    {
+        return [
+            'parent_id' => [
+                'required',
+                'integer',
+                app(ModelExists::class, ['model' => Product::class])->whereNull('parent_id'),
```
SirSplasch: wrap
->whereNull()

src/Rulesets/Product/ResetProductRelationsRuleset.php
```diff
@@ -0,0 +1,37 @@
...
+                'required',
+                'string',
+                Rule::in(app(Product::class)->getInheritableRelations()),
+            ],
+            'relations.*.related_id' => 'integer|nullable',
+            'variant_ids' => 'array|nullable',
+            'variant_ids.*' => [
+                'required',
+                'integer',
+                app(ModelExists::class, ['model' => Product::class])->whereNotNull('parent_id'),
```
SirSplasch: wrap
->whereNotNull()

src/Actions/Product/ResetProductRelations.php
```diff
@@ -0,0 +1,131 @@
...
+use FluxErp\Actions\FluxAction;
+use FluxErp\Models\Product;
+use FluxErp\Rulesets\Product\ResetProductRelationsRuleset;
+use FluxErp\Support\VariantInheritance\PivotInheritanceSync;
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\Relations\BelongsToMany;
+use Illuminate\Database\Eloquent\Relations\HasMany;
+use Illuminate\Database\Eloquent\Relations\MorphToMany;
+use Illuminate\Database\Query\Builder as QueryBuilder;
+use LogicException;
```
SirSplasch: I prefer not to throw a logic exception, as they are currently not being caught in most cases.

src/Rulesets/Product/ResetProductRelationsRuleset.php
```diff
@@ -0,0 +1,37 @@
...
+                'integer',
+                app(ModelExists::class, ['model' => Product::class])->whereNull('parent_id'),
+            ],
+            'relations' => 'required|array|min:1',
+            'relations.*.relation' => [
+                'required',
+                'string',
+                Rule::in(app(Product::class)->getInheritableRelations()),
+            ],
+            'relations.*.related_id' => 'integer|nullable',
```
SirSplasch: Please validate if these relations exist, rather than to throw a LogicException in the action. (Either throught the ruleset or the validateData method of the action.

src/Listeners/Product/DispatchVariantInheritanceMigration.php
```diff
@@ -0,0 +1,32 @@
...
+    {
+        if (! $event->settings instanceof ProductSettings) {
+            return;
+        }
+
+        $wasEnabled = (bool) data_get($event->originalValues, 'variant_inheritance_enabled', false);
+        $isEnabled = (bool) $event->properties->get('variant_inheritance_enabled');
+
+        if (! $wasEnabled && $isEnabled) {
+            MigrateProductVariantInheritanceJob::dispatch()
```
SirSplasch: Allow customization for this job either via resolve_static() or app().

src/Listeners/Product/DispatchVariantInheritanceMigration.php
```diff
@@ -0,0 +1,32 @@
...
+ */
+class DispatchVariantInheritanceMigration
+{
+    public function handle(SavingSettings $event): void
+    {
+        if (! $event->settings instanceof ProductSettings) {
+            return;
+        }
+
+        $wasEnabled = (bool) data_get($event->originalValues, 'variant_inheritance_enabled', false);
```
SirSplasch: No fallback needed, if it does not exist it returns null and then gets cast into false.

src/Jobs/MigrateProductVariantInheritanceJob.php
```diff
@@ -0,0 +1,82 @@
...
+ *
+ * For every parent product, copies the parent's current inheritable field values (+
+ * translations) into non-overriding child columns and seeds `is_inherited = true`
+ * relation rows (prices + categories/suppliers/productProperties) for non-owning
+ * children. Reuses the same propagation helpers the live write path uses, so this is
+ * idempotent — running it again is a no-op for already-materialized data.
+ */
+class MigrateProductVariantInheritanceJob implements ShouldQueue
+{
+    use Dispatchable;
```
SirSplasch: Is this a conscious choice to not use a comma separated list for the traits?

src/Actions/Product/UpdateProduct.php
```diff
@@ -160,6 +210,15 @@ protected function validateData(): void
                     'parent_id' => ['Cycle detected'],
                 ])->errorBag('updateProduct');
             }
+
+            if ($product?->children()->exists()) {
+                throw ValidationException::withMessages([
+                    'parent_id' => [
+                        __('A product with existing variants cannot itself become a variant.'),
```
SirSplasch: Do not use __() here, this will be localized by the toast in the frontend, and should not be localized for the api.

src/Actions/Product/UpdateProduct.php
```diff
@@ -150,6 +187,19 @@ protected function validateData(): void
...
 
+        if (array_key_exists('is_variant_parent', $this->data)
+            && ! $this->getData('is_variant_parent')
+            && resolve_static(Product::class, 'query')
+                ->whereKey($this->getData('id'))
+                ->whereHas('children', fn (Builder $query) => $query->where('is_active', true))
+                ->exists()
+        ) {
+            throw ValidationException::withMessages([
+                'is_variant_parent' => [__('Cannot demote to standalone: active variants still exist.')],
```
SirSplasch: Do not use __() here, this will be localized by the toast in the frontend, and should not be localized for the api.

src/Actions/Product/SyncVariantInheritance.php
```diff
@@ -0,0 +1,148 @@
...
+        }
+    }
+
+    protected function variantQuery(Product $parent): Builder
+    {
+        return resolve_static(Product::class, 'query')
+            ->where('parent_id', $parent->getKey())
+            ->when(
+                $this->getData('variant_ids'),
+                fn (Builder $query, array $variantIds) => $query->whereIntegerInRaw('id', $variantIds)
```
SirSplasch: whereKey()? Or is the whereIntegerInRaw('id') intended?

src/Actions/Product/SyncVariantInheritance.php
SirSplasch, line 1: Does this action need a permission?
I can see both worlds, just want to spark some thought about that to this specific action.

src/Support/VariantInheritance/PivotInheritanceSync.php
```diff
@@ -0,0 +1,185 @@
...
+     * A non-variant (or non-inheriting) product keeps the plain sync() behavior: the
+     * payload is the full desired set, anything absent gets detached.
+     *
+     * A variant taking ownership must not let payload omission delete its inherited
+     * copies — those are materialized by parent propagation and removed only by the
+     * reset actions, never by omission. So only rows the variant already owns and no
+     * longer lists get detached; every listed entry becomes owned (is_inherited =
+     * false), flipping an existing inherited copy in place instead of duplicating it.
+     */
+    public static function syncOwned(BelongsToMany $relation, array $desired, bool $takesOwnership): void
```
SirSplasch, line 36: Have I seen this correctly, is this method only called once in the categorizable trait?
I am just on a quest to get rid of this class, but I understand its purpose, and I currently do not really have a better solution on where to put this code.

src/Models/Price.php
```diff
@@ -55,6 +53,86 @@ class Price extends FluxModel
...
+    {
+        static::saved(function (Price $price): void {
+            // A variant taking ownership through a direct price write supersedes its
+            // materialized inherited copy for the same price list.
+            if (! $price->is_inherited) {
+                resolve_static(static::class, 'query')
+                    ->where('price_list_id', $price->price_list_id)
+                    ->where('product_id', $price->product_id)
+                    ->where('is_inherited', true)
+                    ->whereKeyNot($price->getKey())
```
SirSplasch: whereKeyNot() as first constraint.

src/Models/Price.php
```diff
@@ -67,6 +145,25 @@ protected function casts(): array
         ];
     }
 
+    /**
+     * Resolve the child product ids that should receive inherited-price propagation
+     * for this price, or an empty array if any guard fails.
+     */
+    public function inheritableChildIds(): array
```
SirSplasch: public methods after relations

src/Actions/Product/SyncVariantInheritance.php
SirSplasch, line 1: I am curious why this action handles only updates to variant columns/fields, and not to the relations.
I saw a couple times that the relations handling was called separately, but I do not really understand why.

src/Rulesets/Product/SyncVariantInheritanceRuleset.php
```diff
@@ -0,0 +1,36 @@
...
+{
+    protected static ?string $model = Product::class;
+
+    public function rules(): array
+    {
+        return [
+            'parent_id' => [
+                'required',
+                'integer',
+                app(ModelExists::class, ['model' => Product::class])->whereNull('parent_id'),
```
SirSplasch: wrap
->whereNull()

src/Rulesets/Product/SyncVariantInheritanceRuleset.php
```diff
@@ -0,0 +1,36 @@
...
+            'fields.*' => [
+                'required',
+                'string',
+                Rule::in(app(Product::class)->getInheritableFields()),
+            ],
+            'variant_ids' => 'array|nullable',
+            'variant_ids.*' => [
+                'required',
+                'integer',
+                app(ModelExists::class, ['model' => Product::class])->whereNotNull('parent_id'),
```
SirSplasch: wrap
->whereNotNull()

src/Rulesets/Product/SyncVariantInheritanceRuleset.php
```diff
@@ -0,0 +1,36 @@
...
+                app(ModelExists::class, ['model' => Product::class])->whereNull('parent_id'),
+            ],
+            'fields' => 'array|nullable',
+            'fields.*' => [
+                'required',
+                'string',
+                Rule::in(app(Product::class)->getInheritableFields()),
+            ],
+            'variant_ids' => 'array|nullable',
+            'variant_ids.*' => [
```
SirSplasch, line 30: Please validate if the variants match the parent, currently they are silently ignored.
Either directly in the Ruleset, or in the corresponding action via the validateData method.

src/Support/VariantInheritance/PivotInheritanceSync.php
```diff
@@ -0,0 +1,185 @@
...
+
+        $parentKeys = $parentRows->keys();
+
+        // Drop inherited child copies for related keys the parent no longer owns.
+        $scoped()
+            ->whereIntegerInRaw($foreignPivotKey, $childIds)
+            ->where('is_inherited', true)
+            ->when(
+                $parentKeys->isNotEmpty(),
+                fn (QueryBuilder $query) => $query->whereNotIn($relatedPivotKey, $parentKeys),
```
SirSplasch: whereIntegerNotInRaw()

src/Support/VariantInheritance/PivotInheritanceSync.php
```diff
@@ -0,0 +1,185 @@
...
+            ->delete();
+
+        if ($parentKeys->isEmpty()) {
+            return;
+        }
+
+        $ownedByKey = $scoped()
+            ->whereIntegerInRaw($foreignPivotKey, $childIds)
+            ->where('is_inherited', false)
+            ->whereIntegerInRaw($relatedPivotKey, $parentKeys)
```
SirSplasch, line 117: put this constraint above ->where('is_inherited')

src/Support/VariantInheritance/PivotInheritanceSync.php
```diff
@@ -0,0 +1,185 @@
...
+            ->get([$foreignPivotKey, $relatedPivotKey])
+            ->groupBy($relatedPivotKey)
+            ->map(fn (Collection $rows) => $rows->pluck($foreignPivotKey));
+
+        $existingByKey = $scoped()
+            ->whereIntegerInRaw($foreignPivotKey, $childIds)
+            ->where('is_inherited', true)
+            ->whereIntegerInRaw($relatedPivotKey, $parentKeys)
```
SirSplasch, line 125: put this constraint above ->where('is_inherited')

src/Support/VariantInheritance/PivotInheritanceSync.php
```diff
@@ -0,0 +1,185 @@
...
+                    ->where($relatedPivotKey, $key)
+                    ->where('is_inherited', true)
+                    ->update(collect($extraColumns)
+                        ->mapWithKeys(fn (string $column) => [$column => $parentRow->{$column}])
+                        ->all());
+            }
+
+            if ($toInsertIds->isNotEmpty()) {
+                DB::table($pivotTable)->insert(
+                    $toInsertIds->map(function ($childId) use (
```
SirSplasch: wrap
->map()

src/Support/VariantInheritance/PivotInheritanceSync.php
```diff
@@ -0,0 +1,185 @@
...
+                        foreach ($extraColumns as $column) {
+                            $row[$column] = $parentRow->{$column};
+                        }
+
+                        if ($isMorph) {
+                            $row[$morphType] = $morphClass;
+                        }
+
+                        return $row;
+                    })->all()
```
SirSplasch: wrap
->all()

src/Actions/Product/ResetProductFields.php
```diff
@@ -0,0 +1,75 @@
...
+
+        if ($variants->isNotEmpty()) {
+            // Group by the resulting shape so we issue one UPDATE per distinct JSON value.
+            $groups = $variants->groupBy(function (Product $variant) use ($fields): string {
+                $remaining = array_values(array_diff($variant->overridden_fields ?? [], $fields));
+
+                return $remaining === [] ? '__null__' : json_encode($remaining);
+            });
+
+            DB::transaction(function () use ($groups): void {
```
SirSplasch: The action already runs in a database transaction by default.

src/Actions/Product/ResetProductFields.php
```diff
@@ -0,0 +1,75 @@
...
+            ->whereKey($this->getData('parent_id'))
+            ->firstOrFail(['id']);
+
+        $fields = $this->getData('fields');
+        $variantIds = $this->getData('variant_ids');
+
+        $variants = $parent->children()
+            ->when(
+                $variantIds,
+                fn (Builder $query, array $ids) => $query->whereIntegerInRaw('id', $ids)
```
SirSplasch: I think whereKey() works here, or am I wrong?

src/Actions/Product/ResetProductRelations.php
```diff
@@ -0,0 +1,131 @@
...
+    public function performAction(): int
+    {
+        $parent = resolve_static(Product::class, 'query')
+            ->whereKey($this->getData('parent_id'))
+            ->firstOrFail(['id']);
+
+        $variantIds = $parent->children()
+            ->when(
+                $this->getData('variant_ids'),
+                fn (Builder $query, array $ids) => $query->whereIntegerInRaw('id', $ids)
```
SirSplasch: whereKey()?

src/Actions/Product/ResetProductRelations.php
```diff
@@ -0,0 +1,131 @@
...
+                ->clone()
+                ->distinct()
+                ->count($relationInstance->getForeignKeyName());
+
+            $query->delete();
+
+            return $touched;
+        }
+
+        throw new LogicException('Unsupported relation type for reset: [' . $relation . '].');
```
SirSplasch: validate relation type beforehand rather than to throw an exception during the operation.

src/Actions/Product/ResetProductRelations.php
```diff
@@ -0,0 +1,131 @@
...
+
+            resolve_static(PivotInheritanceSync::class, 'propagateToChildren', ['parent' => $parent]);
+        }
+
+        return $touched;
+    }
+
+    protected function resetRelation(Product $parent, array $variantIds, string $relation, mixed $relatedId): int
+    {
+        $relationInstance = resolve_static(Product::class, 'query')
```
SirSplasch: You are querying the same product over and over again in a foreach loop. Although it might not even be queried again, as the parent product already is queried, and this in only used to get the relation.

src/Actions/Product/ResetProductRelations.php
```diff
@@ -0,0 +1,131 @@
...
+            ->whereKey($variantIds[0])
+            ->first(['id'])
+            ->{'own' . ucfirst($relation)}();
+
+        if ($relationInstance instanceof BelongsToMany) {
```
SirSplasch, line 91: Can do if elseif, no need for a separate if, a relation cannot be a BelongsToMany and a HasMany relation.

src/Actions/Product/ResetProductRelations.php
```diff
@@ -0,0 +1,131 @@
...
+            $query = $relationInstance
+                ->newPivotStatement()
+                ->whereIn($relationInstance->getForeignPivotKeyName(), $variantIds)
```
SirSplasch: whereIntegerInRaw()

SirSplasch (changes requested):

src/Actions/Product/ResetProductRelations.php
```diff
@@ -0,0 +1,174 @@
...
+
+    protected function validateData(): void
+    {
+        parent::validateData();
+
+        $this->validateVariantParentage('resetProductRelations');
+
+        $product = app(Product::class);
+
+        foreach ($this->getData('relations') as $index => $reset) {
```
SirSplasch, line 144: What about pooling the validation messages and sending them all together, rather than throwing each one directly.

src/Actions/Product/ResetProductRelations.php
```diff
@@ -0,0 +1,174 @@
...
+    public function performAction(): int
+    {
+        $parent = resolve_static(Product::class, 'query')
+            ->whereKey($this->getData('parent_id'))
+            ->firstOrFail(['id']);
+
+        $variantIds = $parent->children()
+            ->when(
+                $this->getData('variant_ids'),
+                fn (Builder $query, array $ids) => $query->whereIntegerInRaw('id', $ids)
```
SirSplasch: whereKey()

src/Actions/Product/ResetProductRelations.php
```diff
@@ -0,0 +1,174 @@
...
+            // hook on the price model; pivot relations through the pivot sync.
+            $parent->ownPrices()
+                ->get()
+                ->each
+                ->save();
+
+            resolve_static(PivotInheritanceSync::class, 'propagateToChildren', ['parent' => $parent]);
+        }
+
+        return $touched;
```
SirSplasch: What is the purpose of this return value? I know it is the sum of all relations over all variants that have been reset, but how is that any helpful. Sure, you can work with 0 and not 0, but other than that, it does not help.

src/Actions/Product/ResetProductRelations.php
```diff
@@ -0,0 +1,174 @@
...
+            ->clone()
+            ->distinct()
+            ->count($relationInstance->getForeignKeyName());
+
+        $query->delete();
+
+        return $touched;
+    }
+
+    protected function resolveRelatedIdColumn(string $relation): string
```
SirSplasch: This method is only called once, so why not directly use that one-liner there?

src/Rulesets/Product/ResetProductRelationsRuleset.php
```diff
@@ -0,0 +1,39 @@
...
+                app(ModelExists::class, ['model' => Product::class])
+                    ->whereNull('parent_id'),
+            ],
+            'relations' => 'required|array|min:1',
+            'relations.*.relation' => [
+                'required',
+                'string',
+                Rule::in(app(Product::class)->getInheritableRelations()),
+            ],
+            'relations.*.related_id' => 'integer|nullable',
```
SirSplasch: Why is this restricted to only one integer value? Shouldn't it be possible to reset multiple related records?

src/Actions/Product/ResetProductRelations.php
```diff
@@ -0,0 +1,174 @@
...
+            if (! $relationInstance instanceof BelongsToMany && ! $relationInstance instanceof HasMany) {
+                throw ValidationException::withMessages([
+                    'relations.' . $index . '.relation' => [
+                        'Unsupported relation type for reset: [' . $relation . '].',
+                    ],
+                ])
+                    ->errorBag('resetProductRelations');
+            }
+
+            if (
```
SirSplasch, line 156: Is it intentional not to check for the existence of the related_id in the given relation?
Currently no existing related records are kinda silently ignored.
If this is intentional, maybe due to "unnecessary" validation/complexity, it is fine two me.

SirSplasch:
Also resolve merge conflicts

src/Rulesets/Product/ResetProductRelationsRuleset.php
patrickweh: Already possible. `relations` is an array and every entry carries its own `related_id`, so two price lists are two entries:

```php
'relations' => [
    ['relation' => 'prices', 'related_id' => 3],
    ['relation' => 'prices', 'related_id' => 7],
],
```

`performAction()` runs one reset per entry, `validateData()` reports every offending entry under its own index. A `related_ids` array next to it would be a second spelling of the same thing.

src/Actions/Product/ResetProductRelations.php
patrickweh: Intentional. `related_id` narrows the delete, it does not point at something that has to exist. An id that does not exist resets nothing, which is exactly what an id that exists but has no rows on these variants does, so nothing observable changes.

The check would cost one query per entry on the same table the action deletes from, the extra query you flagged in the two rule threads. What decides the scope is the parent and the variant ids, and both are validated.

SirSplasch (approved):

SirSplasch (approved):

PR #1704, opened 2026-05-08
Fix passkey auth losing options across intermediate requests

patrickweh:
## Summary

Spatie's `GeneratePasskeyAuthenticationOptionsAction` stores the WebAuthn challenge via `Session::flash`, which only survives one subsequent request. Any intermediate Livewire poll, asset load, or background fetch ages the flash out before the user's authenticate POST arrives, so the controller passes `null` into `FindPasskeyToAuthenticateAction` and the action throws a `TypeError` on its non-nullable `string $passkeyOptionsJson` argument.

```
Spatie\LaravelPasskeys\Actions\FindPasskeyToAuthenticateAction::execute():
Argument #2 ($passkeyOptionsJson) must be of type string, null given
```

## Fix

- Override the action to persist the options with `Session::put` instead of `Session::flash`.
- Override the controller to read them with `Session::pull`.

The atomic read-and-delete preserves the same single-use guarantee that the package's flash-based flow relied on, but no longer depends on Laravel's flash lifecycle, so the value survives any number of intermediate requests until the user authenticates.

## Security note

`pull()` is read-and-delete in one step. After a successful or failed authenticate request the challenge is gone from the session, so the same WebAuthn response cannot be replayed. This matches the security posture of the original `flash()` based flow.

## Summary by Sourcery

Ensure passkey authentication options persist across intermediate requests while retaining single-use semantics.

New Features:
- Add a custom GeneratePasskeyAuthenticationOptionsAction that stores WebAuthn authentication options in the session using persistent storage.

Bug Fixes:
- Prevent passkey authentication failures caused by flashed WebAuthn options expiring after intermediate requests by switching to persistent session storage with atomic retrieval.

Enhancements:
- Override the passkey authentication controller to atomically read and delete stored authentication options and handle missing or invalid options more robustly.

Tests:
- Add feature tests to verify that authentication options survive aged flash data and are removed from the session after being pulled to prevent replay.

### DIFF
--- a/src/FluxServiceProvider.php
+++ b/src/FluxServiceProvider.php
@@ -2,6 +2,7 @@
 
 namespace FluxErp;
 
+use FluxErp\Actions\Passkey\GeneratePasskeyAuthenticationOptionsAction;
 use FluxErp\Actions\Passkey\StorePasskeyAction;
 use FluxErp\Assets\AssetManager;
 use FluxErp\Facades\ProductType;
@@ -249,6 +250,7 @@ class FluxServiceProvider extends ServiceProvider
             config(['two-factor.recovery.enabled' => false]);
             config(['passkeys.models.authenticatable' => resolve_static(User::class, 'class')]);
             config(['passkeys.actions.store_passkey' => resolve_static(StorePasskeyAction::class, 'class')]);
+            config(['passkeys.actions.generate_passkey_authentication_options' => resolve_static(GeneratePasskeyAuthenticationOptionsAction::class, 'class')]);
         });
         $this->mergeConfigFrom(__DIR__ . '/../config/flux.php', 'flux');
         $this->mergeConfigFrom(__DIR__ . '/../config/notifications.php', 'notifications');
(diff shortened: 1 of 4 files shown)


src/FluxServiceProvider.php
```diff
@@ -249,6 +250,7 @@ protected function registerConfig(): void
             config(['two-factor.recovery.enabled' => false]);
             config(['passkeys.models.authenticatable' => resolve_static(User::class, 'class')]);
             config(['passkeys.actions.store_passkey' => resolve_static(StorePasskeyAction::class, 'class')]);
+            config(['passkeys.actions.generate_passkey_authentication_options' => resolve_static(GeneratePasskeyAuthenticationOptionsAction::class, 'class')]);
```
SirSplasch: maximum line length exceeded

SirSplasch (changes requested):

SirSplasch (approved):

SirSplasch:
fix failed tests

PR #1705, opened 2026-05-08
Add chunked file uploads for FilePond

patrickweh:
## Summary

Files larger than the configurable chunk threshold (4 MB by default) currently fail to upload through Livewire's single-request `/livewire/upload-file` endpoint when the server enforces a smaller `post_max_size` (e.g. FrankenPHP's 8 MB default). This change adds a chunked upload path that bypasses the single-request limit while staying within the existing `WithFileUploads` flow.

## What changed

- **New controller** `FilePondChunkController` implementing the FilePond chunked upload protocol (`POST` init / `PATCH` chunk / `HEAD` resume) at `/flux/file-pond/chunk`.
- **New `WithFilePond::acceptChunkedUpload(string $signedPath, ?string $collectionName = null): ?string`** — validates the signed transfer ID, attaches the assembled file to `$this->files` as a `TemporaryUploadedFile`, and runs the existing `validateOnDemand` rules.
- **`resources/js/components/file-pond.js`** — files above `CHUNK_THRESHOLD` (4 MB) stream via the chunk endpoint, smaller files keep the existing `$wire.upload()` path so behaviour is unchanged for the common case.
- **`config/flux.php`** — `flux.file_uploads.max_size` (default `1G`) caps total upload size; `flux.file_uploads.chunk_rules` overrides the validation rules applied at finalize time and falls back to `livewire.temporary_file_upload.rules`.

## Security

- Endpoint sits behind `auth:web` + `throttle:300,1` and the existing `web` middleware (CSRF).
- Transfer IDs are HMAC-signed with Livewire's encrypter key (same scheme as `TemporaryUploadedFile::signPath`).
- Each chunk session stores `user_id` + `user_type` (morph alias) and rejects `PATCH`/`HEAD` requests that don't match the originating user.
- Per-chunk and total size are bounded to the declared `Upload-Length`; the assembled file is run through Laravel's validator at finalize and removed on failure.
- S3, GCS, and any non-local disk are rejected up front because the implementation relies on local filesystem locking and append semantics.

## Summary by Sourcery

Introduce chunked FilePond uploads to support large files within Livewire’s existing upload flow.

New Features:
- Add a FilePond chunked upload flow in the frontend that falls back to the existing Livewire upload path for small files.
- Expose a Livewire `acceptChunkedUpload` hook to accept finalized chunked uploads and integrate them into the existing file collection lifecycle.
- Introduce configurable file upload limits and validation rules for chunked uploads via `flux.file_uploads` configuration.

Enhancements:
- Add a chunk-capable FilePond upload controller and route that implement the FilePond chunk protocol while enforcing local-disk-only storage and user-bound sessions.

Tests:
- Add feature tests covering chunk session initialization, multi-chunk assembly, resume semantics, cross-user access rejection, and validation/size error handling.

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -32,7 +32,7 @@ Route::middleware('web')
         });
 
         Route::middleware(['auth:web', 'throttle:300,1'])->group(function (): void {
-            Route::match(['POST', 'PATCH'], '/flux/file-pond/chunk', [FilePondChunkController::class, 'handle'])
+            Route::match(['POST', 'PATCH'], '/file-pond/chunk', [FilePondChunkController::class, 'handle'])
                 ->name('file-pond.chunk');
         });
 
--- a/routes/web.php
+++ b/routes/web.php
@@ -1,6 +1,7 @@
 <?php
 
 use FluxErp\Http\Controllers\AssetController;
+use FluxErp\Http\Controllers\FilePondChunkController;
 use FluxErp\Http\Controllers\LoginLinkController;
 use FluxErp\Http\Controllers\MobileController;
 use FluxErp\Livewire\Features\SignaturePublicLink;
@@ -28,6 +29,11 @@ Route::middleware('web')
         Route::middleware('signed')->group(function (): void {
             Route::get('/login-link', LoginLinkController::class)->name('login-link');
             Route::get('/signature-public/{uuid}', SignaturePublicLink::class)->name('signature.public');
+        });
+
+        Route::middleware(['auth:web', 'throttle:300,1'])->group(function (): void {
+            Route::match(['POST', 'PATCH', 'HEAD'], '/flux/file-pond/chunk', [FilePondChunkController::class, 'handle'])
+                ->name('file-pond.chunk');
         });
 
         Route::middleware('cache.headers:public;max_age=31536000;etag')->group(function (): void {
--- a/routes/web.php
+++ b/routes/web.php
@@ -32,7 +32,7 @@ Route::middleware('web')
         });
 
         Route::middleware(['auth:web', 'throttle:300,1'])->group(function (): void {
-            Route::match(['POST', 'PATCH', 'HEAD'], '/flux/file-pond/chunk', [FilePondChunkController::class, 'handle'])
+            Route::match(['POST', 'PATCH'], '/flux/file-pond/chunk', [FilePondChunkController::class, 'handle'])
                 ->name('file-pond.chunk');
         });
 
--- a/src/Http/Controllers/FilePondChunkController.php
+++ b/src/Http/Controllers/FilePondChunkController.php
@@ -2,9 +2,10 @@
 
 namespace FluxErp\Http\Controllers;
 
+use FluxErp\Helpers\ResponseHelper;
 use Illuminate\Filesystem\FilesystemAdapter;
+use Illuminate\Http\JsonResponse;
 use Illuminate\Http\Request;
-use Illuminate\Http\Response;
 use Illuminate\Support\Facades\Validator;
 use Illuminate\Support\Number;
 use Illuminate\Support\Str;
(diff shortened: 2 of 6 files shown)


routes/web.php
```diff
@@ -30,6 +31,11 @@
             Route::get('/signature-public/{uuid}', SignaturePublicLink::class)->name('signature.public');
         });
 
+        Route::middleware(['auth:web', 'throttle:300,1'])->group(function (): void {
+            Route::match(['POST', 'PATCH'], '/flux/file-pond/chunk', [FilePondChunkController::class, 'handle'])
```
SirSplasch: Why is this "prefixed" with "/flux"?

src/Http/Controllers/FilePondChunkController.php
```diff
@@ -0,0 +1,221 @@
...
+        if ($extension === '' || ! preg_match('/^[a-z0-9]+$/', $extension)) {
+            return response('Invalid file extension', 422);
+        }
+
+        $filename = Str::random(40) . '.' . $extension;
+        $signedPath = TemporaryUploadedFile::signPath($filename);
+
+        $disk = FileUploadConfiguration::storage();
+
+        $user = auth('web')->user();
```
SirSplasch: Why specify 'web'?

src/Http/Controllers/FilePondChunkController.php
```diff
@@ -0,0 +1,221 @@
...
+use Illuminate\Support\Facades\Validator;
+use Illuminate\Support\Number;
+use Illuminate\Support\Str;
+use Livewire\Features\SupportFileUploads\FileUploadConfiguration;
+use Livewire\Features\SupportFileUploads\TemporaryUploadedFile;
+use Symfony\Component\HttpFoundation\File\Exception\FileException;
+
+class FilePondChunkController extends Controller
+{
+    private const TRANSFER_ID_PATTERN = '/^[a-f0-9]{8}:[A-Za-z0-9]{40}\.[a-z0-9]+$/';
```
SirSplasch: This is only used once. Why put it in a property?

src/Http/Controllers/FilePondChunkController.php
```diff
@@ -0,0 +1,221 @@
...
+
+        $disk = FileUploadConfiguration::storage();
+        $sessionPath = FileUploadConfiguration::path($filename . '.chunk');
+
+        if (! $disk->exists($sessionPath)) {
+            return [null, null, null, response('Unknown transfer', 404)];
+        }
+
+        $session = json_decode($disk->get($sessionPath), true) ?: [];
+        if (! isset($session['expected_size'])) {
```
SirSplasch: ! ($session['expected_size'] ?? null)

src/Http/Controllers/FilePondChunkController.php
```diff
@@ -0,0 +1,221 @@
...
+
+        $handle = fopen($absolutePath, 'ab');
+        if ($handle === false) {
+            return response('Unable to open upload target', 500);
+        }
+
+        try {
+            if (! flock($handle, LOCK_EX)) {
+                throw new FileException('Unable to lock upload target');
+            }
```
SirSplasch, line 126: new line after }

src/Http/Controllers/FilePondChunkController.php
```diff
@@ -0,0 +1,221 @@
...
+        $newSize = $offset + $chunkSize;
+
+        if ($newSize === $expected) {
+            $finalizeError = $this->finalize($disk, $filename, $session);
+            if ($finalizeError) {
+                return $finalizeError;
+            }
+        }
+
+        return response('', 204, [
```
SirSplasch: 204 is "no-content", therefore no content is expected. I may suspect that the content is even removed from the response. Probably change the status code.

src/Http/Controllers/FilePondChunkController.php
SirSplasch, line 1: responses are yet again differing from other responses of the application. Like please put some order to it and make it consistent.
I am not a fan of this way of giving responses.

SirSplasch, line 1: Also I am not a big fan of 400 as a status code, as it only is "Bad Request" with nothing more to it.

SirSplasch (changes requested):

src/Http/Controllers/FilePondChunkController.php
```diff
@@ -0,0 +1,245 @@
...
+
+        return null;
+    }
+
+    protected function maxUploadSize(): int
+    {
+        $configured = config('flux.file_uploads.max_size');
+
+        if ($configured) {
+            return (int) Number::fromFileSizeToBytes($configured);
```
SirSplasch: This may throw an exception if $configured has an invalid format or size unit.

SirSplasch (changes requested):

SirSplasch (approved):

PR #1707, opened 2026-05-08
Add lang/*.json key sort check to CI

patrickweh:
## Summary

- New `scripts/sort-lang-json.mjs` enforces case-insensitive lexicographic key order in `lang/*.json` (matches the existing convention used in `de.json`/`en.json`).
- CI runs `node scripts/sort-lang-json.mjs --check` after Prettier; PRs that reorder existing keys or insert new keys at the wrong position fail the lint job instead of slipping through.
- `npm run sort-lang` and `npm run format` apply the fix locally.

`lang/de.json` had a single inversion (`Enable this if you want…` was placed before `Enable`); the script's first pass corrects that one swap.

## Summary by Sourcery

Enforce a consistent, case-insensitive key ordering for all translation JSON files and integrate the check into local scripts and CI.

Build:
- Add npm scripts to sort translation JSON files and include the sorter in the formatting workflow.

CI:
- Run a translation key sort check in the PR lint workflow to prevent unsorted lang/*.json files from being merged.

Chores:
- Introduce a script to sort keys in lang/*.json and apply the correct order to the existing German translations.


scripts/sort-lang-json.mjs
```diff
@@ -0,0 +1,57 @@
...
+function sortKeys(obj) {
+    if (Array.isArray(obj)) return obj.map(sortKeys);
+    if (obj && typeof obj === 'object') {
+        const sortedEntries = Object.keys(obj)
+            .sort((a, b) => {
+                const al = a.toLowerCase();
+                const bl = b.toLowerCase();
+                if (al < bl) return -1;
+                if (al > bl) return 1;
+                return 0;
```
SirSplasch, line 18: does js not have the spaceship operator? al <=> bl

scripts/sort-lang-json.mjs
```diff
@@ -0,0 +1,57 @@
...
+            })
+            .map((k) => [k, sortKeys(obj[k])]);
+        return Object.fromEntries(sortedEntries);
```
SirSplasch, line 22: new line before return

scripts/sort-lang-json.mjs
```diff
@@ -0,0 +1,57 @@
...
+    }
+    return obj;
```
SirSplasch, line 25: new line before return

SirSplasch (changes requested):

scripts/sort-lang-json.mjs
patrickweh: nope

SirSplasch (approved):

PR #1718, opened 2026-05-18
fix: escape slug in MediaFolder descendant slug update

patrickweh:
## Summary
The `saved` hook on `MediaFolder` rebuilds descendant slugs via a raw `UPDATE … SET slug = CONCAT('<slug>', SUBSTRING(slug, X))`. The slug was concatenated into the SQL string without escaping, so any folder whose name contained a single quote (e.g. real-world data like `Zusatzinfo's`) produced a 1064 syntax error and the rename/move silently failed for the whole descendant tree.

Now the new slug is passed through the model's connection PDO `quote()` so it is properly escaped. Also wraps `strlen($original) + 1` in parentheses to make operator precedence explicit (the previous expression worked by coincidence).

## Regression test
Renaming `Zusatzinfo's` to `Zusatzinfo's renamed` while it has a descendant must keep the descendant slug consistent and must not throw. Pinned with `toBe()` so future changes to the slug strategy stay deterministic.

## Summary by Sourcery

Escape media folder slugs when updating descendant records to prevent SQL errors with apostrophes and add a regression test for renaming folders containing apostrophes.

Bug Fixes:
- Fix descendant slug update for media folders whose names contain apostrophes by quoting the new slug in the raw SQL update.

Tests:
- Add a regression test ensuring renaming a media folder with an apostrophe keeps descendant slugs consistent and does not throw.

### DIFF
--- a/src/Models/MediaFolder.php
+++ b/src/Models/MediaFolder.php
@@ -57,11 +57,11 @@ class MediaFolder extends FluxModel implements HasMedia
         static::saved(function (MediaFolder $model): void {
             if ($model->wasChanged(['parent_id', 'slug'])) {
                 $original = $model->getRawOriginal('slug');
-                $replace = $model->slug;
+                $quotedSlug = $model->getConnection()->getPdo()->quote($model->slug);
                 $model->getAllDescendantsQuery()
                     ->update([
-                        'slug' => DB::raw('CONCAT(\'' . $replace
-                            . '\', SUBSTRING(slug, ' . strlen($original) + 1 . '))'
+                        'slug' => DB::raw('CONCAT(' . $quotedSlug
+                            . ', SUBSTRING(slug, ' . (strlen($original) + 1) . '))'
                         ),
                     ]);
             }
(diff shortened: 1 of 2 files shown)


src/Models/MediaFolder.php
```diff
@@ -57,11 +57,11 @@ protected static function booted(): void
         static::saved(function (MediaFolder $model): void {
             if ($model->wasChanged(['parent_id', 'slug'])) {
                 $original = $model->getRawOriginal('slug');
-                $replace = $model->slug;
+                $quotedSlug = $model->getConnection()->getPdo()->quote($model->slug);
```
SirSplasch, line 60: ->quote() can return false, if this happens consider using addslashes() as a fallback although it is not perfect.

SirSplasch (changes requested):

src/Models/MediaFolder.php
```diff
@@ -57,11 +57,16 @@ protected static function booted(): void
...
+
+                if ($quotedSlug === false) {
+                    $quotedSlug = "'" . addslashes($model->slug) . "'";
```
SirSplasch: do not use double quotes. `'\''`

SirSplasch (changes requested):

SirSplasch (approved):

PR #1728, opened 2026-05-19
fix(media-tree): preserve sibling branches when dotted collection names share a prefix

patrickweh:
## Summary
- The inner loop in InteractsWithMedia::getMediaAsTree built nested folders via Arr::set on a dotted path
- Arr::set replaces the array at the final path key, so each iteration that touched an intermediate node (e.g. \`Berater.children.Nextcloud\`) wiped any \`children\` already accumulated on it
- Sorted slug processing meant only the alphabetically last branch survived under any shared prefix; earlier siblings disappeared from the rendered tree even though the records existed in DB
- Real-world repro: address with media under \`Berater.Nextcloud.BRV\`, \`Berater.Nextcloud.Daten Darlehen\`, \`Berater.Nextcloud.PKV\`, … only \`PKV\` showed up

## Fix
- Walk the path by reference and initialise each level idempotently; never replace an existing node or its \`children\`
- Regression test attaches three media records sharing a prefix and asserts all three siblings survive on the parent

## Summary by Sourcery

Ensure media tree construction preserves all sibling branches for dotted collection names sharing a prefix.

Bug Fixes:
- Fix media tree building logic so intermediate nodes created from dotted slugs no longer overwrite existing siblings, preventing branches from disappearing in the rendered tree.

Tests:
- Add a feature test that asserts multiple media collections with a shared dotted prefix all appear as distinct sibling branches in the resulting media tree.

### DIFF
--- a/src/Traits/Model/InteractsWithMedia.php
+++ b/src/Traits/Model/InteractsWithMedia.php
@@ -81,31 +81,35 @@ trait InteractsWithMedia
         $undotted = [];
         foreach ($mediaCollections as $mediaCollection) {
             $slug = data_get($mediaCollection, 'slug') ?? '';
+
             if (
                 is_null(data_get($mediaCollection, 'id'))
                 && str_contains($slug, '.')
             ) {
-                $path = '';
+                $cursor = &$undotted;
                 $childSlug = '';
-                foreach (explode('.', $slug) as $index => $part) {
-                    if ($index === 0 && array_key_exists($part, $undotted)) {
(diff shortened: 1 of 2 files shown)


src/Traits/Model/InteractsWithMedia.php
SirSplasch, line 1: array_key_exists() instead of isset()

SirSplasch (changes requested):

SirSplasch (approved):

PR #1729, opened 2026-05-19
fix(media): stream downloads through a dedicated disk-agnostic route

patrickweh:
## Summary
- `SupportsFileDownloads::download` redirected to the media URL. On the public disk that URL has no `Content-Disposition` header, so browsers rendered images and PDFs inline instead of downloading them
- Returning a `BinaryFileResponse` straight from the Livewire action would have forced the download but Livewire base64-encodes the whole response body for transport, which blows up memory on large files
- Add a signed `media.show` route that uses `Storage::disk(...)` so it adapts to whichever driver the media lives on:
  - local/public disks stream the file with a conditional `Content-Disposition` header (inline by default, attachment when `?download=1` is set)
  - S3-style disks (`providesTemporaryUrls`) redirect to a presigned URL with `ResponseContentDisposition` so the client downloads directly from the bucket without proxying through us
- Mark `download()` and `downloadCollection()` `#[Renderless]` so the component template is not re-rendered on a redirect

### DIFF
--- a/routes/frontend/web.php
+++ b/routes/frontend/web.php
@@ -376,7 +376,8 @@ Route::middleware('web')
         Route::middleware('signed')->group(function (): void {
             Route::get('/media-private/{media}/{filename}', function (Media $media) {
                 return $media;
-            })->name('media.private');
+            })
+                ->name('media.private');
 
             Route::get('/media/{media}', function (Media $media) {
                 $disposition = HeaderUtils::makeDisposition(
@@ -404,7 +405,8 @@ Route::middleware('web')
                     $media->file_name,
                     ['Content-Disposition' => $disposition],
                 );
-            })->name('media.show');
+            })
+                ->name('media.show');
 
             Route::get('/media-collection-download/{token}', function (string $token) {
                 $payload = Crypt::decrypt($token);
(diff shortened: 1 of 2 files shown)


routes/frontend/web.php
```diff
@@ -377,6 +377,30 @@
                 return $media;
             })->name('media.private');
```
SirSplasch: wrap ->name()

routes/frontend/web.php
```diff
@@ -377,6 +377,30 @@
...
+                        )
+                    );
+                }
+
+                return $disk->response(
+                    $path,
+                    $media->file_name,
+                    ['Content-Disposition' => $disposition],
+                );
+            })->name('media.show');
```
SirSplasch: wrap ->name()

src/Traits/Livewire/SupportsFileDownloads.php
```diff
@@ -27,9 +29,7 @@ public function download(Media $media): void
         }
 
         try {
-            DownloadMedia::make([
-                'id' => $media->getKey(),
-            ])
+            DownloadMedia::make(['id' => $media->getKey()])
```
SirSplasch: Did you need to change the array formatting?

SirSplasch (changes requested):

SirSplasch (approved):

PR #1734, opened 2026-05-20
fix(media): serve conversion files via media.private when the conversion disk is private

patrickweh:
## Problem

\`FluxErp\\Support\\MediaLibrary\\UrlGenerator::getUrl()\` chose between a signed \`media.private\` route and \`parent::getUrl()\` purely on the original \`\$media->disk\`. So for any media whose original lives on a private disk — even when a conversion was explicitly requested — it returned a signed URL to the **original** file, never to the generated derivative.

Concrete failure mode in production: VVO stores originals on a Hetzner S3 disk (\`disk=hetzner\`) and conversions on the local public disk (\`conversions_disk=public\`). The Cloud Browser's \`MediaGrid\` calls \`\$media->getUrl('thumb_400x400')\` and shipped the signed \`/media-private/{id}/{original}.pdf\` URL to the browser as a thumbnail \`\`. Result: 5–30 MB PDFs/images downloaded per tile, no actual thumbnail rendered.

## Fix

1. \`UrlGenerator\` now picks the *effective* disk based on whether a conversion is requested:
   - \`null\` conversion → \`\$media->disk\`
   - any conversion → \`\$media->conversions_disk\`
2. When that effective disk is \`public\`, fall through to \`parent::getUrl()\` and let Spatie produce the standard \`/storage/.../-thumb_400x400.jpg\` URL.
3. When the effective disk is private (i.e. the conversion itself needs auth), sign the \`media.private\` route and pass \`?conversion={name}\` so the route handler can serve the matching derivative.
4. The route closure moved into a single-action \`PrivateMediaController\`. With a conversion query parameter present, it streams the conversion file from the conversions disk (or 302s to a temporary URL for disks that support them); without it, returns the original (unchanged behavior).
5. \`404\` when a conversion query parameter is supplied for a conversion that wasn't generated.

## Compatibility

- Public-disk-only installs (originals on \`public\`): no change, both branches go through \`parent::getUrl()\`.
- Private-disk originals without conversions: unchanged — same signed \`media.private\` URL as before.
- Private-disk originals with a public conversions disk (VVO setup): now returns the public \`/storage/...\` thumbnail URL instead of a signed original URL.
- Private-disk originals with a private conversions disk: now signs the route with \`?conversion=…\` and the controller serves the derivative.

## Tests

- \`tests/Unit/Support/MediaLibrary/UrlGeneratorTest.php\`: 4 unit cases for each disk combination.
- \`tests/Feature/Web/PrivateMediaTest.php\`: 4 HTTP tests for the new controller (original, conversion served, unsigned rejected, ungenerated conversion → 404).

\`\`\`
Tests:    8 passed (19 assertions)
\`\`\`

## Summary by Sourcery

Serve media conversions and originals from the appropriate disk, signing private routes when needed and delegating conversion responses to a dedicated controller.

Bug Fixes:
- Ensure media conversions requested for media on private disks return URLs to the conversion files instead of the original files.

Enhancements:
- Determine the effective disk per request (original vs. conversion) when generating media URLs and append the conversion name to signed private routes where applicable.
- Introduce a dedicated PrivateMediaController to handle signed media-private requests for originals and conversions, including streaming or redirecting to temporary URLs for conversion files.

Tests:
- Add unit tests for UrlGenerator to cover combinations of original and conversion disks and URL generation behavior.
- Add feature tests for the media.private route to verify access control, original vs. conversion responses, and 404 handling for missing conversions.

### DIFF
--- a/routes/frontend/web.php
+++ b/routes/frontend/web.php
@@ -6,6 +6,7 @@ use FluxErp\Actions\PushSubscription\UpsertPushSubscription;
 use FluxErp\Http\Controllers\AuthController;
 use FluxErp\Http\Controllers\CalendarEventController;
 use FluxErp\Http\Controllers\CalendarSearchController;
+use FluxErp\Http\Controllers\PrivateMediaController;
 use FluxErp\Http\Controllers\SearchController;
 use FluxErp\Http\Middleware\TrackVisits;
 use FluxErp\Livewire\AbsenceRequest\AbsenceRequest;
@@ -374,9 +375,7 @@ Route::middleware('web')
         });
 
         Route::middleware('signed')->group(function (): void {
-            Route::get('/media-private/{media}/{filename}', function (Media $media) {
-                return $media;
-            })
+            Route::get('/media-private/{media}/{filename}', PrivateMediaController::class)
                 ->name('media.private');
 
             Route::get('/media/{media}', function (Media $media) {
--- a/src/Http/Controllers/PrivateMediaController.php
+++ b/src/Http/Controllers/PrivateMediaController.php
@@ -1,0 +1,52 @@
+<?php
+
+namespace FluxErp\Http\Controllers;
+
+use FluxErp\Models\Media;
+use Illuminate\Http\RedirectResponse;
+use Illuminate\Http\Request;
+use Illuminate\Support\Facades\Storage;
+use Symfony\Component\HttpFoundation\BinaryFileResponse;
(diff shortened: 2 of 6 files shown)


SirSplasch (approved):

SirSplasch (changes requested):
actually take a look at the ai annotations, they are valid.

SirSplasch (approved):

PR #1746, opened 2026-05-26
fix(queue-monitor): resolve queue from connection config when JobQueued has no queue

patrickweh:
## Summary

Fixes a duplicate `QueueMonitor` row that was created on every dispatch when the job was queued without an explicit `onQueue()` call and the connection's default queue differs from `'default'`.

Laravel emits `JobQueued` with `$event->queue = null` when nothing was passed to `dispatch()->onQueue(...)`. `QueueMonitorManager::jobQueued()` previously read `$event->job->queue` (the job class property, also null in that path) and fell back to `'default'`. `jobProcessing()` resolves the runtime queue via `$event->job->getQueue()`, which is the connection's configured default queue (e.g. a tenant-specific queue).

The mismatch meant `firstOrNew(... whereState Queued)` in `jobProcessing()` could not find the Queued monitor stored at dispatch time, so it created a second monitor and then marked the original as stale. Each export then produced:

1. An extra `JobFinishedNotification` for the stale monitor with `accept = null` -- which the frontend persisted with no download action.
2. A `UniqueConstraintViolationException` when the real `JobFinishedNotification` tried to insert the same UUID with the real download URL.

Result: the toast on the user's screen showed "Export ist beendet" but no download button.

Read `$event->queue` first and fall back to the connection's configured queue so the row stored at dispatch time matches what the worker resolves.

## Test plan

- `testbench package:test --filter QueueMonitorManagerTest`

## Summary by Sourcery

Ensure QueueMonitor records use the effective queue name resolved from the connection configuration so queued and processing events refer to the same monitor entry.

Bug Fixes:
- Prevent duplicate QueueMonitor rows when jobs are dispatched without an explicit queue on connections whose default queue is not `default`.

Tests:
- Add unit tests covering queue resolution from JobQueued events, fallback to the connection default queue, and reuse of the existing QueueMonitor during job processing.

### DIFF
--- a/src/Support/QueueMonitor/QueueMonitorManager.php
+++ b/src/Support/QueueMonitor/QueueMonitorManager.php
@@ -137,16 +137,19 @@ class QueueMonitorManager
     {
         $now = Carbon::now();
 
+        $queue = $event->job->getQueue()
+            ?: config("queue.connections.{$event->connectionName}.queue", 'default');
+
         $monitor = resolve_static(QueueMonitor::class, 'query')
             ->where('job_id', $jobId = static::getJobId($event->job))
-            ->where('queue', $event->job->getQueue() ?? config('queue.default'))
+            ->where('queue', $queue)
             ->whereState('state', Queued::class)
             ->firstOrNew();
 
         $monitor->fill([
             'job_uuid' => $event->job->uuid(),
             'job_id' => static::getJobId($event->job),
-            'queue' => $event->job->getQueue() ?: config('queue.default'),
+            'queue' => $queue,
             'name' => static::getJobName($event),
             'started_at' => $now,
             'started_at_exact' => $now->format('Y-m-d H:i:s.u'),
--- a/src/Support/QueueMonitor/QueueMonitorManager.php
+++ b/src/Support/QueueMonitor/QueueMonitorManager.php
@@ -177,7 +177,8 @@ class QueueMonitorManager
                 'job_batch_id' => $event->job->batchId ?? null,
                 'job_id' => $event->id,
                 'name' => static::getJobName($event),
-                'queue' => $event->job->queue ?: 'default',
+                'queue' => $event->queue
+                    ?: config("queue.connections.{$event->connectionName}.queue", 'default'),
                 'state' => Queued::class,
                 'queued_at' => now(),
                 'data' => $data ?? null,
--- a/tests/Unit/Support/QueueMonitor/QueueMonitorManagerTest.php
+++ b/tests/Unit/Support/QueueMonitor/QueueMonitorManagerTest.php
@@ -1,0 +1,91 @@
+<?php
+
+use FluxErp\Jobs\ExportDataTableJob;
+use FluxErp\Models\QueueMonitor;
+use FluxErp\States\QueueMonitor\Running;
+use FluxErp\Support\QueueMonitor\QueueMonitorManager;
+use Illuminate\Contracts\Queue\Job;
+use Illuminate\Queue\Events\JobProcessing;
+use Illuminate\Queue\Events\JobQueued;
+
+test('jobQueued stores the resolved event queue, not the unset job property', function (): void {
+    $job = new ExportDataTableJob(
+        component: serialize(new stdClass()),
+        modelClass: 'FluxErp\\Models\\Tenant',
+        columns: ['name'],
+        userMorph: $this->user->getMorphClass() . ':' . $this->user->getKey(),
+    );
+
+    QueueMonitorManager::handle(new JobQueued(
+        connectionName: 'redis',
+        queue: 'flux_tenant_queue',
+        id: 'job-id-1',
+        payload: '',
+        job: $job,
+        delay: 0,
+    ));
+
+    expect(QueueMonitor::count())->toBe(1);
+    expect(QueueMonitor::first()->queue)->toBe('flux_tenant_queue');
+});
+
+test('jobQueued falls back to the connection default queue when the event queue is null', function (): void {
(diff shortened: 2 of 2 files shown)


src/Support/QueueMonitor/QueueMonitorManager.php
```diff
@@ -177,7 +177,8 @@ protected static function jobQueued(JobQueued $event): void
                 'job_batch_id' => $event->job->batchId ?? null,
                 'job_id' => $event->id,
                 'name' => static::getJobName($event),
-                'queue' => $event->job->queue ?: 'default',
+                'queue' => $event->queue
+                    ?: config("queue.connections.{$event->connectionName}.queue", 'default'),
```
patrickweh, line 184: this fallback uses `queue.connections.{$connectionName}.queue`, but `jobProcessing()` falls back to `config('queue.default')` -- which returns the default *connection* name, not a queue. inconsistent, and that one is technically a dead bug (only hit if `getQueue()` returns null). worth aligning while we're here.

tests/Unit/Support/QueueMonitor/QueueMonitorManagerTest.php
```diff
@@ -0,0 +1,91 @@
...
+use FluxErp\Jobs\ExportDataTableJob;
+use FluxErp\Models\QueueMonitor;
+use FluxErp\States\QueueMonitor\Running;
+use FluxErp\Support\QueueMonitor\QueueMonitorManager;
+use Illuminate\Contracts\Queue\Job;
+use Illuminate\Queue\Events\JobProcessing;
+use Illuminate\Queue\Events\JobQueued;
+
+test('jobQueued stores the resolved event queue, not the unset job property', function (): void {
+    $job = new ExportDataTableJob(
```
patrickweh: same 5-arg job fixture is built three times. extract to a helper or beforeEach.

tests/Unit/Support/QueueMonitor/QueueMonitorManagerTest.php
```diff
@@ -0,0 +1,91 @@
...
+        job: $job,
+        delay: 0,
+    ));
+
+    expect(QueueMonitor::count())->toBe(1);
+    expect(QueueMonitor::first()->queue)->toBe('flux_tenant_queue');
+});
+
+test('jobQueued falls back to the connection default queue when the event queue is null', function (): void {
+    config()->set('queue.connections.redis.queue', 'flux_tenant_queue');
```
patrickweh, line 54: config mutation leaks between tests. RefreshDatabase doesn't reset config. add an `afterEach` to restore, or both tests will silently depend on each other's ordering.

tests/Unit/Support/QueueMonitor/QueueMonitorManagerTest.php
```diff
@@ -0,0 +1,91 @@
...
+    QueueMonitorManager::handle(new JobQueued(
+        connectionName: 'redis',
+        queue: null,
+        id: 'job-id-null-queue',
+        payload: '',
+        job: $job,
+        delay: 0,
+    ));
+
+    expect(QueueMonitor::first()->queue)->toBe('flux_tenant_queue');
```
patrickweh: missing `expect(QueueMonitor::count())->toBe(1)` like test 1. inconsistent.

tests/Unit/Support/QueueMonitor/QueueMonitorManagerTest.php
```diff
@@ -0,0 +1,91 @@
...
+});
+
+test('jobProcessing reuses the QueueMonitor created by a null-queue dispatch on the same connection', function (): void {
+    config()->set('queue.connections.redis.queue', 'flux_tenant_queue');
+
+    $userJob = new ExportDataTableJob(
```
patrickweh: naming inconsistency: `$job` in tests 1/2, `$userJob` here. just `$job`.

tests/Unit/Support/QueueMonitor/QueueMonitorManagerTest.php
```diff
@@ -0,0 +1,91 @@
...
+    expect(QueueMonitor::count())->toBe(1);
+
+    $queueJob = Mockery::mock(Job::class);
+    $queueJob->shouldReceive('resolveName')->andReturn(ExportDataTableJob::class);
+    $queueJob->shouldReceive('getJobId')->andReturn('job-id-1');
+    $queueJob->shouldReceive('getRawBody')->andReturn('');
+    $queueJob->shouldReceive('getQueue')->andReturn('flux_tenant_queue');
+    $queueJob->shouldReceive('uuid')->andReturn('uuid-1');
+    $queueJob->shouldReceive('attempts')->andReturn(1);
+    $queueJob->shouldReceive('payload')->andReturn(['data' => ['commandName' => ExportDataTableJob::class]]);
```
patrickweh: mock is tightly coupled to the call list inside `jobProcessing()`. add a single internal call there and this breaks for non-functional reasons. fragile -- a real Job impl or shallow stub would survive refactors better.

patrickweh:
Addressed the review notes in b41d7449e:

- jobProcessing now uses the same connection-config fallback as jobQueued (the `config('queue.default')` connection-name miss was dead-fallback in practice, but cleaner aligned)
- job fixture + queue-job mock factory extracted to beforeEach closures
- mutated `queue.connections.redis.queue` config restored in afterEach
- added missing `QueueMonitor::count()` assertion in the null-queue test
- collapsed two-call `expect` blocks into single `->and()` chains

The mock surface in the third test is still shaped to what `jobProcessing` actually calls (resolveName/getJobId/getRawBody/getQueue/uuid/attempts/payload); collapsed into a factory so any future change only needs adjusting in one place rather than per test.

PR #1753, opened 2026-05-28
Fix tenant postal_address_one_line to include city after postcode

patrickweh:
## Summary
- `postcode` and `city` were combined with `??` having lower precedence than `.`, so the expression collapsed to just `$attributes['postcode'] ?? ' ' . $attributes['city']` and the city was always dropped
- Fix wraps both sides of the concatenation in `(... ?? '')`, then `?: null` keeps an empty segment out of the `array_filter` chain
- Adds a unit test covering full data, missing city, missing postcode, and both empty

## Summary by Sourcery

Fix tenant postal address formatting so postcode and city are correctly combined on one line while omitting empty postcode/city segments.

Bug Fixes:
- Correct postal_address_one_line composition to ensure both postcode and city are included when present and omitted when empty.

Tests:
- Add unit tests covering postal_address_one_line formatting for full data, missing city, missing postcode, and both values empty.

### DIFF
--- a/src/Models/Tenant.php
+++ b/src/Models/Tenant.php
@@ -114,7 +114,7 @@ class Tenant extends FluxModel implements HasMedia
                 array_filter([
                     $attributes['name'] ?? null,
                     $attributes['street'] ?? null,
-                    trim($attributes['postcode'] ?? null . ' ' . $attributes['city'] ?? null),
+                    trim(($attributes['postcode'] ?? '') . ' ' . ($attributes['city'] ?? '')) ?: null,
                 ])
             )
         );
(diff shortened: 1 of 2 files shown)


patrickweh:
Re extraction into a local variable: the same inline pattern is used in `src/Models/Address.php` for both `firstname/lastname` and `zip/city` segments in `postalAddress`/`name`:

```php
trim(($attributes['firstname'] ?? '') . ' ' . ($attributes['lastname'] ?? '')),
trim(($attributes['zip'] ?? '') . ' ' . ($attributes['city'] ?? '')),
```

Keeping the arrow-function form consistent with that. Dropped the redundant containment assertion in the latest commit.

PR #1769, opened 2026-06-03
Add stack hook for extending order position modal fields

patrickweh:
## Summary
- New `@stack('order-edit-position-modal-fields')` inside the order position edit modal body so downstream packages can append fields without overriding the full modal.

## Summary by Sourcery

New Features:
- Introduce an order-edit-position-modal-fields Blade stack hook within the order position edit modal body for downstream field extensions.


resources/views/livewire/order/order-list-header.blade.php
SirSplasch, line 1: indentations on some selects and x-text do not match.

SirSplasch (changes requested):

SirSplasch (approved):

PR #1770, opened 2026-06-03
Mention infrastructure: @/# mentions, state dots & clickable comment pills

Slupi:
Central mention infrastructure in flux-core.

- Generic `mentions` table + `Mentionable` trait, parser/renderer/sync, `MentionableSearchController`. Convention: `@` = people (User), `#` = records.
- Comments: mentions are indexed from the stored HTML; pills are clickable links (`<a href>`) with a type tag.
- Live **state dot** on record pills (color from the state, full name in the tooltip) — live in chat and in comments (refreshed on load).
- Label/URL are derived from `InteractsWithDataTables`, so a model becomes mentionable with just `use Mentionable`. Default records: **Ticket, Task, Project, Order, Address** (User as `@`). `@user` deliberately does not link.
- Search endpoint is resilient: if one type fails (e.g. a missing Scout index) it's skipped instead of failing the whole typeahead.

## Summary by Sourcery

Introduce a centralized, token-based mention system for users and records across comments and the editor, with searchable, type-aware pills and live state indicators on referenced records.

New Features:
- Add generic mention infrastructure with a mentions table, model trait, parser, renderer, and sync logic to track and materialize user and record mentions from text and HTML.
- Enable @-based user mentions and #-based record mentions in the Tiptap editor with scoped typeahead search, type badges, and clickable pills.
- Expose a mentionable search endpoint that aggregates mention candidates across types, supports type scoping, and tolerates per-type search failures.
- Add live-refreshable mention pills in comments that show record state dots and type labels, including graceful handling of deleted targets.
- Introduce mention notifications so users are alerted when mentioned, with support for custom notification types from source models.

Enhancements:
- Make core models (User, Ticket, Task, Project, Order, Address) mentionable via a shared trait that derives labels, URLs, and optional state metadata for mention pills.
- Refine mention pill styling with per-type colors, state-dot indicators, and optional type badges, while keeping non-mention links unaffected.
- Wire comments and Livewire comment flows to the new mention system, replacing legacy HTML-based mention parsing and subscription logic.

Tests:
- Add extensive unit, feature, Livewire, browser (skipped), and migration tests covering mention parsing, rendering, syncing, search behavior, notifications, state propagation, and trait behavior.


database/migrations/2026_05_28_000001_create_mentions_table.php
```diff
@@ -0,0 +1,29 @@
...
+
+return new class() extends Migration
+{
+    public function up(): void
+    {
+        Schema::create('mentions', function (Blueprint $table): void {
+            $table->id();
+            $table->morphs('mention_source');
+            $table->string('mention_target_type')->nullable();
+            $table->unsignedBigInteger('mention_target_id')->nullable();
```
SirSplasch: nullableMorph()

resources/views/components/editor.blade.php
SirSplasch, line 47: Can these indentations be fixed? Or are they affected from prettier?

src/Enums/MentionTypeEnum.php
```diff
@@ -0,0 +1,18 @@
...
+{
+    use EnumTrait;
+
+    case User = 'user';
+
+    case Record = 'record';
+
+    case Channel = 'channel';
+
+    case Here = 'here';
```
SirSplasch: These two mention types do not make sense from the view of the core, as there is no "channel" or "here".
Probably change this enum to a FluxEnum and overwrite it in the chat package?

src/Models/Comment.php
```diff
@@ -81,6 +61,19 @@ public function broadcastChannel(): string
...
 
+    /**
+     * @return array<int, string>
+     */
+    public function mentionableTextFields(): array
+    {
+        return ['comment'];
+    }
+
+    public function mentionScannableText(): string
```
SirSplasch: put these methods below broadcastWith() to match alphabetical ordering in the public methods section.

src/Traits/Model/Mentionable.php
```diff
@@ -0,0 +1,73 @@
...
+trait Mentionable
+{
+    public static function mentionTypeKey(): string
+    {
+        return morph_alias(static::class);
+    }
+
+    public static function mentionTypeLabel(): string
+    {
+        return Str::headline(class_basename(static::class));
```
SirSplasch: Str::headline(morph_alias(static::class))

tests/Feature/Migrations/CreateMentionsTableTest.php
SirSplasch, line 1: Remove this test, no migration has been tested as of today.

tests/Feature/MentionsArchTest.php
SirSplasch, line 1: I don't know about this test tbh. I would remove it.

src/Support/Mentions/MentionableTypesManager.php
SirSplasch, line 1: Better use a Facade with a manager class and register it in the service provider as a singleton. Like for example Widget Facade with its WidgetManager and WidgetServiceProvider.

src/Models/Ticket.php
```diff
@@ -75,6 +76,11 @@ public static function scoutIndexSettings(): ?array
         ];
     }
 
+    public static function mentionTypeIcon(): string
```
SirSplasch: Swap this method with scoutIndexSetting() to match alphabetical ordering in the public static methods section.

src/Models/User.php
```diff
@@ -108,6 +110,24 @@ public static function scoutIndexSettings(): ?array
         ];
     }
 
+    public static function mentionTypeIcon(): string
+    {
+        return 'user';
+    }
+
+    public static function searchMentionCandidates(string $query, int $limit = 5): Collection
```
SirSplasch: Order methods alphabetical inside the public static methods section.

database/migrations/2026_05_28_000001_create_mentions_table.php
```diff
@@ -0,0 +1,29 @@
...
+            $table->string('mention_type', 16);
+            $table->foreignId('user_id')->nullable()->constrained()->cascadeOnDelete();
```
SirSplasch, line 13: user_id after id

database/migrations/2026_05_28_000001_create_mentions_table.php
```diff
@@ -0,0 +1,29 @@
...
+            $table->timestamp('created_at')->useCurrent();
+
+            $table->index(['mention_target_type', 'mention_target_id', 'created_at']);
+            $table->index(['user_id', 'created_at']);
```
SirSplasch: what are these indexes for? foreign key and morph already have an index by its own and is created_at even queried?

database/migrations/2026_05_28_000001_create_mentions_table.php
```diff
@@ -0,0 +1,29 @@
...
+return new class() extends Migration
+{
+    public function up(): void
+    {
+        Schema::create('mentions', function (Blueprint $table): void {
+            $table->id();
+            $table->morphs('mention_source');
+            $table->string('mention_target_type')->nullable();
+            $table->unsignedBigInteger('mention_target_id')->nullable();
+            $table->string('mention_type', 16);
```
SirSplasch: I would not constraint these column to 16 characters, as this can be anything.

SirSplasch: Maybe rename this column to mention_type_enum? Not 100% sure on this one, but renaming it would be more consistent.

database/migrations/2026_05_28_000001_create_mentions_table.php
```diff
@@ -0,0 +1,29 @@
...
+            $table->foreignId('user_id')->nullable()->constrained()->cascadeOnDelete();
+            $table->timestamp('created_at')->useCurrent();
```
SirSplasch: $table->timestamp('created_at')->nullable();
            $table->string('created_by')->nullable();
            $table->timestamp('updated_at')->nullable();
            $table->string('updated_by')->nullable();

src/Observers/RecordsMentionsObserver.php
```diff
@@ -0,0 +1,80 @@
...
+        Mention::query()
+            ->where('mention_source_type', $source->getMorphClass())
+            ->where('mention_source_id', $source->getKey())
+            ->delete();
+    }
+
+    /**
+     * @param  array<string, mixed>  $row
+     */
+    private function dispatchForAddedRow(Model $source, array $row): void
```
SirSplasch: protected

src/Support/Mentions/MentionParser.php
SirSplasch, line 1: make methods protected instead of private

src/Support/Mentions/MentionPillRefresher.php
SirSplasch, line 1: make methods protected instead of private

src/Support/Mentions/MentionRenderer.php
SirSplasch, line 1: make methods protected instead of private

src/Support/Mentions/MentionSync.php
SirSplasch, line 1: make methods protected instead of private

src/Support/Mentions/MentionState.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+namespace FluxErp\Support\Mentions;
+
+readonly class MentionState
```
SirSplasch: Why readonly?

SirSplasch (changes requested):
Tests it() => test()
Services => Support
remove e() function calls

database/migrations/2026_05_28_000001_create_mentions_table.php
Slupi: done

database/migrations/2026_05_28_000001_create_mentions_table.php
Slupi: Done

src/Enums/MentionTypeEnum.php
Slupi: done -> moved to the chat package

tests/Feature/Migrations/CreateMentionsTableTest.php
Slupi: removed

tests/Feature/MentionsArchTest.php
Slupi: removed

src/Support/Mentions/MentionableTypesManager.php
Slupi: Done — MentionableTypesManager + Facades\MentionableTypes, registered as a singleton

src/Models/Ticket.php
Slupi: done

src/Models/User.php
Slupi: done

src/Models/Comment.php
Slupi: done

src/Traits/Model/Mentionable.php
Slupi: done

src/Support/Mentions/MentionParser.php
Slupi: done

src/Support/Mentions/MentionPillRefresher.php
Slupi: done

src/Support/Mentions/MentionRenderer.php
Slupi: done

src/Support/Mentions/MentionSync.php
Slupi: done

src/Support/Mentions/MentionState.php
Slupi: It's an immutable value object, so readonly. Happy to drop it if you'd prefer — no functional change

database/migrations/2026_05_28_000001_create_mentions_table.php
Slupi: done

database/migrations/2026_05_28_000001_create_mentions_table.php
Slupi: Removed

database/migrations/2026_05_28_000001_create_mentions_table.php
Slupi: done

resources/views/components/editor.blade.php
Slupi: It's Prettier's output (prettier --check is green)

src/Observers/RecordsMentionsObserver.php
Slupi: done

database/migrations/2026_05_28_000001_create_mentions_table.php
Slupi: Done

Slupi:
Tests it() => test()  - done
Services => Support - done
remove e() function calls — not done: the e() calls are intentional. These strings are concatenated directly into the HTML output rather than going through Blade's {{ }} auto-escaping, so removing e() would expose an XSS vector for any user-controlled value in them. I'd prefer to keep them as the escaping safeguard.

src/Facades/MentionableTypes.php
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * Resolves the models that use the Mentionable trait into a keyed map and
+ * exposes the record-mentionable keys (every mentionable type except User).
+ *
+ * @method static array map()
+ * @method static array recordKeys()
+ *
+ * @see MentionableTypesManager
+ */
+class MentionableTypes extends Facade
```
SirSplasch: class name in singular

src/Support/Mentions/MentionableTypesManager.php
```diff
@@ -0,0 +1,51 @@
+<?php
+
+namespace FluxErp\Support\Mentions;
+
+use FluxErp\Models\User;
+use FluxErp\Traits\Model\Mentionable;
+use Illuminate\Support\Traits\Macroable;
+
+class MentionableTypesManager
```
SirSplasch: class name in singular MentionableTypeManager

src/Support/Mentions/MentionableTypesManager.php
SirSplasch, line 1: Refactor this, to work like the WidgetManager or ActionManager.

autoDiscover() method which will be called in a MentionableTypeServiceProvider like WidgetServiceProvider
register, unregister method
all() method to get all both record mention and user mention types.
getRecordMentionableTypes() and getUserMentionableTypes() to get each one respectively.
Maybe add a boolean or alike as a parameter to get only the keys as I think you need them with the current recordKeys() method.

Probably two properties: protected array $record = [], protected array $user = []

src/Contracts/MentionsContent.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace FluxErp\Contracts;
+
+interface MentionsContent
+{
+    /**
+     * @return array<int, string>
+     */
+    public function mentionableTextFields(): array;
```
SirSplasch: Isn't is more like mentionableColumns?

src/Support/Mentions/MentionHtml.php
```diff
@@ -0,0 +1,24 @@
...
+
+namespace FluxErp\Support\Mentions;
+
+use FluxErp\Models\User;
+
+class MentionHtml
+{
+    public static function toTokens(string $html): string
+    {
+        $userKey = morph_alias(User::class);
```
SirSplasch: Get the user key(s) from the MentionableType facade

src/Support/Mentions/MentionHtml.php
```diff
@@ -0,0 +1,24 @@
...
+
+        return preg_replace_callback(
+            '/<(?:span|a)\b[^>]*\bdata-id="([a-z][a-z0-9_]*):(\d+)"[^>]*>.*?<\/(?:span|a)>/i',
+            function (array $m) use ($userKey): string {
```
SirSplasch: Not a fan of singular character variables.

src/Models/Mention.php
```diff
@@ -0,0 +1,42 @@
...
+
+class Mention extends FluxModel
+{
+    public $timestamps = false;
+
+    protected $table = 'mentions';
+
+    protected $guarded = [
+        'id',
+    ];
```
SirSplasch: properties can be removed

src/Models/Mention.php
```diff
@@ -0,0 +1,42 @@
...
+
+    protected function casts(): array
+    {
+        return [
+            'mention_type_enum' => MentionTypeEnum::class,
+            'created_at' => 'datetime',
+            'updated_at' => 'datetime',
+        ];
+    }
+
```
SirSplasch, line 19: add //Relations comment to be consistent to the other models

src/Models/Mention.php
```diff
@@ -0,0 +1,42 @@
...
+    public function source(): MorphTo
+    {
+        return $this->morphTo('source', 'mention_source_type', 'mention_source_id');
+    }
+
+    public function target(): MorphTo
+    {
+        return $this->morphTo('target', 'mention_target_type', 'mention_target_id');
```
SirSplasch: If you pass the name "correctly" laravel can guess the right columns for the morph.
"mention_source" and "mention_target" if I am not mistaken.

src/Support/Mentions/MentionState.php
```diff
@@ -0,0 +1,34 @@
...
+        public string $color,
+    ) {}
+
+    public function cssColor(): string
+    {
+        if (preg_match('/^[a-z]+$/', $this->color) !== 1) {
+            return 'currentColor';
+        }
+
+        return 'var(--color-' . $this->color . '-' . self::COLOR_SHADE . ')';
```
SirSplasch: static::COLOR_SHADE to allow customization

src/Traits/Model/RecordsMentions.php
```diff
@@ -0,0 +1,37 @@
...
+     * @return Collection<int, object>|null
+     */
+    public function mentionableMembersScope(): ?Collection
+    {
+        return null;
+    }
+
+    public function mentionScannableText(): string
+    {
+        return collect($this->mentionableTextFields())
```
SirSplasch: Why does the default implementation differ from the implementation used in the comment model? Shouldn't it be the same, and therefore is not needed to be overwritten in the comment model?

src/Traits/Model/Mentionable.php
```diff
@@ -0,0 +1,73 @@
...
+    {
+        return morph_alias(static::class);
+    }
+
+    public static function mentionTypeLabel(): string
+    {
+        return Str::headline(morph_alias(static::class));
+    }
+
+    public static function mentionTypeIcon(): string
```
SirSplasch: mentionTypeIcon before mentionTypeKey to keep alphabetical ordering.

src/Traits/Model/Mentionable.php
```diff
@@ -0,0 +1,73 @@
...
+
+        return static::query()
+            ->where('name', 'like', $trimmed . '%')
+            ->limit($limit)
+            ->get();
+    }
+
+    public function getMentionLabel(): string
+    {
+        if ($this instanceof InteractsWithDataTables) {
```
SirSplasch: I would rather check for method_exists($this, 'getLabel')

src/Traits/Model/Mentionable.php
```diff
@@ -0,0 +1,73 @@
...
+            return $this->getLabel() ?? '#' . $this->getKey();
+        }
+
+        return (string) ($this->name ?? $this->getKey());
+    }
+
+    public function getMentionUrl(): ?string
+    {
+        return $this instanceof InteractsWithDataTables ? $this->getUrl() : null;
```
SirSplasch: same here, method_exists()

src/Traits/Model/Mentionable.php
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    public function getMentionState(): ?MentionState
+    {
+        $state = $this->state ?? null;
```
SirSplasch: Better add a method which returns the state column, so that this is no blind guess on $this->state, as the column can differ.

src/Traits/Model/Mentionable.php
```diff
@@ -0,0 +1,73 @@
...
+
+        if (! $state instanceof State) {
+            return null;
+        }
+
+        return new MentionState(__(Str::headline((string) $state)), $state->color());
```
SirSplasch: app() or make the MentionState makeable and call MentionState::make()

src/Traits/Model/Mentionable.php
```diff
@@ -0,0 +1,73 @@
...
+            ->get();
+    }
+
+    public function getMentionLabel(): string
+    {
+        if ($this instanceof InteractsWithDataTables) {
+            return $this->getLabel() ?? '#' . $this->getKey();
+        }
+
+        return (string) ($this->name ?? $this->getKey());
```
SirSplasch: Return the same as the fallback in the if above, to be consistent, no blind guess on $this->name.

src/Traits/Model/RecordsMentions.php
```diff
@@ -0,0 +1,37 @@
...
+
+        static::deleted(function (Model&MentionsContent $model): void {
+            app(RecordsMentionsObserver::class)->deleted($model);
+        });
+    }
+
+    /**
+     * @return Collection<int, object>|null
+     */
+    public function mentionableMembersScope(): ?Collection
```
SirSplasch: What is this for? Is this really needed? I don't fully grasp the purpose of these members shenanigans.

src/Livewire/Support/Comments.php
```diff
@@ -114,10 +115,22 @@ function ($query): void {
...
+
+            $data = $comments->getCollection()->map(function ($comment) use ($refresher): Comment {
                 $comment->is_current_user = $comment->getRawOriginal('created_by')
                     === auth()->user()?->getMorphClass() . ':' . auth()->id();
 
+                if (is_string($comment->comment)) {
+                    $comment->comment = $refresher->refresh($comment->comment);
+                }
+
+                $comment->children->each(function ($child) use ($refresher): void {
```
SirSplasch: I don't think this is that easy, as this only covers the first set of children, but what happens to the children of the children and so on?

src/Support/Mentions/MentionPillRefresher.php
```diff
@@ -0,0 +1,70 @@
...
+class MentionPillRefresher
+{
+    protected const PILL = '/<(a|span)\b([^>]*\bdata-id="([a-z][a-z0-9_]*):(\d+)"[^>]*)>/i';
+
+    public function refresh(string $html): string
+    {
+        $userKey = morph_alias(User::class);
+        $types = MentionableTypes::map();
+
+        if (! preg_match_all(self::PILL, $html, $matches, PREG_SET_ORDER)) {
```
SirSplasch: static::PILL

src/Support/Mentions/MentionPillRefresher.php
```diff
@@ -0,0 +1,70 @@
...
+            return $html;
+        }
+
+        $idsByKey = [];
+        foreach ($matches as $m) {
```
SirSplasch: No singular character variable names

src/Support/Mentions/MentionPillRefresher.php
```diff
@@ -0,0 +1,70 @@
...
+            $key = strtolower($m[3]);
+            if ($key !== $userKey && isset($types[$key])) {
```
SirSplasch: array_key_exists($key, $types)?

src/Support/Mentions/MentionPillRefresher.php
```diff
@@ -0,0 +1,70 @@
...
+                $idsByKey[$key][] = (int) $m[4];
+            }
+        }
+
+        $recordsByKey = [];
+        foreach ($idsByKey as $key => $ids) {
+            $recordsByKey[$key] = $types[$key]::query()
```
SirSplasch: resolve_static()

src/Support/Mentions/MentionPillRefresher.php
```diff
@@ -0,0 +1,70 @@
...
+                ->whereKey(array_unique($ids))
+                ->get()
+                ->keyBy(fn ($record) => $record->getKey());
+        }
+
+        return preg_replace_callback(
+            self::PILL,
```
SirSplasch: static::PILL

src/Support/Mentions/MentionPillRefresher.php
```diff
@@ -0,0 +1,70 @@
...
+            function (array $m) use ($userKey, $types, $recordsByKey): string {
+                $tag = $m[1];
+                $attrs = $this->stripStateAttributes($m[2]);
+                $key = strtolower($m[3]);
+                $id = (int) $m[4];
+
+                if ($key === $userKey || ! isset($types[$key])) {
```
SirSplasch: array_key_exists()

src/Support/Mentions/MentionPillRefresher.php
```diff
@@ -0,0 +1,70 @@
...
+                $state = ($recordsByKey[$key][$id] ?? null)?->getMentionState();
+                $stateAttrs = $state?->toPillAttributes() ?? '';
+
+                return '<' . $tag . $attrs . $stateAttrs . '>';
+            },
+            $html,
+        ) ?? $html;
+    }
+
+    protected function stripStateAttributes(string $attrs): string
```
SirSplasch: string $attributes

src/Support/Mentions/MentionPillRefresher.php
```diff
@@ -0,0 +1,70 @@
...
+                ->whereKey(array_unique($ids))
+                ->get()
+                ->keyBy(fn ($record) => $record->getKey());
+        }
+
+        return preg_replace_callback(
+            self::PILL,
+            function (array $m) use ($userKey, $types, $recordsByKey): string {
+                $tag = $m[1];
+                $attrs = $this->stripStateAttributes($m[2]);
```
SirSplasch: $attributes

src/Support/Mentions/MentionPillRefresher.php
```diff
@@ -0,0 +1,70 @@
...
+                $key = strtolower($m[3]);
+                $id = (int) $m[4];
+
+                if ($key === $userKey || ! isset($types[$key])) {
+                    return $m[0];
+                }
+
+                $state = ($recordsByKey[$key][$id] ?? null)?->getMentionState();
+                $stateAttrs = $state?->toPillAttributes() ?? '';
```
SirSplasch: $stateAttributes

src/Support/Mentions/MentionPillRefresher.php
```diff
@@ -0,0 +1,70 @@
...
+        foreach ($idsByKey as $key => $ids) {
+            $recordsByKey[$key] = $types[$key]::query()
+                ->whereKey(array_unique($ids))
+                ->get()
+                ->keyBy(fn ($record) => $record->getKey());
+        }
+
+        return preg_replace_callback(
+            self::PILL,
+            function (array $m) use ($userKey, $types, $recordsByKey): string {
```
SirSplasch: No single character variable names

src/Support/Mentions/MentionPillRefresher.php
```diff
@@ -0,0 +1,70 @@
...
+use FluxErp\Facades\MentionableTypes;
+use FluxErp\Models\User;
+
+class MentionPillRefresher
+{
+    protected const PILL = '/<(a|span)\b([^>]*\bdata-id="([a-z][a-z0-9_]*):(\d+)"[^>]*)>/i';
+
+    public function refresh(string $html): string
+    {
+        $userKey = morph_alias(User::class);
```
SirSplasch: Use facade to get user key(s), can be multiple in theory.

SirSplasch (changes requested):
Check for the following in Support/Mentions files and the Observer
- customization issues
- no single character or unclear variable names
- allow multiple user mentions, as multiple "user" mentionable types can be registered

Maybe check tests which are useful and what not, or which are "outdated" with the changes that were made. I currently do not have the mental strength to go through all the tests with its fixtures.

src/Contracts/MentionsContent.php
Slupi: Renamed mentionableTextFields()/contentColumns → mentionableColumns()

src/Support/Mentions/MentionHtml.php
Slupi: MentionableType::getUserMentionableTypes()

src/Models/Mention.php
Slupi: Removed $timestamps, $table, $guarded

src/Models/Mention.php
Slupi: added

src/Models/Mention.php
Slupi: Relations renamed mentionSource()/mentionTarget() with bare morphTo(); columns inferred, matching migration.

src/Support/Mentions/MentionState.php
Slupi: Added protected const COLOR_SHADE = '500', used via static::

src/Traits/Model/RecordsMentions.php
Slupi: Removed the divergent default; mentionableColumns() is contract-required

src/Traits/Model/Mentionable.php
Slupi: reorderd

src/Traits/Model/Mentionable.php
Slupi: getMentionLabel() uses method_exists($this, 'getLabel')

src/Traits/Model/Mentionable.php
Slupi: getMentionUrl() uses method_exists($this, 'getUrl')

src/Traits/Model/Mentionable.php
Slupi: Added mentionStateColumn() (defaults 'state'); getMentionState() reads via it.

src/Traits/Model/Mentionable.php
Slupi: MentionState now use Makeable; callers use MentionState::make()

src/Traits/Model/Mentionable.php
Slupi: Both branches return '#' . $this->getKey()

src/Traits/Model/RecordsMentions.php
Slupi: Removed default mentionableMembersScope(); now optional, called by MentionSync only via method_exists() for scoping @-mentionable users.

src/Livewire/Support/Comments.php
Slupi: refreshMentionPills() now recurses through nested children when loaded.

src/Support/Mentions/MentionPillRefresher.php
Slupi: Pattern extracted to protected const PILL, used via static::PILL

src/Support/Mentions/MentionPillRefresher.php
Slupi: Renamed ($match, $key, $id)

src/Support/Mentions/MentionPillRefresher.php
Slupi: Uses array_key_exists($key, $types)

src/Support/Mentions/MentionPillRefresher.php
Slupi: Done

src/Support/Mentions/MentionPillRefresher.php
Slupi: Uses static::PILL in the second pass

src/Support/Mentions/MentionPillRefresher.php
Slupi: Uses static::PILL in the second pass

src/Support/Mentions/MentionPillRefresher.php
Slupi: Done

src/Support/Mentions/MentionPillRefresher.php
Slupi: Renamed to $attributes

src/Support/Mentions/MentionPillRefresher.php
Slupi: Renamed

src/Support/Mentions/MentionPillRefresher.php
Slupi: Done

src/Support/Mentions/MentionPillRefresher.php
Slupi: done

src/Support/Mentions/MentionableTypesManager.php
Slupi: Done. Renamed to singular MentionableTypeManager

src/Support/Mentions/MentionableTypesManager.php
Slupi: Done — MentionableTypeManager now mirrors WidgetManager: autoDiscover() called from a new MentionableTypeServiceProvider, register()/unregister(), all() plus getRecordMentionableTypes()/getUserMentionableTypes(), each with a bool $keysOnly = false param (replaces recordKeys()), backed by protected array $record = [] / $user = []

src/Facades/MentionableTypes.php
Slupi: Done

src/Traits/Model/Mentionable.php
```diff
@@ -0,0 +1,84 @@
...
+            return $this->getLabel() ?? '#' . $this->getKey();
+        }
+
+        return '#' . $this->getKey();
+    }
+
+    public function getMentionState(): ?MentionState
+    {
+        $column = $this->mentionStateColumn();
+        $state = $column !== null ? ($this->{$column} ?? null) : null;
```
SirSplasch: blank($column) instead of !== null check. As empty string is probably not a valid value.

src/Traits/Model/Mentionable.php
```diff
@@ -0,0 +1,84 @@
...
+
+        return MentionState::make(__(Str::headline((string) $state)), $state->color());
+    }
+
+    public function getMentionUrl(): ?string
+    {
+        return method_exists($this, 'getUrl') ? $this->getUrl() : null;
+    }
+
+    public function mentionStateColumn(): ?string
```
SirSplasch: Can this method be protected? Or is it better to be public in your opinion?

src/Support/Mentions/MentionableTypeManager.php
```diff
@@ -0,0 +1,90 @@
...
+        if ($class::mentionType() === MentionTypeEnum::User) {
+            $this->user[$key] = $class;
+        } else {
+            $this->record[$key] = $class;
+        }
+    }
+
+    public function unregister(string $key): void
+    {
+        unset($this->record[$key], $this->user[$key]);
```
SirSplasch: Add a control flag to this method, which controls on which array the key should be removed, either record, user, or both.

src/Support/Mentions/MentionableTypeManager.php
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    /**
+     * @return ($keysOnly is true ? array<int, string> : array<string, class-string>)
+     */
+    public function all(bool $keysOnly = false): array
+    {
+        $types = $this->record + $this->user;
```
SirSplasch: In theory the same key can exist in both arrays, handle this case.
Currently if this happens the user key will be silently ignored.

src/Support/Mentions/MentionableTypeManager.php
```diff
@@ -0,0 +1,90 @@
...
+
+    public function unregister(string $key): void
+    {
+        unset($this->record[$key], $this->user[$key]);
+    }
+
+    /**
+     * @return ($keysOnly is true ? array<int, string> : array<string, class-string>)
+     */
+    public function all(bool $keysOnly = false): array
```
SirSplasch, line 79: How does one know which value in the returned array is a record and which a user mentionable type?

src/Support/Mentions/MentionableTypeManager.php
```diff
@@ -0,0 +1,90 @@
...
+     */
+    protected array $user = [];
+
+    public function autoDiscover(): void
+    {
+        if (! function_exists('get_models_with_trait')) {
+            return;
+        }
+
+        foreach (get_models_with_trait(Mentionable::class, fn (string $class): string => $class) as $class) {
```
SirSplasch: get_models_with_trait() does not work for customized models as only the classes which are directly registered in the morph map are checked, but not overridden classes which use the same morph key.

src/Support/Mentions/MentionableTypeManager.php
```diff
@@ -0,0 +1,90 @@
...
+    {
+        if (! in_array(Mentionable::class, class_uses_recursive($class), true)) {
+            throw new InvalidArgumentException(
+                "The provided class '{$class}' does not use the Mentionable trait."
+            );
+        }
+
+        $key = $class::mentionTypeKey();
+
+        if ($class::mentionType() === MentionTypeEnum::User) {
```
SirSplasch, line 58: You cannot simply use the passed class string, if this model is for example customized, the results can differ, but the base model may never be actually used.
resolve_static($class, 'class')

src/Traits/Model/RecordsMentions.php
```diff
@@ -0,0 +1,29 @@
...
+
+        static::deleted(function (Model&MentionsContent $model): void {
+            app(RecordsMentionsObserver::class)->deleted($model);
+        });
+    }
+
+    public function mentionScannableText(): string
+    {
+        return collect($this->mentionableColumns())
+            ->map(fn (string $column): string => MentionHtml::toTokens((string) $this->getAttribute($column)))
```
SirSplasch: resolve_static(MentionHtml::class, 'toTokens')

src/Support/Mentions/MentionSync.php
```diff
@@ -0,0 +1,93 @@
...
+namespace FluxErp\Support\Mentions;
+
+use FluxErp\Contracts\MentionsContent;
+use FluxErp\Models\Mention;
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Support\Collection;
+
+class MentionSync
+{
+    public function __construct(private readonly MentionParser $parser) {}
```
SirSplasch: protected instead of private

src/Support/Mentions/MentionParser.php
```diff
@@ -0,0 +1,139 @@
...
+            }
+
+            $out[] = [
+                'type' => MentionTypeEnum::Record,
+                'user_id' => null,
+                'mentionable_type' => $key,
+                'mentionable_id' => (int) $matches[2][$index][0],
+                'raw' => $fullMatch[0],
+            ];
+            $consumed[] = [$fullMatch[1], $fullMatch[1] + strlen($fullMatch[0])];
```
SirSplasch: wrap the array
$consumed[] = [
    ...,
    ...,
];

src/Support/Mentions/MentionParser.php
```diff
@@ -0,0 +1,139 @@
...
+                $id = (int) $matches[2][$index][0];
+
+                $out[] = [
+                    'type' => MentionTypeEnum::User,
+                    'user_id' => ($userTypes[$key] ?? null) === User::class ? $id : null,
+                    'mentionable_type' => $key,
+                    'mentionable_id' => $id,
+                    'raw' => $fullMatch[0],
+                ];
+                $consumed[] = [$fullMatch[1], $fullMatch[1] + strlen($fullMatch[0])];
```
SirSplasch: wrap the array

src/Support/Mentions/MentionParser.php
```diff
@@ -0,0 +1,139 @@
...
+
+        preg_match_all(
+            '/(?<!\\\\)(?<![A-Za-z0-9._-])@([A-Za-z0-9._-]+)/',
+            $stripped,
+            $matches,
+            PREG_OFFSET_CAPTURE,
+        );
+
+        $byFirstname = $members->groupBy(fn ($member) => strtolower((string) ($member->firstname ?? '')))->all();
+        $byCode = $members->keyBy(fn ($member) => strtolower((string) ($member->user_code ?? '')))->all();
```
SirSplasch: You are blind guessing firstname and user_code, idk. about that one, especially as $member can be anything.

src/Support/Mentions/MentionParser.php
```diff
@@ -0,0 +1,139 @@
...
+                    $out[] = $this->plain(MentionTypeEnum::User, (int) $member->id, $raw);
+                }
+
+                continue;
+            }
+
+            if (isset($byCode[$token])) {
+                $out[] = $this->plain(MentionTypeEnum::User, (int) $byCode[$token]->id, $raw);
+            }
+        }
```
SirSplasch: I think I get what this is for, but you need some control over that. Maybe by adding a UserCollection class and requiring a UserCollection rather than a Collection as $members.

src/Support/Mentions/MentionParser.php
```diff
@@ -0,0 +1,139 @@
...
+
+        return false;
+    }
+
+    /**
+     * @return array{type: string, user_id: ?int, mentionable_type: ?string, mentionable_id: ?int, raw: string}
+     */
+    protected function plain(string $type, ?int $userId, string $raw): array
```
SirSplasch: This is only used for the members stuff, seems a little out of place, as the $out array is consistently filled directly beforehand.

src/Support/Mentions/MentionSync.php
```diff
@@ -0,0 +1,93 @@
...
+    {
+        $text = $source->mentionScannableText();
+
+        $members = method_exists($source, 'mentionableMembersScope')
+            ? ($source->mentionableMembersScope() ?? collect())
+            : collect();
+
+        $parsed = $this->parser->parse($text, $members);
+
+        $existing = Mention::query()
```
SirSplasch: resolve_static()

src/Support/Mentions/MentionSync.php
```diff
@@ -0,0 +1,93 @@
...
+            ->where('mention_source_type', $source->getMorphClass())
+            ->where('mention_source_id', $source->getKey())
+            ->get();
+
+        $existingKeys = $existing->map(fn (Mention $mention) => $this->keyForExisting($mention))->all();
```
SirSplasch: wrap
->map()
->all();

src/Support/Mentions/MentionRenderer.php
```diff
@@ -0,0 +1,185 @@
...
+            $key = strtolower($rawKey);
+            if (array_key_exists($key, $types)) {
+                $idsByKey[$key][] = (int) $matches[2][$index];
+            }
+        }
+
+        $recordsByKey = [];
+        foreach ($idsByKey as $key => $ids) {
+            $recordsByKey[$key] = resolve_static($types[$key], 'query')
+                ->whereKey(array_unique($ids))
```
SirSplasch: No need for an array_unique(), the query will not return the same record more than once.

src/Support/Mentions/MentionRenderer.php
```diff
@@ -0,0 +1,185 @@
...
+                    return $matches[0];
+                }
+
+                $record = $recordsByKey[$key][$id] ?? null;
+                if ($record === null) {
+                    return '<span class="mention mention--missing">' . e(__('@deleted entry')) . '</span>';
+                }
+
+                $label = e($record->getMentionLabel());
+                $typeLabel = e($types[$key]::mentionTypeLabel());
```
SirSplasch: resolve_static($types[$key], 'mentionTypeLabel')

src/Support/Mentions/MentionRenderer.php
```diff
@@ -0,0 +1,185 @@
...
+
+        $idsByKey = [];
+        foreach ($matches[1] as $index => $rawKey) {
+            $idsByKey[strtolower($rawKey)][] = (int) $matches[2][$index];
+        }
+
+        $usersByKey = [];
+        foreach ($idsByKey as $key => $ids) {
+            $usersByKey[$key] = resolve_static($userTypes[$key], 'query')
+                ->whereKey(array_unique($ids))
```
SirSplasch: no need for array_unique()

src/Support/Mentions/MentionRenderer.php
```diff
@@ -0,0 +1,185 @@
...
+                    return '<span class="mention mention--missing">' . e(__('@deleted entry')) . '</span>';
+                }
+
+                return $this->renderUserPill($user, $key);
+            },
+            $text,
+        ) ?? $text;
+    }
+
+    protected function renderMemberTokens(string $text, Collection $members): string
```
SirSplasch: Consider requiring a UserCollection rather than a simple collection.

src/Support/Mentions/MentionPillRefresher.php
```diff
@@ -0,0 +1,69 @@
...
+            $key = strtolower($match[3]);
+            if (! in_array($key, $userKeys, true) && array_key_exists($key, $types)) {
+                $idsByKey[$key][] = (int) $match[4];
+            }
+        }
+
+        $recordsByKey = [];
+        foreach ($idsByKey as $key => $ids) {
+            $recordsByKey[$key] = resolve_static($types[$key], 'query')
+                ->whereKey(array_unique($ids))
```
SirSplasch: no need for array_unique()

src/Observers/RecordsMentionsObserver.php
```diff
@@ -0,0 +1,100 @@
...
+use FluxErp\Enums\MentionTypeEnum;
+use FluxErp\Models\Mention;
+use FluxErp\Models\User;
+use FluxErp\Notifications\MentionNotification;
+use FluxErp\Support\Mentions\MentionSync;
+use Illuminate\Database\Eloquent\Model;
+
+class RecordsMentionsObserver
+{
+    public function __construct(private readonly MentionSync $sync) {}
```
SirSplasch: protected

src/Observers/RecordsMentionsObserver.php
```diff
@@ -0,0 +1,100 @@
...
+        $result = $this->sync->sync($source);
+
+        foreach ($result->added as $row) {
+            $this->dispatchForAddedRow($source, $row);
+        }
+    }
+
+    public function deleted(Model&MentionsContent $source): void
+    {
+        Mention::query()
```
SirSplasch: resolve_static()

src/Observers/RecordsMentionsObserver.php
```diff
@@ -0,0 +1,100 @@
...
+    public function saved(Model&MentionsContent $source): void
+    {
+        $result = $this->sync->sync($source);
+
+        foreach ($result->added as $row) {
+            $this->dispatchForAddedRow($source, $row);
+        }
+    }
+
+    public function deleted(Model&MentionsContent $source): void
```
SirSplasch, line 31: Are you intentionally not unsubscribing the notification channel that might have been subscribed to by the mention?

src/Observers/RecordsMentionsObserver.php
```diff
@@ -0,0 +1,100 @@
...
+    /**
+     * @param  array<string, mixed>  $row
+     */
+    protected function dispatchForAddedRow(Model $source, array $row): void
+    {
+        $type = $row['mention_type_enum'];
+
+        if ($type === MentionTypeEnum::User) {
+            $notifiable = $this->resolveTarget($row)
+                ?? ($row['user_id'] ? User::query()->whereKey($row['user_id'])->first() : null);
```
SirSplasch: resolve_static() and you potentially have to wrap the code then, to not exceed the maximum line length.

src/Observers/RecordsMentionsObserver.php
```diff
@@ -0,0 +1,100 @@
...
+
+            if ($notifiable && method_exists($notifiable, 'notify')) {
+                $notification = $source instanceof ProvidesMentionNotification
+                    ? $source->mentionNotification()
+                    : new MentionNotification($source);
```
SirSplasch: Allow customization of the MentionNotification, do not instantiate it with new

src/Observers/RecordsMentionsObserver.php
```diff
@@ -0,0 +1,100 @@
...
+    /**
+     * @param  array<string, mixed>  $row
+     */
+    protected function resolveTarget(array $row): ?Model
+    {
+        if (! $row['mention_target_type']) {
+            return null;
+        }
+
+        $targetClass = function_exists('morphed_model') ? morphed_model($row['mention_target_type']) : null;
```
SirSplasch: no need to check if this function exists, this package implements it.

src/Observers/RecordsMentionsObserver.php
```diff
@@ -0,0 +1,100 @@
...
+
+        return $targetClass
+            ? $targetClass::query()->whereKey($row['mention_target_id'])->first()
```
SirSplasch: wrap
->whereKey()
->first()

src/Observers/RecordsMentionsObserver.php
```diff
@@ -0,0 +1,100 @@
...
+
+        $target = $this->resolveTarget($row);
+        if (! $target) {
+            return;
+        }
+
+        $creator = auth()->user();
+        $canMention = $creator
+            && method_exists($creator, 'can')
+            && $creator->can('mention', $target);
```
SirSplasch: How does this work?
In this package, we do not use can(), we use checkPermissionTo() to check if the user has the permission to do something. The issue here is, I don't see it being anywhere "defined" whether the authenticated user can or cannot mention a specific target.

src/Observers/RecordsMentionsObserver.php
```diff
@@ -0,0 +1,100 @@
...
+        if (! $canMention) {
+            return;
+        }
+
+        if (method_exists($target, 'subscribeNotificationChannel')) {
+            $target->subscribeNotificationChannel(
+                channel: method_exists($source, 'broadcastChannel')
+                    ? $source->broadcastChannel()
+                    : $source->getMorphClass() . '.' . $source->getKey(),
+                event: 'eloquent.created: ' . $source::class,
```
SirSplasch: This event name may not be correct, as the saved() method does not differ between created or updated.

src/Support/Mentions/MentionSync.php
```diff
@@ -0,0 +1,93 @@
...
+        if ($toCreate !== []) {
+            Mention::insert($toCreate);
+        }
+
+        return (object) ['added' => $toCreate, 'removed' => $toDelete];
+    }
+
+    protected function keyForExisting(Mention $mention): string
+    {
+        $type = is_object($mention->mention_type_enum)
```
SirSplasch: No need to do this is_object() check, the model has a cast on mention_type_enum and therefore the expected value should be (string) $mention->mention_type_enum?->value ?? ''.
So this variable is not needed, and instead the value can be passed directly into the implode() function.

src/Support/Mentions/MentionSync.php
```diff
@@ -0,0 +1,93 @@
...
+            ];
+        }
+
+        $toDelete = $existing->reject(
+            fn (Mention $mention) => in_array($this->keyForExisting($mention), $parsedKeys, true)
+        );
+
+        if ($toDelete->isNotEmpty()) {
+            Mention::whereKey($toDelete->pluck('id'))->delete();
+        }
```
SirSplasch, line 64: new line after }

src/Support/Mentions/MentionSync.php
```diff
@@ -0,0 +1,93 @@
...
+        if ($toCreate !== []) {
+            Mention::insert($toCreate);
+        }
+
+        return (object) ['added' => $toCreate, 'removed' => $toDelete];
```
SirSplasch: Why do you need to cast this into an object? An array is in general easier to work with.

src/Support/Mentions/MentionSync.php
```diff
@@ -0,0 +1,93 @@
...
+                'updated_at' => now(),
+            ];
+        }
+
+        $toDelete = $existing->reject(
+            fn (Mention $mention) => in_array($this->keyForExisting($mention), $parsedKeys, true)
+        );
+
+        if ($toDelete->isNotEmpty()) {
+            Mention::whereKey($toDelete->pluck('id'))->delete();
```
SirSplasch: resolve_static()

src/Support/Mentions/MentionSync.php
```diff
@@ -0,0 +1,93 @@
...
+        }
+        if ($toCreate !== []) {
+            Mention::insert($toCreate);
```
SirSplasch: resolve_static()

src/Http/Controllers/MentionableSearchController.php
```diff
@@ -0,0 +1,100 @@
...
+use Throwable;
+
+class MentionableSearchController extends Controller
+{
+    protected static bool $hasPermission = false;
+
+    public function __invoke(Request $request): JsonResponse
+    {
+        $request->validate([
+            'q' => 'string|nullable',
```
SirSplasch: query instead of q

src/Http/Controllers/MentionableSearchController.php
```diff
@@ -0,0 +1,100 @@
...
+            'types' => 'array',
+            'types.*' => 'string',
+        ]);
+
+        $typesByKey = MentionableType::all();
+        $requested = collect($request->input('types', array_keys($typesByKey)));
+
+        $unknown = $requested->reject(fn (string $key): bool => isset($typesByKey[$key]));
```
SirSplasch: array_key_exists()

src/Http/Controllers/MentionableSearchController.php
```diff
@@ -0,0 +1,100 @@
...
+        if ($unknown->isNotEmpty()) {
+            abort(422, 'Unknown mentionable types: ' . $unknown->implode(', '));
+        }
+
+        $query = trim((string) $request->input('q', ''));
```
SirSplasch: no need to put empty string as a fallback, this will be cast into string, and (string) null == ''

src/Http/Controllers/MentionableSearchController.php
```diff
@@ -0,0 +1,100 @@
...
+        $scope = null;
+        if (preg_match('/^([A-Za-z_]+):(.*)$/s', $query, $matches)) {
+            $candidateKey = strtolower($matches[1]);
+            if ($requested->contains($candidateKey)) {
+                $scope = $candidateKey;
+                $query = trim($matches[2]);
+            }
+        }
+
+        if ($query === '' && $scope === null) {
```
SirSplasch: is_null($scope)

src/Http/Controllers/MentionableSearchController.php
```diff
@@ -0,0 +1,100 @@
...
+            return response()->json($this->scopeChips($requested, $typesByKey));
+        }
+
+        $searchTypes = $scope !== null ? collect([$scope]) : $requested;
```
SirSplasch: ! is_null($scope)

src/Http/Controllers/MentionableSearchController.php
```diff
@@ -0,0 +1,100 @@
...
+        $user = $request->user();
+        $results = collect();
+        $userKeys = MentionableType::getUserMentionableTypes(keysOnly: true);
+
+        foreach ($searchTypes as $key) {
+            $class = $typesByKey[$key];
+
+            try {
+                $candidates = $class::searchMentionCandidates($query, 5)
```
SirSplasch: resolve_static()

src/Http/Controllers/MentionableSearchController.php
```diff
@@ -0,0 +1,100 @@
...
+                    ->filter(fn ($record): bool => $user ? $user->can('view', $record) : true)
```
SirSplasch: How does this work? This is the same as the "can('mention', $target)" issue. We use checkPermission() to check if a user can do something. But I don't think what you want to achieve is currently covered, but I may be wrong about that.

src/Http/Controllers/MentionableSearchController.php
```diff
@@ -0,0 +1,100 @@
...
+        if ($requested->count() < 2) {
+            return [];
+        }
+
+        return $requested
+            ->map(fn (string $key): array => [
+                'kind' => 'scope',
+                'scope_key' => $key,
+                'label' => $typesByKey[$key]::mentionTypeLabel(),
+                'type_icon' => $typesByKey[$key]::mentionTypeIcon(),
```
SirSplasch: resolve_static()

SirSplasch (changes requested):
Also resolve merge conflicts.

src/Traits/Model/Mentionable.php
Slupi: done

src/Traits/Model/Mentionable.php
Slupi: Made it protected — it's only called from getMentionState(), nothing references it externally

src/Support/Mentions/MentionableTypeManager.php
Slupi: Done — unregister(string $key, ?string $mentionType = null); MentionTypeEnum::Record/::User removes from one array, null from both.

src/Support/Mentions/MentionableTypeManager.php
Slupi: Handled — register() now calls unregister($key) before inserting, so the last registration wins and a key can never exist in both arrays.

src/Support/Mentions/MentionableTypeManager.php
Slupi: A key can now only live in one of the two arrays (register() unregisters the key first, see the thread above), and getRecordMentionableTypes() / getUserMentionableTypes() give the typed subsets. all() is only used where the distinction doesn't matter (key existence checks, scope chips).

src/Support/Mentions/MentionableTypeManager.php
Slupi: autoDiscover() now iterates Relation::morphMap() directly and checks the trait on resolve_static($class, 'class'), so overridden classes behind the same morph key are picked up.

src/Http/Controllers/MentionableSearchController.php
Slupi: done

src/Http/Controllers/MentionableSearchController.php
Slupi: It didn't work out of the box, correct: core registers no policies and the only Gate::before is the Super Admin bypass, so a plain can('view', $record) denied everything for non-super-admins. checkPermissionTo() doesn't fit here either — that resolves named permissions, and there is no per-record permission covering mention visibility.

Changed the filter to be policy-gated: it only applies when Gate::getPolicyFor($record) returns a policy, otherwise the record passes. So vanilla core doesn't filter, and a project that registers a policy for a mentionable model gets its search results filtered by its view method. Covered in MentionableSearchControllerTest with a fixture policy.

src/Http/Controllers/MentionableSearchController.php
Slupi: done

src/Http/Controllers/MentionableSearchController.php
Slupi: done

src/Http/Controllers/MentionableSearchController.php
Slupi: done

src/Http/Controllers/MentionableSearchController.php
Slupi: done, fallback dropped

src/Http/Controllers/MentionableSearchController.php
Slupi: done

src/Http/Controllers/MentionableSearchController.php
Slupi: done, renamed everywhere

src/Support/Mentions/MentionSync.php
Slupi: done

src/Support/Mentions/MentionSync.php
Slupi: done

src/Support/Mentions/MentionSync.php
Slupi: Changed to an array (['added' => …, 'removed' => …]); observer and tests updated

src/Support/Mentions/MentionSync.php
Slupi: done

src/Support/Mentions/MentionSync.php
Slupi: done

src/Observers/RecordsMentionsObserver.php
Slupi: The subscription targets future events on the source's channel, so whether the current save was a create or an update doesn't change it — but eloquent.created is only the right event for comment-like sources (next comment on the same channel). For sources where updates are the relevant signal it would be a dead subscription. Added an overridable mentionSubscriptionEvent() to the RecordsMentions trait (default stays eloquent.created: <class>, matching the old Comment::booted() behavior); the observer uses it for subscribe and unsubscribe.

src/Observers/RecordsMentionsObserver.php
Slupi: Same situation as the can('view') thread on the search controller: nothing defines this ability in core, so with no policy registered the Gate denies it for everyone except Super Admins — record mentions never subscribed in a vanilla install. Made it policy-gated: canMention() only enforces can('mention', $target) when there is an authenticated user and Gate::getPolicyFor($target) returns a policy; otherwise it allows, which matches the pre-consolidation Comment behavior. Projects opt into enforcement by registering a policy with a mention method.

src/Observers/RecordsMentionsObserver.php
Slupi: done

src/Observers/RecordsMentionsObserver.php
Slupi: removed

src/Observers/RecordsMentionsObserver.php
Slupi: Now resolved via app(MentionNotification::class, ['source' => $source])

src/Observers/RecordsMentionsObserver.php
Slupi: done

src/Observers/RecordsMentionsObserver.php
Slupi: No, that was a gap. saved() now unsubscribes the targets of removed record mentions and deleted() unsubscribes all record-mention targets before deleting the rows. Subscribe and unsubscribe share the same subscriptionChannel()/subscriptionEvent() helpers so they always match.

src/Observers/RecordsMentionsObserver.php
Slupi: done

src/Observers/RecordsMentionsObserver.php
Slupi: done

src/Support/Mentions/MentionPillRefresher.php
Slupi: removed

src/Support/Mentions/MentionRenderer.php
Slupi: Done — same UserCollection requirement as the parser, and both now key members through the shared groupByFirstname() / keyByUserCode() methods (the renderer previously key'd by firstname last-wins, which silently dropped duplicate firstnames).

src/Support/Mentions/MentionRenderer.php
Slupi: removed

src/Support/Mentions/MentionRenderer.php
Slupi: done

src/Support/Mentions/MentionRenderer.php
Slupi: removed

src/Support/Mentions/MentionSync.php
Slupi: done

src/Support/Mentions/MentionSync.php
Slupi: done

src/Support/Mentions/MentionParser.php
Slupi: $out is now filled inline in both member branches like everywhere else

src/Support/Mentions/MentionParser.php
Slupi: Added FluxErp\Support\Collection\UserCollection (same pattern as AddressCollection), User::newCollection() returns it, and MentionParser::parse() / MentionRenderer::tokensToHtml() now require it. The member keying (groupByFirstname() / keyByUserCode()) lives on the collection, so parser and renderer share one implementation.

src/Support/Mentions/MentionParser.php
Slupi: $members is now a UserCollection and the keying moved onto it as groupByFirstname() / keyByUserCode() with fn (User $user) closures — so these are typed User columns instead of guesses against an untyped collection

src/Support/Mentions/MentionParser.php
Slupi: done

src/Support/Mentions/MentionParser.php
Slupi: done

src/Support/Mentions/MentionSync.php
Slupi: done

src/Traits/Model/RecordsMentions.php
Slupi: done

src/Support/Mentions/MentionableTypeManager.php
Slupi: done — register() resolves the class first

src/Support/Mentions/MentionableTypeManager.php
```diff
@@ -0,0 +1,101 @@
...
+            unset($this->user[$key]);
+        }
+    }
+
+    /**
+     * @return ($keysOnly is true ? array<int, string> : array<string, class-string>)
+     */
+    public function all(bool $keysOnly = false): array
+    {
+        $types = $this->record + $this->user;
```
SirSplasch: I would still prefer using array_merge() rather than array concatination.

SirSplasch (changes requested):

Slupi:
Extracted the policy gating into its own track as discussed: the policy infrastructure now lives in #1995 (to be merged first), and this PR no longer touches authorization. The mention search/subscription gating returns as a small follow-up on top of #1995, checking view instead of a dedicated mention ability.

src/Support/Mentions/MentionableTypeManager.php
Slupi: done

SirSplasch (approved):

SirSplasch:
resolve merge conflicts

PR #1792, opened 2026-06-10
Prevent child amount reset when parent amount is cleared

patrickweh:
## Problem

Clearing the amount on a free-text group parent position (or setting it to 0/null) divided all child position amounts by the previous parent amount. The propagation introduced in 80028a2a4 coerces a zero/empty new amount to 1 and recalculates ratio-based children from `child amount / previous amount`, effectively shrinking every child. Example: parent amount 4 with children 4/4/16 ends up with children 1/1/4 after the parent amount is cleared.

## Fix

`UpdateOrderPosition` now only propagates amount changes to children when the new amount is positive. Ratio-based children (no `amount_bundle`) are only rescaled when a positive previous amount exists as a baseline; otherwise they are left untouched. Negative amounts never trigger a rescale.

## Tests

- clearing amount (0 and null) on a free-text group parent keeps child amounts
- proportional rescale on positive amount change still works
- bundle children with `amount_bundle` still rescale, but not on cleared amount
- negative new amount and non-positive previous amount never rescale children

## Summary by Sourcery

Adjust order position amount propagation so child positions are only rescaled when the parent amount changes to a positive value, avoiding unintended shrinking when amounts are cleared or non-positive.

Bug Fixes:
- Prevent child order position amounts from being reduced when a free-text group parent amount is cleared or set to zero or null.
- Ensure negative parent amounts and non-positive previous amounts do not trigger proportional rescaling of child positions.

Tests:
- Add feature tests covering clearing and changing amounts on free-text group parents and verifying proportional rescaling only occurs for positive amounts.
- Add feature tests ensuring bundle children rescale based on amount_bundle only on positive parent amount changes and remain unchanged when the parent amount is cleared.
- Add feature tests covering negative and non-positive previous amounts to confirm they do not act as a baseline for rescaling.

### DIFF
--- a/src/Actions/OrderPosition/UpdateOrderPosition.php
+++ b/src/Actions/OrderPosition/UpdateOrderPosition.php
@@ -89,24 +89,26 @@ class UpdateOrderPosition extends FluxAction
         unset($orderPosition->discounts, $orderPosition->unit_price);
         $orderPosition->save();
 
-        if ($orderPosition->wasChanged('amount')) {
+        if ($orderPosition->wasChanged('amount') && bccomp($orderPosition->amount ?? 0, 0) === 1) {
             $children = resolve_static(OrderPosition::class, 'query')
                 ->where('parent_id', $orderPosition->getKey())
                 ->whereNotNull('amount')
                 ->get();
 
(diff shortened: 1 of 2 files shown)


patrickweh:
Replaced by #1794 (branch renamed).

PR #1797, opened 2026-06-11
feat(mail): broadcast sync progress toast to the triggering user

patrickweh:
## What

When a user clicks "get new messages" in the /mail module, the sync jobs were dispatched fire-and-forget with no feedback. This adds the same toast progress feedback that mail sending already has.

## How

`SyncMailAccountJob` now implements `ShouldBeMonitored` and uses the `IsMonitored` trait, so it runs through the existing QueueMonitor pipeline: started, processing (with progress bar and remaining time), and finished toasts are broadcast via the user's private channel.

- One toast per account, matching the one-job-per-account dispatch in `Mail::getNewMessages()`
- `getName()` includes the mail account email so each toast identifies its mailbox
- Progress advances per synced folder via `queueProgressChunk()`
- Scheduled runs have no authenticated user, so the monitor is attached to nobody and no toast is sent

## Summary by Sourcery

Integrate mail account sync jobs with the queue monitoring pipeline to provide user-facing progress feedback for manual mail syncs.

New Features:
- Add monitoring and progress tracking for mail account sync jobs so users see toast notifications when manually triggering mail synchronization.

Tests:
- Add feature tests to verify that mail sync jobs are monitored and that the job name includes the associated mail account email.

### DIFF
--- a/src/Jobs/SyncMailAccountJob.php
+++ b/src/Jobs/SyncMailAccountJob.php
@@ -87,28 +87,30 @@ class SyncMailAccountJob implements Repeatable, ShouldBeMonitored, ShouldBeUniqu
             ->get();
         $total = $folders->count();
 
-        $folders->each(function (MailFolder $folder, int $index) use ($driver, $total): void {
+        try {
+            $folders->each(function (MailFolder $folder, int $index) use ($driver, $total): void {
+                if ($driver instanceof ReportsSyncProgress) {
+                    $driver->withProgressCallback(
+                        fn (int $processed, int $totalMessages) => $this->queueUpdate([
+                            'progress' => ($index + ($totalMessages > 0 ? min($processed / $totalMessages, 1) : 1))
+                                / $total
+                                * 100,
+                            'message' => __(':processed of :total mails (:folder)', [
+                                'processed' => $processed,
+                                'total' => $totalMessages,
+                                'folder' => $folder->name,
+                            ]),
+                        ])
+                    );
+                }
+
+                $driver->syncMessages($folder);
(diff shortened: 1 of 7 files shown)


tests/Unit/Mail/ImapMessageBuilderTest.php
```diff
@@ -70,6 +70,24 @@
     expect($communication->refresh()->message_uid)->toBe('42');
 });
 
+it('reports progress for each stored message', function (): void {
```
SirSplasch: test

src/Jobs/SyncMailAccountJob.php
```diff
@@ -74,9 +82,39 @@ public function handle(): void
...
+                );
+            }
+
+            $driver->syncMessages($folder);
+            $this->queueProgressChunk($total, 1);
+        });
+
+        if ($driver instanceof ReportsSyncProgress) {
+            $driver->withProgressCallback(null);
+        }
```
SirSplasch: Why must the callback reset to null? Does it somehow persist?

SirSplasch (changes requested):

src/Jobs/SyncMailAccountJob.php
patrickweh: yes, `MailDriverManager` is a singleton and `Illuminate\Support\Manager` caches driver instances, so in a long-running worker the driver (and the callback holding a job reference) survives the job. moved the reset into a `finally` so it also runs when `syncMessages()` throws.

SirSplasch (approved):

PR #1806, opened 2026-06-12
fix(printing): allow page breaks inside order position descriptions

patrickweh:
## Summary

Order positions with long descriptions were pushed entirely to the next page in printed PDFs (offers, order confirmations, invoices), often leaving the first page partially or completely blank. dompdf cannot break a single table row across pages, so a position rendered as one row was atomic.

## Changes

- New `split_html_for_print()` helper that splits rendered HTML into top-level chunks. Lists are split into one list element per item, ordered list numbering is kept via the `start` attribute and vertical margins between list chunks are removed so they still read as one list.
- The `print.order.order-position` component now renders the position meta data (number, name, amount, price) in one row and each description chunk in its own table row, so dompdf can insert page breaks between them.
- Positions without a description render exactly as before. The zebra background and paddings are preserved across the chunk rows.

A single top-level element that is taller than one page still cannot be broken, but realistic descriptions consist of multiple paragraphs and list items.

### DIFF
--- a/src/Printing/HtmlPrintChunker.php
+++ b/src/Printing/HtmlPrintChunker.php
@@ -9,11 +9,6 @@ use DOMText;
 
 class HtmlPrintChunker
 {
-    protected function render(DOMNode $node): string
-    {
-        return (string) $node->ownerDocument->saveHTML($node);
-    }
-
     /**
      * Split rendered HTML into top-level chunks so the PDF renderer can
      * insert page breaks between them - dompdf cannot break a single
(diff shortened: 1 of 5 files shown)


src/Printing/HtmlPrintChunker.php
```diff
@@ -0,0 +1,147 @@
...
+namespace FluxErp\Printing;
+
+use DOMDocument;
+use DOMElement;
+use DOMNode;
+use DOMText;
+
+class HtmlPrintChunker
+{
+    protected function render(DOMNode $node): string
```
SirSplasch: consider rename this method, so that pint does not put it above public methods.

SirSplasch (changes requested):

SirSplasch (approved):

PR #1818, opened 2026-06-17
feat: "Show" dropdown on task widgets and bulk actions on the task list

patrickweh:
Adds a "Show" dropdown to the task widgets and bulk select actions to the task list, plus a generic job for running any FluxAction in bulk.

## Task widgets
- "My Tasks" and "My Responsible Tasks" now have an ellipsis "Show" dropdown next to the title (same pattern as chart widgets implementing `HasWidgetOptions`).
- Selecting it stores a `SessionFilter` scoped to the current user's tasks (assigned / responsible, excluding end states) and redirects to the task list, which opens pre-filtered. The filter label shown in the list is the widget label ("My Tasks" / "My Responsible Tasks").

## Task list bulk actions
- The task list is now selectable and offers two select actions:
  - **Delete** the selected tasks (with confirmation).
  - **Change state** of the selected tasks via a state-picker modal.
- Both are gated by the corresponding action permission.

## Generic bulk job
- New `ExecuteActionsJob(string $action, array $payloads, ?string $name)` runs any `FluxAction` over a set of payloads as a queued, monitored job with a progress toast. Payloads are hybrid: scalar ids (wrapped as `['id' => ...]`) or full data arrays. Per-item failures are reported and counted without aborting the batch.

## Finished-toast timeout
- Finished job toasts without an action now auto-expire after 30s; toasts carrying an action (e.g. a download link) stay persistent. Configurable per job via `toastPersistent()` / `toastTimeout()` on `IsMonitored`.
- The toast frontend now arms its auto-dismiss timer when a persistent progress toast is updated into a timed one (TallStackUI only arms it once on init).

## Summary by Sourcery

Add widget shortcuts to open a pre-filtered task list, introduce bulk actions on tasks via a generic monitored job, and refine job-finished toast behavior.

New Features:
- Add a "Show" dropdown to the My Tasks and My Responsible Tasks widgets that opens the main task list pre-filtered to the widget’s tasks, excluding end states.
- Introduce bulk delete and bulk state-change actions on the task list, driven by a reusable ExecuteActionsJob for running FluxAction classes over multiple payloads.

Enhancements:
- Make task list rows selectable and expose selected-row actions for changing state or deleting tasks, respecting existing action permissions.
- Add a modal-driven task state picker wired to the task list for bulk state updates.
- Extend monitored jobs with configuration helpers for toast persistence and timeout, and adjust finished job notifications to default to timed or persistent toasts based on presence of actions or overrides.
- Ensure TallStackUI toasts that transition from persistent to timed gain an auto-dismiss timer even when created without one.

Tests:
- Add feature and Livewire tests for task list selection and bulk actions, including validation of dispatched ExecuteActionsJob payloads.
- Add widget tests verifying the new Show dropdown, task-list redirection, and correct session filters for My Tasks and My Responsible Tasks widgets.
- Add unit tests for JobFinishedNotification toast timeout and persistence behavior, including explicit overrides.
- Add tests covering ExecuteActionsJob behavior for scalar IDs, full payload arrays, non-FluxAction classes, and empty payload sets.


src/Livewire/Task/TaskList.php
```diff
@@ -24,6 +30,8 @@ public function mount(): void
     {
         parent::mount();
 
+        $this->isSelectable = true;
```
SirSplasch: Why set isSelectable = true in mount, and not override the property with = true?

src/Jobs/ExecuteActionsJob.php
```diff
@@ -0,0 +1,73 @@
...
+        if ($total === 0) {
+            return;
+        }
+
+        $done = 0;
+        $failed = 0;
+
+        foreach (array_values($this->payloads) as $index => $payload) {
+            try {
+                $this->action::make(is_array($payload) ? $payload : ['id' => $payload])
```
SirSplasch: I don't know about this fallback, this can lead to unexpected behavior.
I think I would prefer an empty array as a fallback rather than this.

src/Jobs/ExecuteActionsJob.php
```diff
@@ -0,0 +1,73 @@
...
+            } catch (Throwable $e) {
+                report($e);
+
+                $failed++;
+            }
+
+            $this->queueProgress((int) (($index + 1) / $total * 100));
+        }
+
+        $this->message(trim(
```
SirSplasch: Do you need to trim this? As this is only relevant if the localization contains something that can be trimmed.

src/Jobs/ExecuteActionsJob.php
SirSplasch, line 1: I am not sure, but isn't this class name somewhat misleading? As the job only allows for a singular type of action to be executed over and over again. So its purpose is essentially to make a non-bulk action a bulk action.
And if this is the purpose isn't it "smarter" to implement a system that "defines" whether an action can be run as a bulk or not. And if this action is a DispatchableFluxAction it can be run as a job as well.

I understand that this current solution is the easy way, but do we want that?

src/Livewire/Task/TaskList.php
```diff
@@ -47,6 +55,90 @@ protected function getTableActions(): array
...
+
+        $this->js(<<<'JS'
+            $tsui.open.modal('change-task-state-modal');
+        JS);
+    }
+
+    #[Renderless]
+    public function changeState(): bool
+    {
+        if (! in_array($this->selectedState, Arr::pluck($this->availableStates, 'name'), true)) {
```
SirSplasch, line 111: No validation error if state is not valid?

SirSplasch (changes requested):

src/Livewire/Task/TaskList.php
```diff
@@ -47,6 +55,96 @@ protected function getTableActions(): array
...
+            exception_to_notifications(
+                ValidationException::withMessages([
+                    'selectedState' => [__('The selected state is invalid.')],
```
SirSplasch, line 114: add localization key to de.json

src/Support/Bus/BulkExecutor.php
```diff
@@ -0,0 +1,75 @@
...
+    public function __construct(
+        protected string $action,
+        protected array $payloads,
+    ) {}
+
+    /**
+     * @param  class-string  $action
+     * @param  array<int, array<string, mixed>>  $payloads
+     */
+    public static function make(string $action, array $payloads): static
```
SirSplasch: Consider using the Makeable trait instead of implementing this method here. (Also this current implementation is not safe for customization)

src/Support/Bus/BulkExecutor.php
```diff
@@ -0,0 +1,75 @@
...
+
+        if (blank($this->payloads)) {
+            return;
+        }
+
+        $jobs = array_map(
+            fn (array $payload) => $this->action::make($payload)
+                ->checkPermission()
+                ->validate(),
+            array_values($this->payloads),
```
SirSplasch: no need to do array_values() as only the values of the array are used and not its keys.

SirSplasch (changes requested):

src/Support/Bus/BulkExecutor.php
patrickweh: Switched BulkExecutor to the Makeable trait, so make() resolves through resolve_static/the container and stays customizable. Done in 9f24b76.

src/Support/Bus/BulkExecutor.php
patrickweh: Removed array_values(), payloads are passed directly. Done in 9f24b76.

src/Livewire/Task/TaskList.php
patrickweh: Added the 'The selected state is invalid.' key to lang/de.json. Done in 9f24b76.

SirSplasch (approved):

PR #1824, opened 2026-06-19
add bulk validate and execute to FluxAction

SirSplasch:
## Summary by Sourcery

Add bulk-aware validation and execution behavior to FluxAction when actions implement bulk execution support and receive list data.

New Features:
- Support executing actions in bulk when the action implements bulk execution and input data is a list.
- Support validating multiple payloads in bulk by iterating over each item and aggregating the prepared/validated data.

Enhancements:
- Ensure bulk executions and validations respect existing transaction handling and action event lifecycle while restoring original data structure after processing.

### DIFF
--- a/src/Actions/FluxAction.php
+++ b/src/Actions/FluxAction.php
@@ -212,36 +212,9 @@ abstract class FluxAction
         // transaction counter — breaking RefreshDatabase rollback in tests and
         // risking inconsistent state on deadlock retry in production.
         if (DB::transactionLevel() === 0) {
-            DB::transaction(
-                function () use ($isBulk) {
-                    if ($isBulk) {
-                        $bulk = $this->data;
-
-                        foreach ($bulk as $data) {
-                            $this->data = $data;
-                            $this->result[] = $this->performAction();
-                        }
-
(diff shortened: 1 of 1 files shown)


src/Actions/FluxAction.php
```diff
@@ -300,15 +303,30 @@ final public function validate(): static
             $this->setRulesFromRulesets();
         }
 
-        $this->fireActionEvent(event: 'preparingForValidation');
-        $this->prepareForValidation();
+        if ($isBulk = $this instanceof SupportsBulkExecution && array_is_list($this->data)) {
```
patrickweh: `array_is_list([])` returns `true`, so an empty payload on a `SupportsBulkExecution` action is treated as a zero-length bulk run: the loop iterates zero times, nothing is validated, and `execute()` returns `[]`. Empty/malformed input passes silently. Guard the empty case explicitly instead of relying on `array_is_list`.

patrickweh (changes requested):
One edge case to handle before merge — inline.

patrickweh (approved):

PR #1826, opened 2026-06-19
feat: per-menu unread notification indicator

patrickweh:
Adds a per-menu-area badge showing the count of unread notifications, kept in sync in real time over the existing broadcast socket.

## How it works
- `NotificationAction::route($name, $params)` now stores the route name in `data.accept.route` (in addition to the url).
- `Notification::menuArea()` maps that route name to the top level menu key (the segment before the first dot, e.g. `tickets.id` → `tickets`).
- `Navigation` groups the user's unread notifications by area for the badges, and on `mount()` marks the visited area's notifications as read so the badge clears (and the bell count drops with it).
- `Features\\Notifications` dispatches a shared `notifications-changed` event on every change (new notification via the existing Echo `.notification()` listener, accept, mark read), which re-renders the menu badges and recomputes the bell's unread count. No second socket.
- A notification can opt out of the menu indicator with `ToastNotification::hideFromMenu()` (or `menuIndicator(false)`); notifications without a routed target never produce a badge.

No new storage, no schema change, no url reverse-matching.

## Summary by Sourcery

Add per-menu unread notification indicators and keep notification-related UI in sync via a shared Livewire event.

New Features:
- Expose per-menu unread notification counts to the navigation view for rendering area-level badges.
- Allow toast notifications to opt out of contributing to menu unread indicators via a configurable flag.

Enhancements:
- Derive a notification’s menu area from its stored route name to group notifications by top-level menu section.
- Automatically mark notifications for the current menu area as read when the navigation component mounts, updating unread state across the UI.
- Broadcast a shared `notifications-changed` Livewire event from notification actions so other components can refresh their unread counts.

Tests:
- Add coverage for per-menu unread notification counts in the navigation component.
- Add tests ensuring notification components recompute unread counts when notifications change.
- Add unit tests for the `Notification` model’s `menuArea` derivation and opt-out behavior.

### DIFF
--- a/src/Livewire/Navigation.php
+++ b/src/Livewire/Navigation.php
@@ -102,9 +102,8 @@ class Navigation extends Component
         }
 
         return $user->unreadNotifications
-            ->groupBy(fn (Notification $notification): ?string => $notification->menuArea())
+            ->countBy(fn (Notification $notification): ?string => $notification->menuArea())
             ->forget('')
-            ->map->count()
             ->all();
     }
 
--- a/src/Livewire/Navigation.php
+++ b/src/Livewire/Navigation.php
@@ -3,24 +3,39 @@
 namespace FluxErp\Livewire;
 
 use FluxErp\Facades\Menu;
+use FluxErp\Models\Notification;
 use Illuminate\Contracts\Foundation\Application;
 use Illuminate\Contracts\View\Factory;
 use Illuminate\Contracts\View\View;
 use Illuminate\Support\Collection;
 use Illuminate\Support\Facades\Auth;
+use Illuminate\Support\Facades\Route;
 use Illuminate\Support\Facades\Session;
 use Illuminate\Support\Str;
+use Livewire\Attributes\On;
 use Livewire\Component;
 
 class Navigation extends Component
 {
+    public function mount(): void
+    {
+        $this->markAreaRead(Route::currentRouteName());
+    }
+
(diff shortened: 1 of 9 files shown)


src/Livewire/Navigation.php
```diff
@@ -49,6 +64,50 @@ public function deleteFavorite(int $id): void
...
+        $user = auth()->user();
+
+        if (! method_exists($user, 'unreadNotifications')) {
+            return [];
+        }
+
+        return $user->unreadNotifications
+            ->groupBy(fn (Notification $notification): ?string => $notification->menuArea())
+            ->forget('')
+            ->map->count()
```
SirSplasch: Is this not wrapped, as the count() is the mapping function?

src/Models/Notification.php
```diff
@@ -8,13 +8,23 @@
...
     use MassPrunable, ResolvesRelationsThroughContainer;
 
     // Public methods
+    public function menuArea(): ?string
+    {
+        if (data_get($this->data, 'menu_indicator') === false) {
+            return null;
+        }
+
+        return Str::before(data_get($this->data, 'accept.route', ''), '.') ?: null;
```
SirSplasch: data_get() ?? '' instead of passing '' as a default

SirSplasch (changes requested):

SirSplasch (approved):

PR #1827, opened 2026-06-19
feat: user access token endpoint

patrickweh:
Adds `POST /user/access-token` (under `auth:web`) that issues a personal access token owned by the currently authenticated user, implemented as a FluxAction (`CreateUserAccessToken`) with its own ruleset.

## Why
Native/desktop clients authenticate through the web login (password, password managers, passkeys, 2FA) and then need an API token. The existing `POST /api/tokens` creates a token-owned token whose permissions are a separate snapshot, so it does not reflect the user's permissions. This endpoint issues a **user-owned** token, so permission checks run against the user's live permissions and stay correct when they change.

## Details
- `name` required; optional `abilities` (defaults to `['user']`), validated against the same set as `CreateToken`.
- The action sets `$hasPermission = false` (no permission gate); access is controlled by the `auth:web` session.
- Returns `{ "token": "<plain-text-token>" }`.

## Summary by Sourcery

Add a web-authenticated endpoint for issuing user-owned API access tokens.

New Features:
- Introduce POST /user/access-token route under auth:web to create personal access tokens for the authenticated user.

Enhancements:
- Add a FluxAction (CreateUserAccessToken) and validation ruleset for creating user-owned access tokens with optional abilities configuration.

### DIFF
--- a/src/Rulesets/Token/CreateUserAccessTokenRuleset.php
+++ b/src/Rulesets/Token/CreateUserAccessTokenRuleset.php
@@ -1,0 +1,27 @@
+<?php
+
+namespace FluxErp\Rulesets\Token;
+
+use FluxErp\Rulesets\FluxRuleset;
+use Illuminate\Validation\Rule;
+
+class CreateUserAccessTokenRuleset extends FluxRuleset
+{
+    public function rules(): array
+    {
+        return [
+            'name' => 'required|string|max:255',
+            'abilities' => 'nullable|array',
+            'abilities.*' => [
+                'required',
+                'string',
+                Rule::in([
+                    '*',
+                    'interface',
+                    'user',
+                    'address',
+                ]),
+            ],
+        ];
+    }
+}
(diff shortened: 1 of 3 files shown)


src/Rulesets/Token/CreateUserAccessTokenRuleset.php
```diff
@@ -0,0 +1,27 @@
...
+{
+    public function rules(): array
+    {
+        return [
+            'name' => 'required|string|max:255',
+            'abilities' => 'nullable|array',
+            'abilities.*' => [
+                'required',
+                'string',
+                Rule::in([
```
SirSplasch, line 18: The api only has user or interface as currently validated abilites, so I question if allowing a address ability does make sense. As this can currently only exist via customization. And therefore should be adjusted in the customization likewise.

SirSplasch (changes requested):

SirSplasch (approved):

PR #1830, opened 2026-06-22
feat: expose dashboard widgets via the API

patrickweh:
Dashboard widgets can be invoked over the API to return their computed result as JSON, behaving like the Livewire component — just available via API. Mirrors how invokable actions are exposed as routes.

## How it works
- `Contracts\HasApiResponse` — marks a widget as API-exposable (`toApiResponse(): array`).
- `Traits\Livewire\Widget\RespondsToApiRequests` — overrides Livewire's page-component `__invoke()` so the widget class is routable directly. It:
  - validates the request against the widget's `apiRules()` (422 on invalid/missing input),
  - applies the validated values to the matching public properties like the attributes a dashboard passes at mount,
  - runs the real Livewire mount lifecycle (`trigger('mount')`, including trait mount hooks),
  - returns the computed public state as JSON.
- `apiRules()` is overridable per widget to declare/refine its parameters.

## Permissions
Routes live in the standard authenticated group, so the route permission (`api.widgets.{name}.get`) is seeded and enforced by the existing permission middleware — exactly like every other API route and action.

## First opt-in widget
`CurrentWorkTimeModel` at `GET /api/widgets/current-work-time-model`, requiring `employeeId`:

```
GET /api/widgets/current-work-time-model?employeeId=42
→ { "data": { "employeeId": 42, "sum": "8.00h", "subValue": "5 days per week" } }
```

## Tests
- Returns the computed result for the given parameters.
- Validates the request parameters (422).
- Forbidden without the route permission (403).

### DIFF
--- a/src/Livewire/Widgets/Employee/CurrentWorkTimeModel.php
+++ b/src/Livewire/Widgets/Employee/CurrentWorkTimeModel.php
@@ -50,6 +50,13 @@ class CurrentWorkTimeModel extends ValueBox implements HasApiResponse
         $this->subValue = __(':days days per week', ['days' => $workTimeModel->work_days_per_week]);
     }
 
+    protected function apiRules(): array
+    {
+        return [
+            'employeeId' => 'required|integer|exists:employees,id',
+        ];
+    }
+
     protected function icon(): string
     {
         return 'wrench';
--- a/src/Livewire/Widgets/Employee/CurrentWorkTimeModel.php
+++ b/src/Livewire/Widgets/Employee/CurrentWorkTimeModel.php
@@ -2,15 +2,19 @@
 
 namespace FluxErp\Livewire\Widgets\Employee;
 
+use FluxErp\Contracts\HasApiResponse;
 use FluxErp\Livewire\Employee\Dashboard;
 use FluxErp\Livewire\Support\Widgets\ValueBox;
 use FluxErp\Models\Employee;
 use FluxErp\Models\WorkTimeModel;
+use FluxErp\Traits\Livewire\Widget\RespondsToApiRequests;
 use Illuminate\Support\Number;
 use Livewire\Attributes\Locked;
 
-class CurrentWorkTimeModel extends ValueBox
+class CurrentWorkTimeModel extends ValueBox implements HasApiResponse
 {
+    use RespondsToApiRequests;
+
     #[Locked]
     public ?int $employeeId = null;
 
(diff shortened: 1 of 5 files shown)


src/Livewire/Widgets/Employee/CurrentWorkTimeModel.php
```diff
@@ -46,6 +50,13 @@ public function calculateSum(): void
         $this->subValue = __(':days days per week', ['days' => $workTimeModel->work_days_per_week]);
     }
 
+    protected function apiRules(): array
+    {
+        return [
+            'employeeId' => 'required|integer|exists:employees,id',
```
SirSplasch: use ModelExists Rule

src/Traits/Livewire/Widget/RespondsToApiRequests.php
```diff
@@ -0,0 +1,53 @@
...
+
+namespace FluxErp\Traits\Livewire\Widget;
+
+use ReflectionClass;
+use ReflectionProperty;
+use function Livewire\trigger;
+
+trait RespondsToApiRequests
+{
+    protected array $apiResponseExcept = [
```
SirSplasch: I don't know about the optimistic approach, I prefer the pessimistic approach (only expose the given properties).

src/Traits/Livewire/Widget/RespondsToApiRequests.php
```diff
@@ -0,0 +1,53 @@
...
+            $name = $property->getName();
+
+            if (in_array($name, $this->apiResponseExcept, true)) {
+                continue;
+            }
+
+            $data[$name] = $this->{$name};
+        }
+
+        return array_filter($data, fn (mixed $value): bool => ! is_null($value));
```
SirSplasch: Why filter null values? Aren't they valid values?

src/Traits/Livewire/Widget/RespondsToApiRequests.php
```diff
@@ -0,0 +1,53 @@
...
+    public function __invoke()
+    {
+        $name = app('livewire.finder')->normalizeName(static::class);
+        $component = app('livewire')->new($name);
+
+        foreach (request()->validate($component->apiRules()) as $parameter => $value) {
+            $component->{$parameter} = $value;
+        }
+
+        trigger('mount', $component, [], null, null, []);
```
SirSplasch: Be aware that set properties can be overridden in mount.

And can't you call the mount method directly?

SirSplasch (changes requested):

src/Traits/Livewire/Widget/RespondsToApiRequests.php
```diff
@@ -0,0 +1,53 @@
...
+        'options',
+        'widgetId',
+    ];
+
+    public function __invoke()
+    {
+        $name = app('livewire.finder')->normalizeName(static::class);
+        $component = app('livewire')->new($name);
+
+        foreach (request()->validate($component->apiRules()) as $parameter => $value) {
```
SirSplasch: How does this work? The apiRules method has protected visibility, so you shouldn't be able to call this method on $component. $this->apiRules() would work though.

SirSplasch (changes requested):

SirSplasch (approved):

PR #1831, opened 2026-06-22
fix: scope the user broadcast channel to its owner; expose connection config

patrickweh:
## Security fix
The `user.{id}` private broadcast channel was authorized by **existence**, not ownership: the `User` model goes through the generic morph-map channel loop, which registers an existence-scoped channel (`exists()`). As a result **any authenticated client could subscribe to another user's private channel** — meaning user-scoped broadcasts (e.g. work-time changes) leaked to anyone authenticated. The previous explicit ownership channel was shadowed and additionally had a broken parameter binding.

Fix:
- Skip the `User` model in the generic existence-channel loop.
- Register a dedicated **ownership-scoped** `user.{id}` channel (`$user->getKey() === $id`), authorized for the `web`, `address`, `sanctum` and `token` guards.

This closes the leak for both the web app and API/token clients, and lets token-authenticated clients authorize their own user channel.

## Connection config
Add `GET /api/broadcasting/connection` returning only the **public** connection values (`key`, `host`, `port`, `scheme`) from the default broadcast connection, so clients can discover where to connect. The app secret never leaves the server.

## Tests
- The user channel authorizes only its owner (own → true, other → false).
- The user model no longer has a generic existence channel (regression guard).
- The connection endpoint exposes only the public values (no secret).

## Summary by Sourcery

Secure user-specific broadcasting channels and expose public broadcasting connection details via API.

Bug Fixes:
- Restrict the user private broadcast channel so only the owning user can subscribe, and remove the generic existence-scoped user channel.

Enhancements:
- Expose a new /api/broadcasting/connection endpoint that returns only public broadcasting connection parameters for clients to discover connection settings.

Tests:
- Add feature tests to verify user-channel ownership authorization, absence of a generic user existence channel, and that the connection endpoint omits secret values.

### DIFF
--- a/routes/api.php
+++ b/routes/api.php
@@ -417,6 +417,7 @@ Route::prefix('api')
     ->group(function (): void {
         Route::get('/health', [MobileController::class, 'health']);
         Route::get('/mobile/config', [MobileController::class, 'config']);
+        Route::get('/broadcasting/connection', [MobileController::class, 'broadcastingConnection']);
         Route::delete('/mobile/device-token/{deviceId}', [MobileController::class, 'deleteDeviceToken']);
 
         Route::post('/auth/token', [AuthController::class, 'authenticate']);
--- a/routes/api.php
+++ b/routes/api.php
@@ -418,7 +418,6 @@ Route::prefix('api')
         Route::get('/broadcasting/connection', [MobileController::class, 'broadcastingConnection']);
         Route::get('/health', [MobileController::class, 'health']);
         Route::get('/mobile/config', [MobileController::class, 'config']);
-        Route::delete('/mobile/device-token/{deviceId}', [MobileController::class, 'deleteDeviceToken']);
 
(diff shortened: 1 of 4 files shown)


routes/api.php
```diff
@@ -417,6 +417,7 @@
     ->group(function (): void {
         Route::get('/health', [MobileController::class, 'health']);
         Route::get('/mobile/config', [MobileController::class, 'config']);
+        Route::get('/broadcasting/connection', [MobileController::class, 'broadcastingConnection']);
         Route::delete('/mobile/device-token/{deviceId}', [MobileController::class, 'deleteDeviceToken']);
```
SirSplasch: You sure about exposing these routes without middleware? Like these are open for everyone.

SirSplasch: Also order them, these seem to have random ordering.

SirSplasch (changes requested):

SirSplasch (approved):

PR #1837, opened 2026-06-23
feat: user login-url endpoint and weekly target on the work time widget

patrickweh:
Two additions for the Nuxbe Desktop client.

## Magic login URL
`POST /api/user/login-url` (gated to the `user` ability) returns a short-lived, single-use, signed magic login URL for the authenticated user. A client can open the web UI at that URL to land already authenticated, instead of being prompted to log in again. `User::generateLoginLink()` is made public so the endpoint reuses the existing mechanism (signed temporary route + cached single-use token, 15 min).

The URL is a login credential — it is short-lived, single-use (`Cache::pull`), HTTPS-only, and the desktop client excludes it from logging.

## Weekly target
`CurrentWorkTimeModel` now also exposes `weeklyTarget` (the work time model's weekly hours), so the desktop can show a weekly worked/target card.

## Tests
- The login-url endpoint returns a signed magic login URL.
- It requires the `user` ability (403 otherwise).
- The widget returns `weeklyTarget`.

## Summary by Sourcery

Add an authenticated magic login URL endpoint and expose weekly work-time targets in the employee work time widget.

New Features:
- Expose a POST /api/user/login-url endpoint that returns a short-lived, signed magic login URL for the authenticated user, gated by the user ability.
- Expose weeklyTarget on the CurrentWorkTimeModel widget API to surface weekly work-time targets to clients.

Tests:
- Add feature tests covering the login-url endpoint behavior and authorization requirements.
- Extend widget API tests to assert the presence and value of the weeklyTarget field.

### DIFF
--- a/src/Http/Controllers/AuthController.php
+++ b/src/Http/Controllers/AuthController.php
@@ -75,6 +75,14 @@ class AuthController extends Controller
         ])->onlyInput('email');
     }
 
+    public function loginUrl(Request $request): JsonResponse
+    {
+        return ResponseHelper::createResponseFromBase(
+            statusCode: 200,
+            data: ['url' => $request->user()->generateLoginLink()],
+        );
+    }
+
     public function logout(Request $request): JsonResponse
     {
         $request->user()->tokens()->delete();
@@ -88,13 +96,5 @@ class AuthController extends Controller
     public function validateToken(): JsonResponse
     {
         return response()->json(['status' => 'token valid']);
-    }
-
-    public function loginUrl(Request $request): JsonResponse
-    {
-        return ResponseHelper::createResponseFromBase(
-            statusCode: 200,
-            data: ['url' => $request->user()->generateLoginLink()],
-        );
     }
 }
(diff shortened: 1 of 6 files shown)


src/Http/Controllers/AuthController.php
```diff
@@ -89,4 +89,12 @@ public function validateToken(): JsonResponse
     {
         return response()->json(['status' => 'token valid']);
     }
+
+    public function loginUrl(Request $request): JsonResponse
```
SirSplasch: order alphabetical

src/Models/User.php
```diff
@@ -278,16 +278,7 @@ public function sendLoginLink(): void
...
-    protected function password(): Attribute
-    {
-        return Attribute::set(
-            fn ($value) => Hash::info($value)['algoName'] !== 'bcrypt' ? Hash::make($value) : $value,
-        );
-    }
-
-    // Protected methods
-    protected function generateLoginLink(): string
+    public function generateLoginLink(): string
```
SirSplasch: order alphabetically in the public methods section.

SirSplasch (changes requested):

SirSplasch (approved):

PR #1840, opened 2026-06-23
Reload comments when a renderless parent swaps the bound record

patrickweh:
## Problem

The comments tab goes empty (or shows the previous record's comments) after opening one record, opening a different one, then returning to the first — when the embedding component opens its modal **renderlessly**.

Reproduced end-to-end on a live instance (recruitment list → open contact A → open contact B → back to A → comments tab empty). Confirmed via DOM inspection that the comments component's `modelId` was stale/`null` while the parent's bound value was correct.

## Root cause

`Support\Comments` receives the record id via `wire:model` (Modelable). Modelable only propagates to the child on a **parent render**. Embedders open their comment modal with a renderless action (e.g. `TransactionAssignments::showComments`, to avoid re-rendering large lists), so the bound `modelId` never updates and the Alpine `$watch('modelId')` never fires. `loadComments()` then runs with a stale/null `modelId`.

## Fix

- `comments.js` resolves the **live** record id from the parent through the Modelable binding (`wire:model="$parent.<expr>"` → `$wire.$parent.get(expr)`) and passes it explicitly to `loadComments`/`loadStickyComments`.
- Reload when the section becomes visible again (tab switch / modal reopen) via an `IntersectionObserver`, if the resolved id differs from the last loaded one.
- `loadComments(?int $modelId)` / `loadStickyComments(?int $modelId)` accept the id and set `$this->modelId`, keeping the renderless contract.

Falls back to the previous behaviour when the binding isn't a `$parent.*` expression.

## Tests

`tests/Livewire/Support/CommentsModelIdTest.php` covers the server contract: `loadComments`/`loadStickyComments` honour the passed id and switch records correctly. The browser-level behaviour was verified manually against the live reproduction.

## Summary by Sourcery

Ensure the comments component reloads the correct record when a renderless parent swaps the bound model, preventing stale or empty comment lists.

Bug Fixes:
- Fix comments not updating or appearing empty when navigating between records opened via renderless parent modals by resolving and passing the current model id explicitly.

Enhancements:
- Resolve the live model id from the parent binding in the comments frontend and track the last loaded id to decide when to reload.
- Observe the comments section visibility to trigger reloads when it becomes visible again after a tab or modal change.
- Allow server-side comment-loading methods to accept an optional model id and update internal state accordingly.

Tests:
- Add Livewire tests validating that loadComments and loadStickyComments honor a passed model id, switch records correctly, and return an empty result when no model id is set.

### DIFF
--- a/resources/js/components/comments.js
+++ b/resources/js/components/comments.js
@@ -4,24 +4,17 @@ export default function comments() {
         stickyComments: [],
         initialized: false,
 
-        loadedModelId: null,
-
         async init() {
             this.loadComments();
 
             this.$wire.$watch('modelId', () => this.loadComments());
 
-            // The modelId is bound from the parent via wire:model (Modelable). When the parent
-            // swaps the bound record while skipping its own render (e.g. a renderless modal
-            // opener), the bound value stays stale and $watch never fires. Reload whenever this
-            // section becomes visible again (tab switch / modal reopen) if the parent's current
-            // value differs from what we last loaded.
             this.commentsObserver = new IntersectionObserver((entries) => {
                 if (
                     entries.some((entry) => entry.isIntersecting) &&
-                    this.$resolveModelId() !== this.loadedModelId
+                    this.$resolveModelId() !== this.$wire.modelId
                 ) {
-                    this.loadComments();
+                    this.$wire.modelId = this.$resolveModelId();
                 }
             });
             this.commentsObserver.observe(this.$root);
@@ -39,12 +32,9 @@ export default function comments() {
         },
 
         loadComments() {
-            const modelId = this.$resolveModelId();
-            this.loadedModelId = modelId;
-
             Promise.all([
-                this.$wire.loadComments(modelId),
-                this.$wire.loadStickyComments(modelId),
+                this.$wire.loadComments(),
+                this.$wire.loadStickyComments(),
             ]).then(([commentsResponse, stickyCommentsResponse]) => {
                 this.comments = commentsResponse.data;
                 this.stickyComments = stickyCommentsResponse;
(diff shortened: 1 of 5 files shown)


src/Livewire/Support/Comments.php
SirSplasch, line 1: I don't know about passing a modelId into these methods, as these can not exists, or the user should not be able to access the comments of the passed modelId. These are public methods which can be called by ill will. if I am not mistaken.

SirSplasch (changes requested):

src/Livewire/Support/Comments.php
patrickweh: modelId is a #[Modelable] public property (not #[Locked]), so the client can already set it via the Livewire request regardless of a method parameter — the param wasn't a new vector. I removed it anyway: the frontend now re-syncs modelId from the parent and the public methods keep their original signature.

The actual authorization is the scoped record lookup: loadComments already returns [] when resolve_static(modelType,'query')->whereKey(modelId)->first() yields nothing (global ownership/tenant scopes). loadStickyComments was missing that check — added it (->exists() on the scoped query), so an arbitrary modelId only ever returns comments for records the user can actually access.

resources/js/components/comments.js
```diff
@@ -17,10 +32,36 @@ export default function comments() {
...
+            this.commentsObserver?.disconnect();
+        },
+
+        /**
+         * Read the live record id from the parent through the Modelable binding instead of
+         * trusting the possibly-stale local modelId.
+         */
+        resolveModelId() {
+            const root = this.$root.closest('[wire\\:id]');
+            const binding = root?.getAttribute('wire:model');
```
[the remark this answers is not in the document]
patrickweh: Resolved: resolveModelId now scans getAttributeNames() for any attribute starting with 'wire:model', so bindings with modifiers (wire:model.live, .defer, ...) are matched instead of only the bare wire:model.

SirSplasch (approved):

PR #1842, opened 2026-06-24
feat: add targeted loading feedback to action buttons

patrickweh:
Adds the TallStackUI `loading` attribute to action buttons (save, delete, send, create, assign, sync, process, download) that previously gave no visual feedback during the server round-trip. On click the button now shows a spinner and disables, so the UI feels responsive instead of frozen.

Each `loading` target mirrors the button's `wire:click` call including its parameters (e.g. `loading="delete({{ $id }})"`), so only the clicked button shows the spinner rather than every button on the component. A few pre-existing untargeted `loading` / `wire:loading.attr` attributes that reacted to any component request were replaced with the targeted form.

Scope: only genuine server actions. Buttons whose `wire:click` runs pure JS, opens a modal, or navigates were left untouched, as were per-row buttons whose target depends on an Alpine runtime variable (loading can't be matched per-instance there).

## Summary by Sourcery

Improve user feedback on long-running Livewire actions by adding targeted loading states to action buttons across the app.

Enhancements:
- Add TallStackUI loading targets to save, delete, assign, send, download, and similar Livewire action buttons so they show a spinner and disable during server requests.
- Replace generic component-wide loading attributes with action-specific loading bindings to ensure only the triggered button reflects the in-progress state.

### DIFF
--- a/resources/views/livewire/accounting/transaction-assignments.blade.php
+++ b/resources/views/livewire/accounting/transaction-assignments.blade.php
@@ -48,8 +48,8 @@
             @stack('transaction-assign-orders-modal-footer')
             <x-button
                 :text="__('Assign')"
-                wire:click="assignOrders"
-                loading="assignOrders"
+                wire:click="assignOrders()"
+                loading="assignOrders()"
             />
         </x-slot:footer>
     </x-modal>
--- a/resources/views/livewire/accounting/transaction-assignments.blade.php
+++ b/resources/views/livewire/accounting/transaction-assignments.blade.php
@@ -46,7 +46,11 @@
                 "
             />
             @stack('transaction-assign-orders-modal-footer')
-            <x-button :text="__('Assign')" wire:click="assignOrders" />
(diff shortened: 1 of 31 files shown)


resources/views/livewire/accounting/transaction-assignments.blade.php
```diff
@@ -46,7 +46,11 @@ class="mt-2 flex w-full flex-row justify-between border-t border-slate-200 pt-2"
                 "
             />
             @stack('transaction-assign-orders-modal-footer')
-            <x-button :text="__('Assign')" wire:click="assignOrders" />
+            <x-button
+                :text="__('Assign')"
+                wire:click="assignOrders"
+                loading="assignOrders"
```
SirSplasch: add () to the method calls to be consistent

SirSplasch (changes requested):

SirSplasch (approved):

PR #1843, opened 2026-06-24
feat: optional redirect target for the magic login url

Slupi:
Builds on #1837.

- `generateLoginLink(?string $intendedUrl = null)` — uses the given URL as the post-login destination, falling back to the previous behaviour (session intended → dashboard) when none is passed.
- `POST /api/user/login-url` accepts an optional `redirect`, sanitised to the app's own host (relative paths or same-host absolute URLs only) to avoid an open redirect.

Lets a client (e.g. the desktop/extension) deep-link straight to a record after logging in. Tests cover a same-host redirect being honoured and an external one falling back to the dashboard.

## Summary by Sourcery

Add support for specifying a safe, optional redirect target when generating magic login URLs and ensure external redirects are rejected.

New Features:
- Allow clients to pass an optional redirect target to the magic login URL endpoint, which is used as the post-login destination when valid.

Enhancements:
- Sanitise and constrain login redirect targets to same-host URLs to prevent open redirect vulnerabilities.

Tests:
- Add feature tests verifying that same-host redirect targets are honoured and external redirect targets fall back to the dashboard.

### DIFF
--- a/src/Http/Controllers/AuthController.php
+++ b/src/Http/Controllers/AuthController.php
@@ -2,18 +2,15 @@
 
 namespace FluxErp\Http\Controllers;
 
-use Closure;
 use FluxErp\Helpers\ResponseHelper;
+use FluxErp\Http\Requests\LoginUrlRequest;
 use FluxErp\Models\User;
 use Illuminate\Http\JsonResponse;
 use Illuminate\Http\RedirectResponse;
 use Illuminate\Http\Request;
-use Illuminate\Routing\Exceptions\UrlGenerationException;
 use Illuminate\Support\Facades\Auth;
 use Illuminate\Support\Facades\Hash;
-use Illuminate\Support\Facades\Route;
 use Illuminate\Support\Facades\Validator;
-use Illuminate\Validation\ValidationException;
 
 class AuthController extends Controller
 {
@@ -79,27 +76,13 @@ class AuthController extends Controller
         ])->onlyInput('email');
     }
 
-    public function loginUrl(Request $request): JsonResponse
+    public function loginUrl(LoginUrlRequest $request): JsonResponse
     {
-        $validated = $request->validate([
-            'redirect' => ['nullable', 'string', function (string $attribute, mixed $value, Closure $fail): void {
-                if (! blank($value) && ! Route::has($value)) {
-                    $fail(__('The :attribute must be a valid route name.', ['attribute' => $attribute]));
-                }
-            }],
-            'redirect_params' => ['nullable', 'array'],
-        ]);
+        $validated = $request->validated();
 
-        $intended = null;
-        if (! blank($validated['redirect'] ?? null)) {
-            try {
-                $intended = route($validated['redirect'], $validated['redirect_params'] ?? []);
-            } catch (UrlGenerationException) {
-                throw ValidationException::withMessages([
-                    'redirect' => __('The :attribute route parameters are invalid.', ['attribute' => 'redirect']),
-                ]);
-            }
-        }
+        $intended = blank($validated['redirect'] ?? null)
+            ? null
+            : route($validated['redirect'], $validated['redirect_params'] ?? []);
 
         return ResponseHelper::createResponseFromBase(
             statusCode: 200,
--- a/src/Http/Controllers/AuthController.php
+++ b/src/Http/Controllers/AuthController.php
@@ -2,14 +2,18 @@
 
 namespace FluxErp\Http\Controllers;
 
+use Closure;
 use FluxErp\Helpers\ResponseHelper;
 use FluxErp\Models\User;
 use Illuminate\Http\JsonResponse;
 use Illuminate\Http\RedirectResponse;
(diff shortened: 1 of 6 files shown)


src/Http/Controllers/AuthController.php
```diff
@@ -97,4 +103,23 @@ public function validateToken(): JsonResponse
...
+
+        if (! str_contains($redirect, '://') && ! str_starts_with($redirect, '//') && ! str_starts_with($redirect, '/')) {
+            $redirect = '/' . $redirect;
+        }
+
+        if (str_starts_with($redirect, '/') && ! str_starts_with($redirect, '//')) {
+            return url($redirect);
+        }
+
+        return parse_url($redirect, PHP_URL_HOST) === parse_url(config('app.url'), PHP_URL_HOST)
```
SirSplasch: `The [Uri\Rfc3986\Uri](https://www.php.net/manual/en/class.uri-rfc3986-uri.php) and [Uri\WhatWg\Url](https://www.php.net/manual/en/class.uri-whatwg-url.php) classes strictly follow the RFC 3986 and WHATWG URL Standards respectively. It is strongly recommended to use these classes for all newly written code and to migrate existing uses of the parse_url() function to these classes, unless the parse_url() behavior needs to be preserved for compatibility reasons.`

src/Http/Controllers/AuthController.php
```diff
@@ -77,9 +77,15 @@ public function authenticateWeb(Request $request): RedirectResponse
 
     public function loginUrl(Request $request): JsonResponse
     {
+        $validated = $request->validate([
+            'redirect' => ['nullable', 'string', 'max:2048'],
```
SirSplasch: Does it make sense to validate the redirect to be a valid url?
https://laravel.com/docs/13.x/validation#rule-url

src/Http/Controllers/AuthController.php
```diff
@@ -97,4 +103,23 @@ public function validateToken(): JsonResponse
     {
         return response()->json(['status' => 'token valid']);
     }
+
+    protected function resolveSafeRedirect(?string $redirect): ?string
```
SirSplasch: Isn't this kinda part of the validation and therefore should return a validation exception if redirect is not valid?

SirSplasch: I am not sure where the redirect should be possible, but in my mind it should be an existing route of the application (or null). Therefore can't you call `route($redirect)` as a validation method or does this bring other negative/unwanted side-effects?

SirSplasch (changes requested):

src/Http/Controllers/AuthController.php
SirSplasch: Resolved this, as the redirect string should probably the name of a named route.

Slupi:
Reworked the redirect to use **named routes** as suggested (bcbbda43c):

- `redirect` is now a route **name**, with optional `redirect_params`; resolved via `route($name, $params)`.
- Only existing application routes are reachable, so there is no open-redirect surface and `parse_url`/`resolveSafeRedirect` are gone.
- An unknown route name **or** missing required route parameters now fail validation with a `422` (`redirect` validation error).
- Absent/null `redirect` falls back to the dashboard as before.

Tests cover: named route honoured, named route with params, unknown route → 422, missing params → 422.

Slupi:
Follow-up (d961cbb3e): moved the redirect validation into the standard layers to match the codebase style —

- new `LoginUrlRequest` (extends `BaseFormRequest`, like `GetPrintViewsRequest`) so the 422 goes through the shared `ResponseHelper` `failedValidation`;
- new `FluxErp\Rules\RouteExists` (`DataAwareRule`/`ValidationRule`, modelled on `MorphExists`/`ViewExists`) that checks the route exists **and** resolves with `redirect_params`;
- controller is now just `loginUrl(LoginUrlRequest $request)` → `route(...)` → `ResponseHelper`. No inline closure / try-catch in the controller.

src/Rules/RouteExists.php
```diff
@@ -0,0 +1,45 @@
...
+use Illuminate\Contracts\Validation\DataAwareRule;
+use Illuminate\Contracts\Validation\ValidationRule;
+use Illuminate\Routing\Exceptions\UrlGenerationException;
+use Illuminate\Support\Facades\Route;
+
+class RouteExists implements DataAwareRule, ValidationRule
+{
+    protected array $data = [];
+
+    private ?string $parameterAttribute;
```
SirSplasch: protected to allow customization

src/Rules/RouteExists.php
SirSplasch, line 1: add localization keys to de.json

SirSplasch (changes requested):

src/Rules/RouteExists.php
Slupi: done

src/Rules/RouteExists.php
Slupi: done

SirSplasch (approved):

PR #1844, opened 2026-06-24
fix: stream a PDF response from /api/print/pdf

Slupi:
The `/api/print/pdf` route pointed directly at the `Printing` action. Its `execute()` returns a `PrintableView` (intentional — callers also use it to save media), so at the HTTP boundary the action's response return type was violated and the endpoint 500'd (`TypeError … FluxErp\View\Printing\Order\Invoice returned`).

- Route `/api/print/pdf` through `PrintController::pdf`, which runs the action and returns the view's `streamPDF()` `Response`.
- `Printing` action and `/api/print/render` are unchanged.

Test asserts the endpoint returns a `200` `application/pdf` body starting with `%PDF-`.

## Summary by Sourcery

Route the /api/print/pdf endpoint through a controller method that streams the PDF response returned by the Printing action.

Bug Fixes:
- Ensure /api/print/pdf returns a streamed PDF HTTP response instead of a PrintableView object that caused a 500 error.

Tests:
- Add a unit test verifying that the print PDF endpoint returns a 200 response with an application/pdf body starting with a PDF header.

### DIFF
--- a/src/View/Printing/PrintableView.php
+++ b/src/View/Printing/PrintableView.php
@@ -12,6 +12,7 @@ use FluxErp\Contracts\SignablePrintView;
 use FluxErp\Models\Tenant;
 use FluxErp\Printing\Printable;
 use FluxErp\Traits\Makeable;
+use Illuminate\Contracts\Support\Responsable;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Http\Response;
 use Illuminate\Support\Facades\File;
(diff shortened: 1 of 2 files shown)


Slupi:
Addressed by reworking the fix: instead of a controller, **\`PrintableView\` now implements \`Responsable\`** (\`toResponse()\` → \`streamPDF()\`). \`/api/print/pdf\` stays routed at the \`Printing\` action exactly like \`/api/print/render\`, so:

- validation stays shared (the action’s \`PrintingRuleset\`) — no separate \`FormRequest\` or generic \`Request\`;
- \`html\`/\`preview\` derivation stays in the route defaults for both endpoints — no duplicated option-building in a controller.

The action and its \`execute()\` PrintableView contract are unchanged; the only addition is the \`Responsable\` interface so the view can be returned as an HTTP response.

SirSplasch (approved):

PR #1845, opened 2026-06-24
Rework payment reminder run

patrickweh:
## Summary

Reworks the payment reminder run (dunning) page, which was non-functional, and removes the legacy reminder list.

### Eligibility / correctness
- The run wrongly compared the order-type multiplier (a string) with an int, which filtered out every invoice and left the page empty. Fixed with a value-correct `bccomp` check.
- Orders in payment state `paid`, `in_payment` or `in_open_payment_run` are now excluded from the run, even when their balance is non-zero.
- Invoices whose next reminder level has no configured reminder text are kept in the list but skipped on send (no mail goes out).

### Sending
- Sends one reminder mail per invoice (correct, invoice-specific subject) instead of one bundled mail per group.
- Sending runs asynchronously inside a single monitored batch (even for one invoice), giving the same progress toast as bulk mailing and avoiding request timeouts.

### UI
- Filters (tenant, reminder level, search, minimum days overdue) and sorting.
- Days-overdue badges, per-invoice selection, an editable recipient per group and a PDF preview of the reminder.
- Full German localization and user-locale date formatting.

### Cleanup
- Removes the legacy `PaymentReminder` list component, its route, menu entry and test; repoints the "Outstanding" widget's overdue action to the order list.

## Summary by Sourcery

Rework the payment reminder run to correctly identify and process eligible invoices, send reminders asynchronously per invoice, and modernize the UI while removing the legacy reminder list.

New Features:
- Introduce per-invoice selection, filtering, sorting, and PDF preview capabilities for the payment reminder run UI.
- Allow overriding recipient email addresses per reminder group and default-select all eligible invoices for sending.
- Queue individual payment reminder sends as a monitored batch using a dedicated job class.

Bug Fixes:
- Fix payment reminder eligibility by correctly handling order-type multipliers and excluding orders in paid or payment-run states.
- Ensure invoices without configured reminder texts are not sent while remaining in the reminder run list.

Enhancements:
- Refine the payment reminder run grouping and locale-aware display, including days-overdue badges and localized dates.
- Simplify reminder sending flow to reuse the bulk-mail progress feedback without additional toasts.
- Adjust the Outstanding widget overdue action to point to the order list instead of the removed reminder list.

Tests:
- Add coverage for payment reminder run listing, selection, and filtering behavior.
- Add tests for the payment reminder eligibility scope, reminder send batching, and the new send job’s failure handling.
- Update widget redirection tests to reflect the removal of the legacy payment reminder list.

Chores:
- Remove the legacy PaymentReminder Livewire component, its route, menu entry, and associated tests.


lang/de.json
```diff
@@ -1638,12 +1648,14 @@
...
     "Min Notice Days": "Minimale Vorlauftage",
     "Min Purchase": "Mindestbestellmenge",
+    "Min. days overdue": "Min. Tage überfällig",
     "Minimum Contract Duration": "Mindestvertragslaufzeit",
     "Minimum reminder level": "Mindest-Mahnstufe",
     "Minute": "Minute",
     "minute": "Minute",
     "Minutes": "Minuten",
     "minutes": "Minuten",
+    "Missing payment reminder template": "Kein Mahnungs-Mailtemplate hinterlegt",
```
SirSplasch: "Mailvorlage" instead of "Mailtemplate"?

resources/views/livewire/accounting/payment-reminder-run.blade.php
```diff
@@ -2,79 +2,238 @@
...
-                x-bind:disabled="!($wire.selected || []).length"
+            <x-select.styled
+                wire:model.live="sort"
+                required
+                :options="[
+                    ['label' => __('Days overdue (desc)'), 'value' => 'overdue_days_desc'],
+                    ['label' => __('Days overdue (asc)'), 'value' => 'overdue_days_asc'],
+                    ['label' => __('Balance (desc)'), 'value' => 'balance_desc'],
+                    ['label' => __('Balance (asc)'), 'value' => 'balance_asc'],
+                    ['label' => __('Customer'), 'value' => 'contact_asc'],
```
SirSplasch: Why no contact_desc value? to be consistent to overdue_days and balance.

resources/views/livewire/accounting/payment-reminder-run.blade.php
```diff
@@ -2,79 +2,238 @@
...
 
         @if ($this->isEmpty)
             <div class="text-secondary-500 py-12 text-center text-sm">
-                {{ __('No payment reminders are due today.') }}
+                {{ __('No payment reminders are due.') }}
             </div>
         @else
-            <div class="flex flex-col gap-3">
+            @php
+                $totalBalance = collect($groups)->sum(fn ($g) => (float) $g['total_balance']);
```
SirSplasch: not a fan of single character variable names except for i,( j, k) in a/multiple for loop(s).

src/Actions/PaymentReminder/BundlePaymentReminders.php
```diff
@@ -26,162 +25,32 @@ public function performAction(): array
...
-        ];
+        // One mail per invoice. All sends run as a single monitored batch (even a
+        // batch of one) so the user gets the same progress toast as bulk mailing.
+        $jobs = $orders
+            ->map(function (Order $order) use ($recipients): SendPaymentReminderJob {
+                $key = $order->contact_id . '-' . ((int) $order->payment_reminder_current_level + 1);
 
-        foreach ($groups as $group) {
-            $sent = $this->sendGroup($group);
+                return new SendPaymentReminderJob($order->getKey(), data_get($recipients, $key));
```
SirSplasch: Please allow customization, "new" does not allow that.

SirSplasch: I'm sorry, but you cannot expect that the recipients array is keyed with the calculated key(s). The validation only validates the values, and one would not expect to pass an array with perfect matching keys. To me this behavior is a little to far fetched, as one needs to look into this specific implementation to know what's going on and expected.

src/Livewire/Accounting/PaymentReminderRun.php
```diff
@@ -80,73 +161,106 @@ public function isEmpty(): bool
...
+        $this->sendBundle($this->selectedOrders);
         $this->loadData();
     }
 
+    protected function sortGroups(Collection $groups): Collection
+    {
+        return match ($this->sort) {
+            'overdue_days_asc' => $groups->sortBy('max_overdue_days')->values(),
+            'balance_desc' => $groups->sortByDesc(fn (array $g) => (float) $g['total_balance'])->values(),
+            'balance_asc' => $groups->sortBy(fn (array $g) => (float) $g['total_balance'])->values(),
```
SirSplasch: Avoid single character variables.

src/Livewire/Accounting/PaymentReminderRun.php
```diff
@@ -20,7 +25,26 @@ class PaymentReminderRun extends Component
...
     public array $groups = [];
 
-    public array $selected = [];
+    /** Selected order ids (selection is per invoice). */
+    public array $selectedOrders = [];
+
+    /** Editable recipient email per group, keyed by group key. */
+    public array $recipientEmails = [];
+
+    public ?int $filterTenantId = null;
```
SirSplasch: As far as I can see, this is not used in the blade component, and therefore is ever null.

src/Livewire/Accounting/PaymentReminderRun.php
SirSplasch, line 1: reoder public methods alphabetical, except fixed ones like mount/render

src/Livewire/Accounting/PaymentReminderRun.php
```diff
@@ -80,73 +161,106 @@ public function isEmpty(): bool
...
+        if (! $this->selectedOrders) {
+            return;
+        }
 
-        $this->sendBundle($orderIds);
+        $this->sendBundle($this->selectedOrders);
         $this->loadData();
     }
 
+    protected function sortGroups(Collection $groups): Collection
```
SirSplasch: After sendBundle to keep alphabetical ordering of protected methods.

src/Livewire/Accounting/PaymentReminderRun.php
```diff
@@ -80,73 +161,106 @@ public function isEmpty(): bool
...
+        $reminder->reminder_level = (int) $order->payment_reminder_current_level + 1;
+        $reminder->setRelation('order', $order);
+
+        try {
+            $view = PaymentReminderView::make($reminder)
+                ->preview()
+                ->print();
+
+            $this->previewSrc = 'data:application/pdf;base64,' . base64_encode($view->pdf->output());
+            $this->showPreview = true;
```
SirSplasch, line 177: Just curious why you decided to use a boolean rather than js to open/close the modal.

SirSplasch (changes requested):

src/Rulesets/PaymentReminder/BundlePaymentRemindersRuleset.php
SirSplasch, line 1: You did not get the point of what I was pointing to. You are not validating the key of the recipients array, you just changed the key from a concatinated string to the order key. So I will have to give you a solution for this issue.

```
            'orders' => 'required|array|min:1',
            'orders.*.id' => [
                'required',
                'integer',
                app(ModelExists::class, ['model' => Order::class])
                    ->wherePaymentReminderEligible(),
            ],
            'orders.*.recipient' => 'nullable|email',
```

SirSplasch (changes requested):

src/Rulesets/PaymentReminder/BundlePaymentRemindersRuleset.php
patrickweh: Done. Switched to a single `orders` array of `{id, recipient}` and validate `orders.*.id` + `orders.*.recipient` as you suggested, so the recipient is validated together with its order instead of via a separate keyed map. Action and Livewire component build/consume the same shape now.

src/Livewire/Accounting/PaymentReminderRun.php
```diff
@@ -123,30 +227,46 @@ protected function sendBundle(array $orderIds): void
...
+                $emailByOrderId[$order['id']] = filled($email) ? $email : null;
+            }
+        }
+
+        $orders = collect($orderIds)
+            ->map(fn ($orderId): array => [
+                'id' => (int) $orderId,
+                'recipient' => data_get($emailByOrderId, (int) $orderId),
+            ])
+            ->all();
```
SirSplasch: Isn't there a better way of pairing the ids with the recipients?
Currently you are effectively looping twice, first through all groups and then through all orderIds.

Like you can add entries to $orders while looping through the groups.
```
        if (in_array($order['id'], $orderIds)) {
            $orders[] = [
                'id' => $order['id'],
                'recipient' => filled($email) ? $email : null,
            ];
        }
```

SirSplasch (changes requested):

SirSplasch:
Also resolve merge conflicts.

SirSplasch (changes requested):
fix failed tests

SirSplasch (approved):

PR #1847, opened 2026-06-25
Add token-authed search endpoint

Slupi:
Expose GET /api/search/{model} under auth:sanctum, reusing SearchController so API clients get the same formatted {id,label,description,image} results as the web search bar over a Bearer token.

## Summary by Sourcery

Introduce a token-authenticated API search endpoint that reuses the existing search controller behavior for model-based queries.

New Features:
- Add GET /api/search/{model} endpoint under Sanctum auth, delegating to SearchController for formatted search results.

Tests:
- Add feature tests covering successful formatted search responses, permission-based access denial, and 404 handling for unknown models on the new API search endpoint.

### DIFF
--- a/routes/api.php
+++ b/routes/api.php
@@ -1044,6 +1044,10 @@ Route::prefix('api')
                 Route::put('/schedules', UpdateSchedule::class);
                 Route::delete('/schedules/{id}', DeleteSchedule::class);
 
+                // Search
+                Route::get('/search/{model}', SearchController::class)
+                    ->where('model', '[A-Za-z0-9._-]+');
+
                 // SepaMandates
                 Route::get('/sepa-mandates/{id}', [BaseController::class, 'show'])
                     ->defaults('model', SepaMandate::class);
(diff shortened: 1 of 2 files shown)


Slupi:
Narrowed the constraint to [A-Za-z0-9._-]+ in 3b7535638 — it no longer greedily matches multi-segment paths or resolves arbitrary namespaced class paths, and still covers all morph aliases.

routes/api.php
```diff
@@ -1177,6 +1178,9 @@
                 Route::post('/user/login-url', [AuthController::class, 'loginUrl'])
                     ->middleware('ability:user');
 
+                Route::get('/search/{model}', SearchController::class)
```
SirSplasch: Order this route between schedules and sepa mandates and add // Search comment above it.

SirSplasch (changes requested):

routes/api.php
Slupi: done

SirSplasch (approved):

PR #1848, opened 2026-06-25
Add token-authed user notifications endpoint

Slupi:
- `GET /api/user/notifications` — recent notifications (newest 30) + `unread_count`
- `POST /api/user/notifications/read` — mark one (`{id}`) or all (`{all:true}`) read, user-scoped
- For the Nuxbe Companion extension; 3 feature tests

## Summary by Sourcery

Add token-authenticated user notifications API endpoints for listing recent notifications and marking them as read.

New Features:
- Expose GET /api/user/notifications to return the current user’s latest notifications along with an unread count.
- Expose POST /api/user/notifications/read to allow the current user to mark one or all notifications as read and receive the updated unread count.

Tests:
- Add feature tests covering listing notifications with unread count and marking single or all notifications as read.

### DIFF
--- a/routes/api.php
+++ b/routes/api.php
@@ -304,6 +304,7 @@ use FluxErp\Http\Controllers\BaseController;
 use FluxErp\Http\Controllers\CommentController;
 use FluxErp\Http\Controllers\EventSubscriptionController;
 use FluxErp\Http\Controllers\MobileController;
+use FluxErp\Http\Controllers\NotificationController;
 use FluxErp\Http\Controllers\PermissionController;
 use FluxErp\Http\Controllers\PrintController;
 use FluxErp\Http\Controllers\RoleController;
@@ -1174,6 +1175,8 @@ Route::prefix('api')
 
                     return ResponseHelper::createResponseFromBase(statusCode: 200, data: $user);
                 });
+                Route::get('/user/notifications', [NotificationController::class, 'userIndex']);
+                Route::post('/user/notifications/read', [NotificationController::class, 'markRead']);
                 Route::post('/user/login-url', [AuthController::class, 'loginUrl'])
                     ->middleware('ability:user');
 
--- a/routes/api.php
+++ b/routes/api.php
@@ -1175,10 +1175,10 @@ Route::prefix('api')
 
                     return ResponseHelper::createResponseFromBase(statusCode: 200, data: $user);
                 });
-                Route::get('/user/notifications', [NotificationController::class, 'userIndex']);
-                Route::post('/user/notifications/read', [NotificationController::class, 'markRead']);
                 Route::post('/user/login-url', [AuthController::class, 'loginUrl'])
                     ->middleware('ability:user');
+                Route::get('/user/notifications', [NotificationController::class, 'userIndex']);
+                Route::post('/user/notifications/read', [NotificationController::class, 'markRead']);
 
                 Route::get('/users/{id}', [BaseController::class, 'show'])->defaults('model', User::class);
                 Route::get('/users', [BaseController::class, 'index'])->defaults('model', User::class);
--- a/src/Http/Controllers/NotificationController.php
+++ b/src/Http/Controllers/NotificationController.php
@@ -1,0 +1,53 @@
+<?php
+
+namespace FluxErp\Http\Controllers;
+
+use FluxErp\Helpers\ResponseHelper;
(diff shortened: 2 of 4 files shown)


routes/api.php
```diff
@@ -1174,6 +1175,8 @@
 
                     return ResponseHelper::createResponseFromBase(statusCode: 200, data: $user);
                 });
+                Route::get('/user/notifications', [NotificationController::class, 'userIndex']);
+                Route::post('/user/notifications/read', [NotificationController::class, 'markRead']);
```
SirSplasch: Put these after the login-url route

src/Http/Controllers/NotificationController.php
```diff
@@ -0,0 +1,53 @@
...
+                'url' => data_get($notification->data, 'accept.url'),
+                'read_at' => $notification->read_at,
+                'created_at' => $notification->created_at,
+            ]);
+
+        return ResponseHelper::createResponseFromBase(
+            statusCode: 200,
+            data: $notifications,
+            additions: ['unread_count' => $user->unreadNotifications()->count()],
+        )->setEncodingOptions(JSON_UNESCAPED_SLASHES);
```
SirSplasch: wrap ->setEncodingOptions()

src/Http/Controllers/NotificationController.php
```diff
@@ -0,0 +1,53 @@
...
+namespace FluxErp\Http\Controllers;
+
+use FluxErp\Helpers\ResponseHelper;
+use FluxErp\Http\Requests\MarkNotificationsReadRequest;
+use Illuminate\Http\JsonResponse;
+use Illuminate\Http\Request;
+
+class NotificationController extends Controller
+{
+    public function userIndex(Request $request): JsonResponse
```
SirSplasch, line 12: You are retrieving the latest 30 notifications, and number of currently unread notifications.
This sparks the question what if there are unread notifications, that are past the latest 30 notifications, these will be effectively unaccessible via the api and the user cannot mark them as read.

src/Http/Requests/MarkNotificationsReadRequest.php
```diff
@@ -0,0 +1,20 @@
...
+namespace FluxErp\Http\Requests;
+
+class MarkNotificationsReadRequest extends BaseFormRequest
+{
+    public function rules(): array
+    {
+        return [
+            'id' => [
+                'nullable',
+                'string',
```
SirSplasch, line 16: Maybe add ModelExists rule?

src/Http/Controllers/NotificationController.php
```diff
@@ -0,0 +1,53 @@
...
+
+    public function markRead(MarkNotificationsReadRequest $request): JsonResponse
+    {
+        $validated = $request->validated();
+        $user = $request->user();
+
+        if (data_get($validated, 'all')) {
+            $user->unreadNotifications->markAsRead();
+        } elseif (! blank(data_get($validated, 'id'))) {
+            $user->notifications()->whereKey(data_get($validated, 'id'))->first()?->markAsRead();
```
SirSplasch: wrap

SirSplasch: If the given id is invalid, like it simply does not exist or belongs to another user, shouldn't this endpoint return a validation exception (422), as nothing could be marked as read?

SirSplasch (changes requested):

Slupi:
Addressed all points in 1b1687ff3:
- Moved the routes after `/user/login-url`; wrapped the chained `->setEncodingOptions()` / mark-read calls.
- `userIndex` now paginates (per_page/page, `->through()`, `->appends()`) so notifications past the first page stay reachable. (Marking by id already works for any of the user's notifications via the new rule below, not just listed ones.)
- `id` validated with a user-scoped `ModelExists` on `Notification` → an unknown or foreign id returns 422 (new test).
- `required_without` both ways + `prohibits:all` so exactly one of `id`/`all` is required (new test).
- Mark-all uses a single bulk `update(['read_at' => now()])`.

5 tests green, Pint clean.

src/Http/Controllers/NotificationController.php
Slupi: userIndex is now paginated (per_page/page), so all notifications are reachable, not just the latest 30. And mark-read takes any notification id the user owns (validated via a user-scoped ModelExists rule), so even items beyond the first page can be marked read without having to list them first.

SirSplasch (approved):

PR #1849, opened 2026-06-26
Add token-authed user tasks + tickets endpoints

Slupi:
Expose GET /api/user/tasks (open tasks the user is responsible for or assigned to) and GET /api/user/tickets (open tickets assigned to the user) for the Nuxbe Companion extension, normalized to id/name|title/state/url. End states are computed dynamically and the queries/orderings mirror the My-* dashboard widgets.

## Summary by Sourcery

Add authenticated user-focused task and ticket listing endpoints for use by external clients.

New Features:
- Expose GET /api/user/tasks to return non-final tasks where the authenticated user is responsible or assigned, normalized for external consumption.
- Expose GET /api/user/tickets to return non-final tickets assigned to the authenticated user, normalized with key ticket attributes and URLs.

Tests:
- Add feature tests covering authorization, filtering, and response shape for the new user tasks and tickets API endpoints.

### DIFF
--- a/src/Http/Controllers/TaskController.php
+++ b/src/Http/Controllers/TaskController.php
@@ -15,17 +15,12 @@ class TaskController extends Controller
     {
         $userId = $request->user()->getKey();
 
-        $endStates = TaskState::all()
-            ->filter(fn (string $state): bool => $state::$isEndState)
-            ->keys()
-            ->toArray();
-
         $tasks = resolve_static(Task::class, 'query')
             ->where(function (Builder $query) use ($userId): void {
                 $query->where('responsible_user_id', $userId)
                     ->orWhereRelation('users', 'users.id', $userId);
             })
-            ->whereNotIn('state', $endStates)
+            ->whereNotIn('state', TaskState::endStateNames())
             ->orderByDesc('priority')
             ->orderByRaw('ISNULL(due_date), due_date ASC')
             ->get(['id', 'name', 'state'])
--- a/src/Http/Controllers/TaskController.php
+++ b/src/Http/Controllers/TaskController.php
@@ -20,14 +20,16 @@ class TaskController extends Controller
                 $query->where('responsible_user_id', $userId)
                     ->orWhereRelation('users', 'users.id', $userId);
             })
-            ->whereNotIn('state', TaskState::endStateNames())
+            ->whereNotIn('state', TaskState::endStateKeys())
             ->orderByDesc('priority')
             ->orderByRaw('ISNULL(due_date), due_date ASC')
-            ->get(['id', 'name', 'state'])
+            ->get(['id', 'name', 'state', 'priority', 'due_date'])
             ->map(fn (Task $task): array => [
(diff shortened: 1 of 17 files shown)


src/States/EndableState.php
```diff
@@ -5,4 +5,17 @@
 abstract class EndableState extends State
 {
     public static bool $isEndState = false;
+
+    /**
+     * The names of the states that end the lifecycle.
+     *
+     * @return array<int, string>
+     */
+    public static function endStateNames(): array
```
SirSplasch: Isn't this more like endStateKeys or something alike? As this returns keys rather than "names".
For example in the MyTasks widget the name getEndStates is used. I am just not happy with the naming of this method.

SirSplasch: As the functionality of this method is used on several occasions replace them all properly.

src/Http/Controllers/TaskController.php
```diff
@@ -0,0 +1,37 @@
...
+        $tasks = resolve_static(Task::class, 'query')
+            ->where(function (Builder $query) use ($userId): void {
+                $query->where('responsible_user_id', $userId)
+                    ->orWhereRelation('users', 'users.id', $userId);
+            })
+            ->whereNotIn('state', TaskState::endStateNames())
+            ->orderByDesc('priority')
+            ->orderByRaw('ISNULL(due_date), due_date ASC')
+            ->get(['id', 'name', 'state'])
+            ->map(fn (Task $task): array => [
```
SirSplasch, line 27: You are ordering by priority and due_date, but do not return them in the response?

tests/Feature/Api/UserTicketsTest.php
SirSplasch, line 1: Seperate this in two different tests as this affects two different routes and controllers.

SirSplasch (changes requested):

src/States/EndableState.php
Slupi: Renamed

Slupi: Replaced every duplicated XxxState::all()->filter($isEndState)->keys()->toArray() block (controllers, endTaskRemindersJob, and  8 widgets incl. MyTickets/UnassignedTickets) with endStateKeys(); removed the redundant per-widget getEndStates() wrappers for Task/Ticket. (Lead widgets keep their own getEndStates() — they're DB-backed lead_state_id, a different mechanism.) 18 call sites now use the one method.

src/Http/Controllers/TaskController.php
Slupi: TaskController::userIndex now selects + returns priority and due_date (Y-m-d)

tests/Feature/Api/UserTicketsTest.php
Slupi: Split UserTasksTicketsTest into UserTasksTest and UserTicketsTest, each with its own permission setup.

SirSplasch (approved):

PR #1865, opened 2026-07-02
fix: prevent empty Meilisearch settings payload in flux-scout:sync-index-settings

patrickweh:
## Problem

After the 1.4.4 release, every deploy that runs `flux-scout:sync-index-settings` fails with:

```
Meilisearch\Exceptions\ApiException: Invalid value type: expected an object, but found an array: `[]`
```

Since #1853, `Searchable::scoutIndexSettings()` always returned an array containing an `embedders` key (`null` when semantic search is disabled). This caused two regressions:

- Models without configured index settings no longer returned `null`, so the command no longer skipped them. Scout's `MeilisearchEngine::updateIndexSettings()` strips the `embedders` key and calls `updateSettings([])`, which serializes to `[]` instead of an object and is rejected by Meilisearch with a 400 error.
- Model overrides using `baseScoutIndexSettings() ?? [fallback]` (Product, Category, SerialNumber, ...) never fell back to their default `filterableAttributes` anymore, silently dropping them on the next settings sync.

## Solution

- `scoutIndexSettings()` returns only the configured keyword settings again (`null` when nothing is configured), restoring the `?? [fallback]` behavior of the model overrides.
- The embedder definition is exposed via a new `scoutEmbedders()` method and injected centrally by the command after the model settings are resolved.
- When a model ends up with embedder-only settings, the command calls `updateEmbedders()` on the index directly instead of going through `updateIndexSettings()`, which would send the empty keyword settings payload.

Note: Scout never forwards `embedders => null` to Meilisearch, so disabling semantic search leaves previously synced embedders on the index. This matches the behavior before this fix and can be addressed separately if needed.

## Summary by Sourcery

Guard Meilisearch index settings syncs against empty payloads and separate semantic embedders configuration from base Scout index settings.

Bug Fixes:
- Avoid sending empty Meilisearch settings payloads when only semantic embedders are configured.
- Restore null index-settings behavior so models without configured settings are skipped and model-level fallbacks work again.

Enhancements:
- Expose Meilisearch embedders via a dedicated scoutEmbedders() API and inject them centrally during index settings sync.

Tests:
- Add unit tests for the new scoutEmbedders() behavior and for models with no configured index settings.
- Add feature tests covering sync-index-settings behavior for disabled semantic search, embedder-only syncs, and combined keyword+embedder syncs.

### DIFF
--- a/src/Console/Commands/Scout/SyncIndexSettingsCommand.php
+++ b/src/Console/Commands/Scout/SyncIndexSettingsCommand.php
@@ -5,6 +5,7 @@ namespace FluxErp\Console\Commands\Scout;
 use FluxErp\Traits\Scout\Searchable;
 use Illuminate\Database\Eloquent\Relations\Relation;
 use Illuminate\Database\Eloquent\SoftDeletes;
+use Illuminate\Support\Arr;
 use Illuminate\Support\Collection;
 use Laravel\Scout\Console\SyncIndexSettingsCommand as BaseSyncIndexSettingsCommand;
 use Laravel\Scout\EngineManager;
@@ -45,13 +46,8 @@ class SyncIndexSettingsCommand extends BaseSyncIndexSettingsCommand
             ->toArray();
 
         foreach ($searchableModels as $model) {
-            if (! method_exists($model, 'scoutIndexSettings')) {
-                $this->error('The model [' . $model . '] does not have a "scoutIndexSettings" method.');
+            $settings = $model::scoutIndexSettings() ?? [];
 
-                continue;
-            }
-
-            $settings = $model::scoutIndexSettings();
             if (
                 config('scout.soft_delete', false)
                 && in_array(SoftDeletes::class, class_uses_recursive($model))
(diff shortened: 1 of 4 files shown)


src/Traits/Scout/Searchable.php
```diff
@@ -34,16 +34,21 @@ public static function search($query = '', $callback = null): Builder
...
 
-        // Semantic search: inject (or reset via null) the Meilisearch embedder from SearchSettings.
-        if (config('scout.driver') === 'meilisearch') {
-            $settings['embedders'] = ($search = static::activeSearchSettings())
-                ? ['default' => static::embedderDefinition($search)]
-                : null;
+    /**
+     * The Meilisearch embedders derived from SearchSettings, or null when semantic search is off.
+     */
+    public static function scoutEmbedders(): ?array
```
SirSplasch: scoutEmbedders before scoutIndexSettings (alphabetical ordering)

src/Console/Commands/Scout/SyncIndexSettingsCommand.php
```diff
@@ -60,13 +56,26 @@ public function handle(EngineManager $manager): void
...
+                continue;
+            }
+
+            $indexName = $this->indexName($model);
+
+            if (Arr::except($settings, 'embedders')) {
+                $engine->updateIndexSettings($indexName, $settings);
+            } else {
+                // Meilisearch rejects an empty settings payload, sync the embedders directly.
+                $engine->index($indexName)->updateEmbedders($settings['embedders']);
```
SirSplasch: The 'embedders' key may not exist on $settings array, make it null safe.

SirSplasch (changes requested):

SirSplasch (approved):

PR #1866, opened 2026-07-02
fix: exclude alternative positions from child order subtotal display

patrickweh:
## Problem

When creating a retoure or split order, the client-side subtotal and total in the create child order view included alternative positions. Order totals correctly exclude alternative positions everywhere else, so the displayed amount did not match the amount of the order that would actually be created. Alternative positions were also not recognizable as such in the selected positions list.

## Solution

- `CreateChildOrder::takeOrderPositions` now includes the `is_alternative` flag in the position payload
- The Alpine `calculatePositionTotal` helper returns 0 for alternative positions, so subtotal and total match the resulting order
- Alternative positions are marked with a badge in the selected positions list

## Summary by Sourcery

Ensure child order creation correctly handles alternative positions in both backend payloads and frontend calculations.

Bug Fixes:
- Exclude alternative positions from child order subtotal and total calculations so displayed amounts match the resulting order.
- Preserve the alternative position flag when replicating positions to created child orders.

Enhancements:
- Expose the alternative position flag in the create child order payload and visually mark such positions with a badge in the UI.

Tests:
- Add coverage to verify alternative positions are marked in the payload, retained on saved retour orders, and excluded from totals.

### DIFF
--- a/src/Livewire/Order/CreateChildOrder.php
+++ b/src/Livewire/Order/CreateChildOrder.php
@@ -241,6 +241,7 @@ class CreateChildOrder extends Component
                 'discount_percentage',
                 'is_net',
                 'is_free_text',
+                'is_alternative',
             ])
             ->whereKey(array_merge($realPositionIds, $freeTextIds))
             ->get();
@@ -259,6 +260,7 @@ class CreateChildOrder extends Component
                     'discount_percentage' => 0,
                     'is_net' => $orderPosition->is_net,
                     'is_free_text' => true,
+                    'is_alternative' => $orderPosition->is_alternative,
                     'unit_abbreviation' => null,
                 ];
 
@@ -286,6 +288,7 @@ class CreateChildOrder extends Component
                     'total_gross_price' => $orderPosition->total_gross_price,
                     'discount_percentage' => $orderPosition->discount_percentage,
                     'is_net' => $orderPosition->is_net,
+                    'is_alternative' => $orderPosition->is_alternative,
                     'unit_abbreviation' => $orderPosition->product?->unit?->abbreviation,
                 ];
             }
--- a/src/Livewire/Order/CreateChildOrder.php
+++ b/src/Livewire/Order/CreateChildOrder.php
@@ -239,9 +239,9 @@ class CreateChildOrder extends Component
                 'total_net_price',
                 'total_gross_price',
                 'discount_percentage',
-                'is_net',
-                'is_free_text',
                 'is_alternative',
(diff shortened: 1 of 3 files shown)


src/Livewire/Order/CreateChildOrder.php
```diff
@@ -241,6 +241,7 @@ public function takeOrderPositions(): void
                 'discount_percentage',
                 'is_net',
                 'is_free_text',
+                'is_alternative',
```
SirSplasch, line 242: order select matching table ordering

src/Livewire/Order/CreateChildOrder.php
```diff
@@ -259,6 +260,7 @@ public function takeOrderPositions(): void
                     'discount_percentage' => 0,
                     'is_net' => $orderPosition->is_net,
                     'is_free_text' => true,
+                    'is_alternative' => $orderPosition->is_alternative,
```
SirSplasch, line 261: match table ordering

src/Livewire/Order/CreateChildOrder.php
```diff
@@ -286,6 +288,7 @@ public function takeOrderPositions(): void
                     'total_gross_price' => $orderPosition->total_gross_price,
                     'discount_percentage' => $orderPosition->discount_percentage,
                     'is_net' => $orderPosition->is_net,
+                    'is_alternative' => $orderPosition->is_alternative,
```
SirSplasch, line 290: match table ordering

SirSplasch (changes requested):
fix failed test

src/Livewire/Order/CreateChildOrder.php
patrickweh: Fixed, reordered to match the table.

src/Livewire/Order/CreateChildOrder.php
patrickweh: Fixed, reordered to match the table.

src/Livewire/Order/CreateChildOrder.php
patrickweh: Fixed, reordered to match the table.

patrickweh:
Extracted a shared test position helper for the duplicated factory setup. Kept the truthy check on `position.is_alternative`: the payload value comes from the model's boolean cast, and a missing flag falling back to falsy matches the previous behavior, while `=== true` would silently break on any non-strict-boolean value.

SirSplasch (approved):

PR #1868, opened 2026-07-03
Expose my-work widgets via API and drop user ticket/task endpoints

Slupi:
- `GET /api/widgets/my-tickets`, `/api/widgets/my-tasks` and `/api/widgets/overtime-balance-box`: the dashboard widgets now serve token clients directly via `RespondsToApiRequests` (mount guarded for plain Component widgets), so widget and API can't drift; tickets carry a company-first `creator` label (new `Ticket::getCreatorLabel()`), `CurrentWorkTimeModel` additionally returns a numeric `dailyTarget`
- `MyTasks` now shows assigned **or** responsible tasks (matching the endpoint it replaces); its Show deep-link filter widened to match
- Removes `/api/user/tickets` + `/api/user/tasks` and their controllers/tests (both were added in #1849 solely for the browser extension, which switches to the widget endpoints)

## Summary by Sourcery

Expose dashboard widgets for my tickets, my tasks, and overtime balance via API and remove legacy user ticket/task endpoints.

New Features:
- Add API responses for MyTickets and MyTasks widgets including ticket creator labels and task details.
- Expose the OvertimeBalanceBox widget over the API with validated parameters and structured numeric values for overtime and work-time targets.
- Extend the CurrentWorkTimeModel widget API to include numeric daily work-time targets.

Enhancements:
- Align MyTasks widget filtering with the new API by including both assigned and responsible tasks and updating its deep-link filter.
- Provide a company-first creator label for tickets based on address and contact data.
- Make RespondsToApiRequests resilient to widgets without a mount method.

Tests:
- Add API tests for my-tickets, my-tasks, and overtime-balance-box widgets including permission checks and validation.
- Remove obsolete tests for deprecated /api/user/tasks and /api/user/tickets endpoints.

Chores:
- Remove deprecated /api/user/tasks and /api/user/tickets routes and their controllers.


src/Livewire/Widgets/MyTickets.php
```diff
@@ -68,18 +74,45 @@ public function placeholder(): View|Factory
...
+                'authenticatable' => fn (MorphTo $morphTo): MorphTo => $morphTo
+                    ->constrain([
+                        Address::class => fn (Builder $query) => $query
+                            ->select(['id', 'name', 'company', 'contact_id']),
+                        User::class => fn (Builder $query) => $query
+                            ->select(['id', 'name']),
+                    ])
+                    ->morphWith([
+                        Address::class => ['contact:id,main_address_id', 'contact.mainAddress:id,company'],
+                    ]),
```
SirSplasch: Does this work with customized Address and/or User model?

src/Models/Ticket.php
```diff
@@ -136,6 +136,19 @@ public function getContactId(): ?int
             : null;
     }
 
+    public function getCreatorLabel(): ?string
```
SirSplasch: creator can in some kind of form be misleading, as the method returns the label for the authenticatable morph, not the one who actually created it.

src/Models/Ticket.php
```diff
@@ -136,6 +136,19 @@ public function getContactId(): ?int
...
+    {
+        $creator = $this->authenticatable;
```
SirSplasch: Why don't you call the getLabel() method on $this->authenticatable?

src/Livewire/Widgets/MyTickets.php
```diff
@@ -68,18 +74,45 @@ public function placeholder(): View|Factory
         return view('flux::livewire.placeholders.horizontal-bar');
     }
 
+    public function toApiResponse(): array
+    {
+        return $this->getTickets()
+            ->take($this->limit)
```
SirSplasch, line 76: Are you sure about reducing the amount to $limit in the api? This is a fixed number and if there are more records they are not getting exposed.

src/Livewire/Widgets/MyTasks.php
```diff
@@ -80,7 +83,7 @@ public function showInTaskList(): void
         SessionFilter::make(
             Livewire::new(resolve_static(TaskList::class, 'class'))->getCacheKey(),
             fn (Builder $query) => $query
-                ->whereRelation('users', 'users.id', $userId)
+                ->assignedOrResponsible($userId)
```
SirSplasch: There is a difference between MyTasks and MyResponsibleTasks. This widget shows only the tasks which the user is assigned to. The other widget shows the tasks which the user is responsible for, but the user does not necessarily do that task themselves.

src/Livewire/Widgets/Employee/OvertimeBalanceBox.php
```diff
@@ -40,13 +47,35 @@ public function calculateSum(): void
...
+        return [
+            'employeeId' => [
+                'required',
+                'integer',
+                app(ModelExists::class, ['model' => Employee::class]),
+            ],
+        ];
+    }
+
+    protected function apiResponseProperties(): array
```
SirSplasch: apiResponseProperties before apiRules to keep alphabetical ordering.
Please adjust the already existing api widgets as well. I know they are not of your making, I failed to comment on this beforehand.

SirSplasch (changes requested):

src/Livewire/Widgets/MyTickets.php
Slupi: No, it didn't — constrain()/morphWith() keys are matched against the resolved morph class, so overridden models would silently skip the constraints. Removed the whole block in 17e6f7a58: with getLabel() (see thread below) the plain `with('authenticatable:id,name')` eager load is enough and works with customized models.

src/Models/Ticket.php
Slupi: Removed the method entirely in 17e6f7a58; the api field is renamed `creator` → `authenticatable`.

src/Models/Ticket.php
Slupi: Done in 17e6f7a58 — the extra company logic was redundant anyway: `Address::name` is already built as "Company, Firstname Lastname" in the saving hook, so getLabel() carries the company. The response now returns `$ticket->authenticatable?->getLabel()`.

src/Livewire/Widgets/MyTickets.php
Slupi: Changed in 17e6f7a58: `limit` is now a validated api parameter (`integer|min:1`, default 25) on my-tickets and my-tasks, so clients can request as many records as they need.

src/Livewire/Widgets/MyTasks.php
Slupi: Reverted in 17e6f7a58. The scope is now `Task::assignedTo()` (users relation only) and is shared by showInTaskList() and getTasks(), so the widget filter and the api can't drift. Responsible-only tasks are excluded again — covered by tests in MyTasksTest and MyTasksWidgetTest.

src/Livewire/Widgets/Employee/OvertimeBalanceBox.php
Slupi: Reordered in 17e6f7a58 — adjusted CurrentWorkTimeModel and the RespondsToApiRequests trait as well.

src/Livewire/Widgets/MyTasks.php
```diff
@@ -94,11 +97,32 @@ public function placeholder(): View|Factory
...
+        ];
+    }
+
     protected function getTasks(): Collection
     {
-        return auth()
-            ->user()
-            ->tasks()
+        return resolve_static(Task::class, 'query')
+            ->assignedTo(auth()->id())
```
SirSplasch: I think I would prefer the current implementation with auth()->user()->tasks(), as this allows for different authenticatable types to use this widget.

SirSplasch (changes requested):

src/Livewire/Widgets/MyTasks.php
Slupi: Reverted to `auth()->user()->tasks()` in 9c465ecdf. The `assignedTo` scope stays only in showInTaskList(), where the session filter operates on a plain Task query without an authenticatable instance.

SirSplasch (approved):

PR #1877, opened 2026-07-07
Fix destructive same-folder drag and inconsistent media attachment in folder tree

patrickweh:
## Root cause

Two bugs in the media/attachments folder tree drag-and-drop:

1. **Destructive same-folder drag.** Dragging a file within the same folder still went through Spatie's `Media::move()`, which is implemented as `copy()` + `forceDelete()`: the media row was silently deleted and recreated with a new id and a re-copied file, breaking anything referencing the old media id.
2. **Inconsistent attachment when moving between a folder and a fixed collection.** `moveMedia()` reused the subject's old `collection_name` (the folder slug) as the target collection instead of the target collection's own name. The media ended up attached to the owner model under the folder's slug — an inconsistent state that made the file invisible in the tree, because folder children are resolved via `model_type`/`model_id` while collection children are resolved via the collection slug.

## Fix

- `moveMedia()` detects when the media is already in the target folder/collection and skips `Media::move()` in that case, so a same-folder drag keeps the existing media row and file.
- `moveMedia()` now always uses the resolved target path as the new collection name, so moving between a folder and a fixed collection produces a consistent attachment in both directions: folder targets get `model_type = media_folder` / `model_id = <folder id>` with the folder slug as collection (matching native uploads into a folder), collection targets get the owner model with the target collection name.

Ordering is unchanged: files remain sorted alphabetically by name.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/Livewire/Support/FolderTree.php
```diff
@@ -421,11 +429,18 @@ protected function moveMedia(array $subject, array $target, string $targetPath,
...
-        $media->move($targetModel, $collectionName);
+        $isAlreadyInTarget = $targetType === 'folder'
+            ? $media->model_type === morph_alias(MediaFolder::class)
+                && (int) $media->model_id === (int) data_get($target, 'id')
+            : $media->model_type === morph_alias($this->modelType)
+                && (int) $media->model_id === (int) $this->modelId
+                && $media->collection_name === $collectionName;
+
+        if (! $isAlreadyInTarget) {
+            $media->move($targetModel, $collectionName);
```
SirSplasch: Now $collectionName is just an alias for $targetPath, so in this case you can remove this variable.
I am curious why you removed the data_get($subject, 'collection_name') in the calculation of the collection name?

SirSplasch (changes requested):

src/Livewire/Support/FolderTree.php
patrickweh: Done, removed the variable. The data_get reused the subject's old collection_name, which is the folder slug when dragging out of a folder. The media then stayed attached to the owner model under that slug and disappeared from the tree, since folder children resolve via model_type/model_id but collection children via the collection slug. targetPath is always the resolved target collection, covered both directions with a test.

SirSplasch (approved):

PR #1878, opened 2026-07-07
feat(media): allow authenticated users to access private media without a signature

patrickweh:
Private media are served through the `media.private` (`/media-private/{media}/{filename}`) and `media.show` (`/media/{media}`) routes, which previously required a valid signed URL for every request.

## Behavior

A new `AuthenticatedUserMediaOrSignature` middleware (aliased `media.signed`) now gates both routes:

- A request with a **valid signature** is served, unchanged. This preserves the guest / embed / e-mail-link flow.
- An **authenticated `FluxErp\Models\User`** is served **without a signature**. Internal users already have full access to private media today, so requiring a signature for them added no security value.
- Any **other authenticatable** (portal `Address` logins, etc.) and **guests without a valid signature** are rejected with `403`.

## Security model

The `instanceof User` check is the authorization guard: only internal users get the signature bypass; portal `Address` sessions still need a valid signature.

For an authenticated user the route media id is resolved against a scoped exists query on `Media` (`resolve_static(Media::class, query)->whereKey($id)->exists()`). Core registers no visibility scope on `Media`, so any user passes, matching current behavior. This query is the extension point: a customer can register their own global scope on `Media` to restrict visibility, and a hidden media then returns `404`.

`media-collection.download` is unrelated (token-based, not media-bound) and keeps the plain `signed` middleware.

## Tests

`tests/Feature/Web/PrivateMediaTest.php` covers: guest with a valid signature (served), guest without a signature (403), authenticated user without a signature (served), authenticated non-user without a signature (403), and an authenticated user denied (404) when a global scope hides the media.

## Summary by Sourcery

Allow authenticated internal users to access private media without signed URLs while preserving signature-based access for guests and non-user authenticatables.

New Features:
- Introduce AuthenticatedUserMediaOrSignature middleware to gate private media routes based on either a valid signature or an authenticated internal user.
- Allow private media routes to bypass URL signatures for authenticated FluxErp\Models\User instances while still enforcing signatures for other actors.

Enhancements:
- Wire the new media.signed middleware into media.private and media.show routes and keep media-collection.download on the existing signed middleware.
- Ensure private media visibility respects customizable global scopes on the Media model, returning 404 when media is hidden.

Tests:
- Expand private media feature tests to cover guest signature access, unauthenticated rejection, internal user access without signatures, non-user rejection, and 404 behavior when a global Media scope hides the resource.

### DIFF
--- a/routes/frontend/web.php
+++ b/routes/frontend/web.php
@@ -382,8 +382,9 @@ Route::middleware('web')
                 ->defaults('html', false);
         });
 
-        Route::middleware('signed')->group(function (): void {
+        Route::group([], function (): void {
             Route::get('/media-private/{media}/{filename}', PrivateMediaController::class)
+                ->middleware('media.signed')
                 ->name('media.private');
 
             Route::get('/media/{media}', function (Media $media) {
@@ -413,6 +414,7 @@ Route::middleware('web')
                     ['Content-Disposition' => $disposition],
                 );
             })
+                ->middleware('media.signed')
                 ->name('media.show');
 
             Route::get('/media-collection-download/{token}', function (string $token) {
@@ -429,6 +431,7 @@ Route::middleware('web')
                     ['Content-Type' => 'application/zip'],
                 );
             })
+                ->middleware('signed')
                 ->name('media-collection.download');
         });
     });
--- a/routes/frontend/web.php
+++ b/routes/frontend/web.php
@@ -366,7 +366,7 @@ Route::middleware('web')
             })->name('media');
         });
 
-        Route::group(['middleware' => ['auth:web']], function (): void {
+        Route::middleware('auth:web')->group(function (): void {
             Route::any('/search/{model?}', SearchController::class)
                 ->where('model', '(.*)')
                 ->name('search');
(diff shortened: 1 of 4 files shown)


src/Http/Middleware/AuthenticatedUserMediaOrSignature.php
```diff
@@ -0,0 +1,39 @@
...
+        // Guests (and embeds) keep the signed-URL contract.
+        if ($request->hasValidSignature()) {
+            return $next($request);
+        }
+
+        // Internal users already have full access to private media. Portal logins
+        // (Address) and any other authenticatable still require a valid signature.
+        if (Auth::user() instanceof User) {
+            $media = $request->route('media');
+            $mediaId = $media instanceof Media ? $media->getKey() : $media;
```
SirSplasch: Allow full customization of user/media model.
is_a(..., resolve_static()) instead of the instanceof check.

routes/frontend/web.php
```diff
@@ -382,8 +382,9 @@
                 ->defaults('html', false);
         });
 
-        Route::middleware('signed')->group(function (): void {
+        Route::group([], function (): void {
```
SirSplasch: keep the original group, just add another Route::middleware('media.signed')->group() and move the two endpoints into this new group.

routes/frontend/web.php
SirSplasch: Eventually change this to Route::middleware()->group() to keep it consistent.

SirSplasch (changes requested):

SirSplasch (approved):

PR #1883, opened 2026-07-09
Add category form stack seam

patrickweh:
Adds a single `@stack('settings-category-form')` inside the category edit modal content section, so packages can render their own per-category fields without overriding the core component. No core logic — just the stack.

### DIFF
--- a/resources/views/livewire/settings/categories.blade.php
+++ b/resources/views/livewire/settings/categories.blade.php
@@ -77,6 +77,7 @@
                     ]"
                         />
                     </div>
+                    @stack('settings-category-form')
                 @show
             </div>
             <x-slot:footer>
--- a/resources/views/livewire/settings/categories.blade.php
+++ b/resources/views/livewire/settings/categories.blade.php
@@ -58,23 +58,23 @@
                             select="label:label|value:id"
                             unfiltered
                             :request="[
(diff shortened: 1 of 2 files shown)


resources/views/livewire/settings/categories.blade.php
```diff
@@ -77,6 +77,7 @@
                     ]"
```
SirSplasch: fix indentations

SirSplasch (changes requested):

resources/views/livewire/settings/categories.blade.php
patrickweh: Fixed the indentation of the select's :request array (8c6bcfb).

src/Livewire/Settings/Categories.php
```diff
@@ -100,6 +104,9 @@ public function save(): bool
             return false;
         }
 
+        // Let extensions persist their own per-category data after the core save.
+        $this->dispatch('category-saved', categoryId: $this->category->id);
```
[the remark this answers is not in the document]
patrickweh: These dispatch events were removed; the PR now adds only the @stack seam, so this no longer applies.

SirSplasch (approved):

PR #1891, opened 2026-07-14
Add the department field to the new-contact dialog

patrickweh:
The address `department` field already exists on the addresses table, the create/update address rulesets, `AddressForm` and the address edit view, but the create-contact form did not expose it. A department could therefore only be filled in after the contact was created.

Add `department` to `ContactForm` (auto-rendered as a labelled input) and include it in the flat address fields mapped into the new contact's main address, so it can be set directly in the new-contact dialog.

Covered by a test asserting the form maps `department` into `main_address`.

## Summary by Sourcery

Expose the existing address department field in the new-contact flow and ensure it is persisted on the main address.

New Features:
- Allow entering a department when creating a new contact so it is stored on the main address.

Tests:
- Add a Livewire ContactList test verifying the department field is mapped into the new contact's main address.

### DIFF
--- a/src/Livewire/Forms/ContactForm.php
+++ b/src/Livewire/Forms/ContactForm.php
@@ -113,6 +113,8 @@ class ContactForm extends FluxForm
 
     public ?string $company = null;
 
+    public ?string $department = null;
+
     #[RenderAs(
         type: RenderAs::SELECT,
         options: [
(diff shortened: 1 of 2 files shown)


patrickweh:
Closing in favour of doing this as a customer-specific customization in the Ella project instead of a core change.

PR #1901, opened 2026-07-18
feat(accounting): recurring contract payables via purchase subscriptions

patrickweh:
## Problem

Open-ended payables contracts (rent, leasing, insurance) are currently stored as a single purchase order with a 0 EUR total. Assigning the monthly direct debits to that order drives its balance negative and distorts the contact's open items, since a contract without an end date never has a meaningful order balance.

## Solution

Contracts run as `PurchaseSubscription` orders with the current monthly amount; the existing subscription schedule replicates them into real monthly purchase orders that carry the balance. Payments are assigned to those monthly orders, never to the contract itself.

Building blocks:

- **Assignment block**: `CreateOrderTransactionRuleset` rejects any order whose order type is a subscription type. This is the root fix for the distorted balances and covers every assignment path.
- **Document number**: replicated purchase-subscription children get `invoice_number = {contract order number}-{YYYY-MM}` and `invoice_date = performance period start`. `Order::invoice()` falls back to the contract's invoice media (Dauerbeleg), scoped to subscription parents only.
- **AdjustOrderTotal action**: when the actual debit differs from the monthly amount, an explicit, confirmed button in the transaction assignment modal adjusts the monthly order's gross total to the payment amount and accepts the assignment. Only valid for unlocked subscription children with exactly one position; it never runs implicitly.
- **Payment profile matching**: contracts carry a counterpart IBAN (existing `iban` column) and a new `payment_purpose_pattern`. `MatchTransactionsWithOrderJob` matches incoming transactions against these profiles first and suggests the oldest open monthly order of the contract as an unaccepted suggestion. The pattern disambiguates multiple contracts with the same creditor.
- **UI**: subscription orders are excluded from the assignment order picker; the purpose pattern is editable next to the IBAN in the order accounting tab.

## Notes

- Existing 0 EUR contract orders and their assignments need one-time manual cleanup (re-assign to generated monthly orders or book to a ledger account); no automated migration.
- **Follow-up (flux-datev)**: DATEV export lines are currently only created by the media-added listener when an invoice document is attached to an order. Generated monthly orders carry `invoice_number`/`invoice_date` (everything the export needs) but no own media, so they never enter the export. flux-datev needs a hook to create export lines for document-less subscription children.
- One media-related test fails locally due to a broken Imagick/Ghostscript setup (pre-existing, reproduced on unmodified tests); CI runs the same pipeline green.

## Summary by Sourcery

Support recurring payables via purchase subscription contracts by keeping contract orders free of direct payment assignments and shifting balance handling to generated child orders.

New Features:
- Introduce payment purpose pattern field on orders to define matching profiles for recurring bank transactions.
- Add an AdjustOrderTotal action and UI control to align generated subscription child orders with actual payment amounts.
- Generate invoice numbers and dates automatically for purchase subscription child orders and allow them to inherit invoice media from their subscription parent.

Bug Fixes:
- Prevent assigning transactions directly to subscription-type orders to avoid distorted contract balances.
- Exclude subscription and purchase-subscription orders from the accounting order picker to keep contract shells out of regular assignment workflows.

Enhancements:
- Extend automatic transaction matching to use subscription contract IBAN and purpose pattern profiles and suggest the oldest open child order.
- Update order invoice resolution to fall back to the parent subscription contract's invoice document for generated children lacking their own media.

Tests:
- Add feature and Livewire tests covering AdjustOrderTotal behavior and UI, subscription profile-based transaction matching, purchase subscription invoice number generation and media fallback, payment purpose pattern persistence, subscription-order assignment blocking, and order list filtering.
- Add a Livewire test ensuring the accounting order list excludes subscription-type orders from its query.


database/migrations/2026_07_17_000001_add_payment_purpose_pattern_to_orders_table.php
```diff
@@ -0,0 +1,22 @@
...
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class() extends Migration
+{
+    public function up(): void
+    {
+        Schema::table('orders', function (Blueprint $table): void {
+            $table->string('payment_purpose_pattern')->nullable()->after('iban');
```
SirSplasch: Why not after bic, so that the bank information stay together?

src/Rulesets/OrderTransaction/CreateOrderTransactionRuleset.php
```diff
@@ -22,6 +24,17 @@ public function rules(): array
                 'required',
                 'integer',
                 app(ModelExists::class, ['model' => Order::class]),
+                function (string $attribute, mixed $value, Closure $fail): void {
```
SirSplasch: I am not a big fan of passing closures into the ruleset. Either add the respective where constraints to the ModelExists rule, or validate it in the validateData method of the corresponding action if you want to keep the specific validation message.
Also do not use __() for the error message, they will be localized by the toast, and in the api they should not be localized.

src/Actions/Order/AdjustOrderTotal.php
```diff
@@ -0,0 +1,98 @@
...
+            throw ValidationException::withMessages([
+                'id' => [__('Only orders with exactly one position can be adjusted.')],
+            ])->errorBag('adjustOrderTotal');
+        }
+
+        if (bccomp($order->orderPositions()->value('amount') ?? 0, 0, 9) !== 1) {
+            throw ValidationException::withMessages([
+                'id' => [__('Only orders with a positive position amount can be adjusted.')],
+            ])->errorBag('adjustOrderTotal');
+        }
```
SirSplasch, line 100: Curious whether this was a conscious decision to not put any of these checks as a constraint on the ModelExists rule?

src/Actions/Order/AdjustOrderTotal.php
```diff
@@ -0,0 +1,98 @@
...
+            ->with('createdFrom.orderType:id,order_type_enum')
+            ->first(['id', 'created_from_id', 'is_locked']);
+
+        if (
+            is_null($order?->created_from_id)
+            || ! $order->createdFrom?->orderType?->order_type_enum?->isSubscription()
+        ) {
+            throw ValidationException::withMessages([
+                'id' => [__('Only generated subscription orders can be adjusted.')],
+            ])->errorBag('adjustOrderTotal');
```
SirSplasch: Wrap ->errorBag() for all of these validation exceptions

src/Actions/Order/AdjustOrderTotal.php
```diff
@@ -0,0 +1,98 @@
...
+            ->withCount('orderPositions')
+            ->with('createdFrom.orderType:id,order_type_enum')
+            ->first(['id', 'created_from_id', 'is_locked']);
+
+        if (
+            is_null($order?->created_from_id)
+            || ! $order->createdFrom?->orderType?->order_type_enum?->isSubscription()
+        ) {
+            throw ValidationException::withMessages([
+                'id' => [__('Only generated subscription orders can be adjusted.')],
```
SirSplasch: Do not put __() around the string, affects all of the validation exception messages.

src/Jobs/Accounting/MatchTransactionsWithOrderJob.php
```diff
@@ -252,6 +263,33 @@ protected function findOrderByPurpose(
...
+            && str_contains($purpose, strtolower($contract->payment_purpose_pattern));
+
+        $matching = ($normalizedIban ? $this->subscriptionContracts->get($normalizedIban, collect()) : collect())
+            ->filter(fn (Order $contract): bool => ! $contract->payment_purpose_pattern || $patternFits($contract))
+            ->merge($this->patternOnlySubscriptionContracts->filter($patternFits));
+
+        // prefer a purpose-pattern hit when several contracts share the creditor iban
+        $contract = $matching->sortByDesc(
+            fn (Order $contract) => $contract->payment_purpose_pattern ? 1 : 0
+        )->first();
```
SirSplasch: wrap
->first();

SirSplasch (changes requested):

src/Actions/Order/AdjustOrderTotal.php
patrickweh: Conscious. Each check has its own message, as `ModelExists` constraints they'd all collapse into one `validation.exists`. Same tradeoff as in the ruleset comment. Can move them if you prefer the generic message.

SirSplasch (approved):

PR #1902, opened 2026-07-19
feat(calendar): upgrade FullCalendar from v6 to v7

patrickweh:
Upgrades FullCalendar from 6.1.21 to 7.0.1.

## Package changes

v7 renames the vanilla-JS distribution from `@fullcalendar/core` to `fullcalendar` and moves every plugin to a subpath entrypoint of that single package, so the six scoped dependencies collapse into one plus the new `temporal-polyfill` peer dependency. The moment-timezone plugin is gone: named time zones (used here for the user's timezone setting) are handled natively via the polyfill.

## Theming

v7 no longer ships bundled CSS and hashes its internal class names, so the previous `.fc-*` overrides no longer match anything. The classic theme plugin and its stylesheets are now registered explicitly, and the calendar is themed through the theme's CSS variables — brand indigo for the primary/today accent, slate borders and backgrounds in dark mode. The handful of layout tweaks that genuinely differ from the theme (toolbar padding, title size, day-header weight, event hover opacity) are applied to stable class names attached through v7's `*Class` options.

Dark mode: the theme reads `data-color-scheme`, which is now bound on the body alongside the existing `dark` class, so it follows the app's theme toggle.

## Option changes

`buttonText` was removed in favour of `buttons`, whose entries are keyed by button name — for view buttons that is the view name (`dayGridMonth`), not the old shorthand (`month`).

## Verification

The existing calendar browser tests cover page load, event create/edit, date click, and drag-and-drop for recurring and non-recurring events; all pass against v7. Light and dark rendering were checked visually against the v6 output — the calendar-week suffix in the toolbar title, today highlighting, other-month dimming and the toolbar layout all match.

## Summary by Sourcery

Upgrade the calendar implementation to FullCalendar v7 and adapt the app’s configuration, theming, and tests to the new distribution and API.

New Features:
- Adopt FullCalendar v7’s classic theme and palette, including CSS variable-driven theming that integrates with the app’s light/dark mode toggle.
- Expose stable CSS hooks for calendar views, headers, and events via v7’s `*Class` options for consistent styling and browser test selectors.

Bug Fixes:
- Align calendar button configuration with FullCalendar v7’s `buttons` option to ensure toolbar labels render correctly.

Enhancements:
- Replace multiple scoped FullCalendar packages and the moment-timezone plugin with the unified v7 `fullcalendar` package and `temporal-polyfill` for timezone handling.
- Refine calendar layout and typography (toolbar padding, title size, day-header weight, event hover) using v7-compatible classes while keeping the visual appearance consistent with the previous version.

Build:
- Update package dependencies to remove FullCalendar v6 packages and add the v7 `fullcalendar` bundle and `temporal-polyfill`.
- Import FullCalendar v7’s skeleton and classic theme stylesheets into the main CSS pipeline.

Tests:
- Update calendar browser tests to wait on the new stable view class instead of the old `.fc` root selector, preserving coverage for calendar interactions.

### DIFF
--- a/resources/js/components/calendar.js
+++ b/resources/js/components/calendar.js
@@ -1,5 +1,3 @@
-import momentTimezonePlugin from '@fullcalendar/moment-timezone';
-
 const calendar = () => {
     return {
         calendarItem: {},
@@ -376,8 +374,15 @@ const calendar = () => {
                     timeGridPlugin,
                     listPlugin,
                     interactionPlugin,
-                    momentTimezonePlugin,
+                    classicThemePlugin,
                 ],
+                // v7 hashes its own class names, so expose stable hooks for the
+                // application stylesheet and the calendar-week title suffix below.
+                viewClass: 'fc-view',
+                toolbarTitleClass: 'fc-toolbar-title',
+                headerToolbarClass: 'fc-header-toolbar',
+                dayHeaderClass: 'fc-day-header',
+                blockEventClass: 'fc-block-event',
                 initialView: 'dayGridMonth',
                 slotDuration: '00:15:00',
                 initialDate: new Date(),
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,19 +6,15 @@
         "": {
             "dependencies": {
                 "@simplewebauthn/browser": "^13.3.0",
-                "@techstark/opencv-js": "^4.12.0-release.1"
+                "@techstark/opencv-js": "^4.12.0-release.1",
+                "fullcalendar": "^7.0.1",
+                "temporal-polyfill": "^1.0.1"
             },
             "devDependencies": {
                 "@alpinejs/collapse": "^3.15.12",
                 "@alpinejs/sort": "^3.15.12",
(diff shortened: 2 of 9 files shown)


resources/js/components/calendar.js
```diff
@@ -1,5 +1,3 @@
-import momentTimezonePlugin from '@fullcalendar/moment-timezone';
```
SirSplasch, line 1: Does the fullcalendar inherit this plugin in v7?

SirSplasch (changes requested):

resources/js/components/calendar.js
patrickweh: Yes, built in. v7 resolves named IANA zones through `temporal-polyfill` directly (`@full-ui/headless-calendar` imports `temporal-polyfill/fns/ZonedDateTime`), which is why it is now a peer dependency. The `timeZone` option only special cases `local` and `UTC`; anything else is passed through as an IANA name, so `auth()->user()?->timezone` keeps working without the plugin.

SirSplasch (approved):

PR #1904, opened 2026-07-20
fix(mail): skip supplier resolution for own tenant domains

patrickweh:
## Problem

Purchase invoices created automatically from incoming mail resolve the supplier from the sender address via `Address::findAddressByEmail()`. When a mail arrives from one of our own domains, for example an employee mailing themselves an invoice, that lookup finds the internal contact and assigns it as the supplier.

## Change

`CreateMailExecutedSubscriber::createPurchaseInvoice()` now drops the resolved contact when the sender domain belongs to a tenant:

```php
$contact = $this->isTenantDomain($message->from) ? null : $this->address?->contact;
```

`isTenantDomain()` normalizes the sender domain (lowercase, part after `@`, scheme, port, path and `www.` prefix removed) and compares it against every `tenants.email` and `tenants.website` value. A tenant matches on an exact domain or on a real subdomain, so `notteam-nifty.test` does not match `team-nifty.test`.

Tenant domains without a dot are ignored. `website` is only validated as `string|max:255`, so a stray value such as `test` would otherwise match every sender under that suffix.

The tenant query uses `withoutEagerLoads()` because `Tenant::$with = ['media']` would otherwise load media for every tenant just to read two columns.

## Scope

Only supplier determination for purchase invoices. Ticket creation, lead creation and the `[flux:comment:...]` assignment keep resolving the sender address unchanged. The invoice is still created, just without a `contact_id`, so it appears in the unassigned purchase invoices widget.

## Known limitations

* Trashed tenants are not considered, since `query()` applies the soft delete scope.
* No IDN normalization. A tenant domain stored as unicode will not match a punycode sender header.

## Summary by Sourcery

Skip automatic supplier assignment for purchase invoices created from mails sent from tenant-owned domains and add coverage for the new domain-matching rules.

Bug Fixes:
- Prevent internal tenant contacts from being incorrectly assigned as suppliers when purchase invoices are created from emails originating from tenant email or website domains.

Enhancements:
- Introduce tenant-domain detection and normalization logic to identify emails originating from tenant-owned domains when creating purchase invoices.

Tests:
- Add unit tests covering purchase invoice creation from mail for various tenant email/website domain combinations, including subdomains and lookalike/suffix domains.

### DIFF
--- a/src/Listeners/MailMessage/CreateMailExecutedSubscriber.php
+++ b/src/Listeners/MailMessage/CreateMailExecutedSubscriber.php
@@ -261,6 +261,10 @@ class CreateMailExecutedSubscriber
     protected function matchesTenantDomain(Tenant $tenant, string $domain): bool
     {
         foreach ([$tenant->email, $tenant->website] as $value) {
+            if (blank($value)) {
+                continue;
+            }
+
             $tenantDomain = $this->normalizeDomain($value);
 
             // `website` is only validated as a string, so ignore anything without a dot
--- a/src/Listeners/MailMessage/CreateMailExecutedSubscriber.php
+++ b/src/Listeners/MailMessage/CreateMailExecutedSubscriber.php
@@ -61,7 +61,8 @@ class CreateMailExecutedSubscriber
 
     public function createPurchaseInvoice(Communication $message): ?Collection
     {
-        $contact = $this->address?->contact;
+        // Mails from one of our own domains never originate from a supplier.
+        $contact = $this->isTenantDomain($message->from) ? null : $this->address?->contact;
 
         $created = [];
         foreach ($message->getMedia('attachments') as $attachment) {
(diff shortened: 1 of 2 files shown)


src/Listeners/MailMessage/CreateMailExecutedSubscriber.php
```diff
@@ -240,6 +241,44 @@ public function subscribe(): array
...
 
+    protected function isTenantDomain(?string $from): bool
+    {
+        $domain = $this->normalizeDomain(Str::between($from ?? '', '<', '>'));
+
+        if (blank($domain)) {
+            return false;
+        }
+
+        return resolve_static(Tenant::class, 'query')
```
SirSplasch, line 255: For the data passed to the CreatePurchaseInvoice action, the default tenant is queried, if this evaluates to true.
But the matching tenant of this method may not be the same as the default tenant. I think this should be taken into account, and the matching tenant should be passed onto the CreatePurchaseInvoice action.

SirSplasch (changes requested):

src/Listeners/MailMessage/CreateMailExecutedSubscriber.php
```diff
@@ -240,6 +244,53 @@ public function subscribe(): array
...
+            ->withoutEagerLoads()
+            ->get(['id', 'email', 'website'])
+            ->first(fn (Tenant $tenant): bool => $this->matchesTenantDomain($tenant, $domain));
+    }
+
+    protected function matchesTenantDomain(Tenant $tenant, string $domain): bool
+    {
+        foreach ([$tenant->email, $tenant->website] as $value) {
+            $tenantDomain = $this->normalizeDomain($value);
```
SirSplasch, line 268: if blank($value) continue, no need to do anything with null or empty string.

SirSplasch (changes requested):

SirSplasch (approved):

PR #1912, opened 2026-07-21
feat(accounting): contract totals with remaining balance and a contract card

patrickweh:
Follow-up to #1901. Fixed-term contracts (loans, fixed-term service contracts) were only visible as their monthly debit; there was no way to record a contract total or see how much of a contract is still open.

## Data model

One new nullable column, `orders.contract_total_amount`: the contract's total volume, manually maintained and editable at any time — variable interest stays representable by entering the bank's updated total. Works for both `Subscription` and `PurchaseSubscription` (sales: remaining backlog, purchase: remaining debit volume).

The remaining amount reuses the contract order's `balance` column and is therefore a real, sortable database field that already has a column in the accounting order list:

```
contract balance = contract_total_amount − Σ |generated rate totals|
```

Contract orders never receive an `invoice_number`, and every `balance` consumer (`scopeUnpaid`, Outstanding/Address widgets, DirectDebit, MoneyTransfer, payment reminder scopes) guards on it, so the reuse leaks into no aggregation. Rate totals are summed absolute because replication stores them signed by the type multiplier while `AdjustOrderTotal` counteracts it.

Recalculation happens when a rate is generated, a rate total is adjusted, or the contract total changes — all through the order `saved` hook, so every write path is covered. Without a total the balance stays `null`; no automatic contract end when the remaining amount reaches zero.

`contract_total_amount` is rejected on non-subscription orders and excluded from `ReplicateOrder`.

## Subscription UX

Previously a subscription did nothing until someone found the "Schedule" sidebar button and configured the generic scheduling dialog. Now:

* Creating a subscription order auto-creates its schedule (monthly, starting on the 1st of next month), so a contract works with zero clicks.
* Both subscription views get a contract card in business language: repeat frequency (monthly/quarterly/yearly/weekly), next execution, end (never/date), total amount with a calculate button (rate × remaining cycles until `ends_at`) and the remaining amount. The full schedule dialog stays reachable via "Advanced".

## Tests

`tests/Feature/ContractRemainingBalanceTest.php`: balance countdown via `ProcessSubscriptionOrder`, refresh through `AdjustOrderTotal`, recalculation on total change, null total untouched, unpaid-scope regression guard, non-subscription rejection, schedule auto-creation for subscriptions and not for other orders.

## Summary by Sourcery

Add explicit total and remaining balance tracking for subscription-based contracts and surface these details in a new contract card UI, with automatic scheduling and validations to ensure only subscription orders carry contract totals.

New Features:
- Introduce a contract card on subscription order views to display repeat schedule, end date, contract total, and remaining balance with a helper to calculate the total from remaining cycles.
- Automatically create a default monthly schedule for new subscription orders so contracts start generating rates without manual setup.

Enhancements:
- Track fixed-term contract volume on subscription orders via a new contract_total_amount field and reuse the order balance to represent remaining contract amount, kept in sync when child rates are created or adjusted.
- Validate that contract_total_amount is only set on subscription order types and support replication and forms/rulesets for the new field.

Tests:
- Add feature tests covering contract balance countdown as rates are generated, balance recalculation on rate adjustments and total changes, unpaid-scope behavior, validation of contract totals on non-subscription orders, and automatic schedule creation for subscription orders.


src/Actions/Order/ReplicateOrder.php
```diff
@@ -79,6 +79,7 @@ public function performAction(): Order
             $orderData['is_imported'],
             $orderData['is_confirmed'],
             $orderData['is_paid'],
+            $orderData['contract_total_amount'],
```
SirSplasch: put this according to orders table, so after 'delivery_state'

src/Rulesets/Order/CreateOrderRuleset.php
```diff
@@ -180,6 +180,7 @@ public function rules(): array
             'header_discount' => 'numeric|min:0|nullable',
             'shipping_costs_net_price' => 'numeric|nullable',
             'margin' => 'numeric|nullable',
+            'contract_total_amount' => 'numeric|nullable',
             'number_of_packages' => 'integer|nullable',
```
SirSplasch: Can this key be removed, I don't see a number_of_packages column on orders table.

src/Rulesets/Order/CreateOrderRuleset.php
```diff
@@ -180,6 +180,7 @@ public function rules(): array
             'header_discount' => 'numeric|min:0|nullable',
             'shipping_costs_net_price' => 'numeric|nullable',
             'margin' => 'numeric|nullable',
+            'contract_total_amount' => 'numeric|nullable',
```
SirSplasch: app(Numeric::class) instead of 'numeric'

src/Rulesets/Order/UpdateOrderRuleset.php
```diff
@@ -152,6 +152,7 @@ public function rules(): array
             'header_discount' => 'numeric|min:0|nullable',
             'shipping_costs_net_price' => 'numeric|nullable',
             'margin' => 'sometimes|numeric|nullable',
+            'contract_total_amount' => 'numeric|nullable',
```
SirSplasch: app(Numeric::class) instead of 'numeric'

src/Rulesets/Order/UpdateOrderRuleset.php
```diff
@@ -152,6 +152,7 @@ public function rules(): array
...
             'number_of_packages' => 'sometimes|integer|nullable',
```
SirSplasch: Can this key be removed, I don't see a number_of_packages column on orders table.

resources/views/components/order/contract-card.blade.php
```diff
@@ -0,0 +1,87 @@
...
+            />
+        </div>
+        @if (! is_null($order->balance) && ! is_null($order->contract_total_amount))
+            <x-alert
+                color="indigo"
+                :title="__('Remaining') . ': ' . \Illuminate\Support\Number::currency(
+                (float) $order->balance,
+                $order->currency?->iso ?? '',
+                app()->getLocale()
+            )"
```
SirSplasch: fix indentation

src/Livewire/Order/Order.php
```diff
@@ -786,6 +788,52 @@ public function saveReplicate(): void
         $this->redirectRoute('orders.id', ['id' => $this->replicateOrder->id], navigate: true);
     }
 
+    #[Renderless]
+    public function prefillContractTotal(): void
```
SirSplasch, line 792: The handling in the frontend can be confusing, as one can call this method on a newly created subscription, without saving the schedule, and therefore will ever run into the warning toast. But the end date can be filled for the user, which then can cause a lot of confusion.

src/Models/Order.php
```diff
@@ -665,6 +687,27 @@ public function refund(): ?Media
...
+        // the remaining contract volume instead of an open invoice amount. Generated
+        // rates store their totals signed inconsistently (replication applies the
+        // type multiplier, AdjustOrderTotal counteracts it), so sum absolutes.
+        if ($this->orderType?->order_type_enum?->isSubscription()) {
+            $this->balance = is_null($this->contract_total_amount)
+                ? null
+                : bcround(
+                    bcsub(
+                        $this->contract_total_amount,
+                        (string) $this->createdOrders()
```
SirSplasch, line 700: But isn't the sum incorrect, if at least one of those orders has a Retoure and therefore is not fully invoiced?

src/Actions/Order/CreateOrder.php
```diff
@@ -65,9 +67,61 @@ public function performAction(): Order
             $order->users()->sync($users);
         }
 
+        // A subscription without a schedule never generates anything, which nobody
+        // expects. Create a sensible monthly default right away; the contract view
+        // exposes it for editing.
+        if ($order->orderType?->order_type_enum?->isSubscription()) {
+            $this->createDefaultSchedule($order);
```
SirSplasch: I understand why you are creating a default schedule right away. Put doesn't this put pressure on the user to modify it as soon as possible, as otherwise the order creates its subscription orders at the first day of the next month? I can see a world, where this actually creates at least a few unwanted subscription orders.
Also what if I create this order, so that it exists, but do not want to have subscription orders created by it from the start, but earliest a few months later.

To me this comes with obvious issues, even if it is crystal clear communicated to the user, which it is currently not.

src/Actions/Order/CreateOrder.php
```diff
@@ -65,9 +67,61 @@ public function performAction(): Order
...
+            ->value('id');
+
+        // Without a target type the schedule would run but never generate anything;
+        // leave it to the user in that case instead of creating a dead schedule.
+        if (is_null($generatedOrderTypeId)) {
+            return;
+        }
+
+        CreateSchedule::make([
+            'name' => ProcessSubscriptionOrder::name(),
```
SirSplasch: Isn't this an issue for customization? Shouldn't it be resolve_static()
I know this is not the first time this is used that way, but it should eventually be adjusted everywhere.

SirSplasch (changes requested):

src/Models/Order.php
patrickweh: Retoures are excluded here. The sum only covers the subscription's generated rates, and a retoure has created_from_id of the rate, not the subscription, so it never enters this sum. A retoure on a running subscription is unlikely anyway, and if it happens you get a separate credit note from the supplier, which is a different process.

SirSplasch (approved):

PR #1913, opened 2026-07-21
feat(accounting): make the payment purpose pattern field self explanatory

patrickweh:
The payment purpose pattern field (order view, Accounting tab, Bank Connection card) suggested a pattern syntax that does not exist — matching in `MatchTransactionsWithOrderJob` is a plain case-insensitive substring check.

## Changes

* Label renamed from "Payment purpose pattern" to **"Payment purpose contains"**, so the field explains its own semantics.
* Placeholder (`e.g. Loan 0047123456`) and a hint: debits whose payment purpose contains this text are assigned to this contract, case does not matter.
* **Live preview while typing** (debounced): the five most recent transactions whose purpose contains the entered text, with counterpart, amount, purpose and booking date — immediate feedback whether the value hits the intended debits. `LIKE` wildcards in the input are escaped and treated as literals, matching the substring semantics of the matching job.

## Tests

Preview returns only matching transactions; `%` and empty input return nothing.

## Summary by Sourcery

Clarify the payment purpose matching semantics and add a live preview of recent matching debits in the order accounting view.

New Features:
- Add a live preview of recent transactions whose payment purpose contains the entered text in the order accounting bank connection card.

Enhancements:
- Rename the payment purpose field label and add placeholder and hint text to make its substring matching behavior self-explanatory.
- Introduce a renderless Livewire method to fetch formatted recent transactions matching the payment purpose input, escaping LIKE wildcards to reflect substring matching semantics.

Tests:
- Add tests verifying that the payment purpose preview returns only substring-matching transactions and that LIKE wildcard characters and empty input yield no results.


src/Livewire/Order/Accounting.php
```diff
@@ -95,6 +97,36 @@ public function editTransaction(?Transaction $transaction): void
...
+        $escaped = addcslashes(strtolower($pattern), '%_\\');
+
+        return resolve_static(Transaction::class, 'query')
+            ->whereRaw('LOWER(purpose) LIKE ?', ['%' . $escaped . '%'])
+            ->with('currency:id,iso')
+            ->latest('booking_date')
+            ->limit(5)
+            ->get(['id', 'currency_id', 'booking_date', 'amount', 'purpose', 'counterpart_name'])
+            ->map(fn (Transaction $transaction): array => [
+                'booking_date' => $transaction->booking_date?->format('d.m.Y'),
```
SirSplasch: ->isoFormat('L')

SirSplasch (changes requested):

SirSplasch (approved):

PR #1918, opened 2026-07-21
fix(ui): repair prompt dialogs broken by Livewire 4 expression evaluation

patrickweh:
Customer report (Drei Masken Verlag): 500 when creating a tag through the green plus button. Their instance predates #1634 and fails server-side with the old `$promptValue()` expression — but reproducing on current main showed the flow is broken here too, just silently.

## Root cause

Livewire 4's action evaluator resolves unknown identifiers against `$wire`. Every `wire:click` carrying a JS expression behind `wire:flux-confirm` dies with:

```
Livewire Expression Error: $wire.$nuxbe.promptValue is not a function
Expression: "addTag($nuxbe.promptValue())"
```

The request never fires, the dialog just closes. Affected: tag creation on products, addresses, tasks, leads, communications, favorites, plugin install/uninstall, the test mail prompt — every `$nuxbe.promptValue()` call site.

A second, masked bug sat behind it: TallStackUI removes the dialog from the DOM **before** the confirm callback runs (`accept()` removes immediately, the callback fires after a 100ms timeout), so `#prompt-value` would not exist anymore even if the expression evaluated.

## Fix — no blade changes

* The `flux-confirm` confirm callback evaluates the `wire:click` expression through **Alpine** (`Alpine.evaluate` with a `$wire.` prefix), where `$wire`, the `$nuxbe` magic and surrounding `x-for` scope all resolve. Plain method calls (`delete()`, `save()`) behave exactly as before.
* `promptValue()` falls back to a snapshot taken on `dialog:accepted` (fires synchronously while the input still exists).

## Test

New browser test drives the real flow: open the prompt via the tag plus button on a product, type a name, confirm, assert the tag row exists and no JS errors surfaced. Red on main, green with the fix.

## Summary by Sourcery

Repair prompt dialog confirm flows by routing wire:click expressions through Alpine and stashing accepted input values, and harden media downloading to return 404s for missing files with supporting tests.

Bug Fixes:
- Restore prompt dialog actions that were broken by Livewire 4 expression evaluation.
- Ensure media downloads return a 404 instead of a 500 when the underlying file is missing.
- Prevent intermittent media download test failures caused by colliding fake file names.

Enhancements:
- Evaluate flux-confirm wire:click expressions through Alpine to maintain access to $wire, magic properties, and loop scope.
- Persist prompt dialog input values via a snapshot taken on dialog acceptance so they remain available after the dialog DOM is removed.

Tests:
- Add a browser test that covers creating a tag via the prompt dialog and asserts the full UI flow works without JavaScript errors.
- Add an API test verifying missing media files result in a 404 response.

### DIFF
--- a/resources/js/components/alpine.js
+++ b/resources/js/components/alpine.js
@@ -163,6 +163,23 @@ document.addEventListener('livewire:init', () => {
                 .wireable(component.id)
                 [type](title, description)
                 .confirm(confirmLabel, () => {
+                    // Livewire 4's own evaluator resolves unknown identifiers
+                    // against $wire, which breaks expressions like
+                    // addTag($nuxbe.promptValue()). Evaluate through Alpine
+                    // instead, where $wire, magics and x-for scope all work.
+                    const expression = el.getAttribute('wire:click');
+
+                    if (expression) {
+                        window.Alpine.evaluate(
+                            el,
+                            expression.trimStart().startsWith('$wire.')
+                                ? expression
+                                : '$wire.' + expression,
+                        );
+
+                        return;
+                    }
+
                     action();
                 })
                 .cancel(cancelLabel)
--- a/resources/js/nuxbe.js
+++ b/resources/js/nuxbe.js
@@ -7,8 +7,32 @@ import './nuxbe/lightbox/handlers/video.js';
 import './nuxbe/lightbox/handlers/audio.js';
 import './nuxbe/lightbox/handlers/fallback.js';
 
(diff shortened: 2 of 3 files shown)


src/Actions/Media/DownloadMedia.php
SirSplasch, line 1: Why are the DownloadMedia changes in this PR as well?

SirSplasch (changes requested):

SirSplasch (approved):

PR #1919, opened 2026-07-22
fix(accounting): exclude unbounced direct debits from dunning and hide sent reminders

patrickweh:
Two issues from first real use of the reworked payment reminder run (#1845).

## Direct debit invoices were dunned before the debit ever bounced

Direct debit invoices are collected by us — reminding the customer while the debit has not even been attempted (or succeeded) is wrong. `wherePaymentReminderEligible` now only admits direct debit orders that carry at least one **not successful** direct debit payment run; money transfer invoices are unaffected. Since the guard sits in the shared scope, the run list, `AutoSendPaymentRemindersJob` and the bundle validation all follow the same rule.

## Sent reminders stayed in the list until a manual reload

The sends run as a queued batch, so the reload directly after dispatch found the database unchanged and kept listing the sent orders — to the user it looked like nothing was sent. Dispatched order ids are now tracked on the component (`#[Locked]`) and hidden optimistically until the batch catches up; the selection is cleared with them. If a send fails, the batch toast reports it and a fresh page load shows the order again.

## Tests

* Scope: direct debit without failed run excluded, with `not_successful` run included, with only `successful` runs excluded
* Component: after `sendSelected` with a faked queue the group list is empty and the selection cleared, no manual reload
* `PaymentTypeFactory` randomizes `is_direct_debit` — all reminder fixtures pin it now

## Summary by Sourcery

Adjust payment reminder eligibility and UI behaviour to avoid dunning unbounced direct debit invoices and to hide sent reminders immediately.

New Features:
- Track dispatched payment reminder order IDs in the payment reminder run component to optimistically hide them from subsequent loads.

Bug Fixes:
- Exclude direct debit orders from payment reminders until at least one direct debit payment run has failed.
- Ensure orders removed by sending payment reminders disappear from the run list and selection without requiring a manual page reload.

Tests:
- Extend payment reminder due scope tests to cover direct debit orders with no runs, failed runs, and only successful runs.
- Add a Livewire test verifying that sending selected reminders clears the groups and selected orders without a manual reload.
- Pin payment type fixtures to non-direct-debit in reminder-related feature and job tests to avoid randomness affecting dunning behaviour.

### DIFF
--- a/src/Livewire/Accounting/PaymentReminderRun.php
+++ b/src/Livewire/Accounting/PaymentReminderRun.php
@@ -31,6 +31,14 @@ class PaymentReminderRun extends Component
     /** Editable recipient email per group, keyed by group key. */
     public array $recipientEmails = [];
 
+    /**
+     * Orders whose reminders were dispatched in this session. The sends run as a
+     * queued batch, so the database still lists them as due when the component
+     * reloads; hide them optimistically instead of forcing a manual reload.
+     */
+    #[Locked]
+    public array $sentOrderIds = [];
+
     public ?string $filterLevel = null;
 
     public ?string $search = null;
@@ -73,6 +81,10 @@ class PaymentReminderRun extends Component
 
         $orders = resolve_static(Order::class, 'query')
             ->wherePaymentReminderDue()
+            ->when(
+                $this->sentOrderIds,
+                fn (Builder $query, array $sentOrderIds) => $query->whereIntegerNotInRaw('id', $sentOrderIds)
+            )
             ->when(
                 filled($this->filterLevel),
                 fn (Builder $query) => $query->where(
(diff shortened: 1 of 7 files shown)


src/Livewire/Accounting/PaymentReminderRun.php
```diff
@@ -73,6 +81,10 @@ public function loadData(): void
 
         $orders = resolve_static(Order::class, 'query')
             ->wherePaymentReminderDue()
+            ->when(
+                $this->sentOrderIds,
+                fn (Builder $query, array $sentOrderIds) => $query->whereIntegerNotInRaw('id', $sentOrderIds)
```
SirSplasch: whereKeyNot()

SirSplasch (changes requested):

SirSplasch (approved):

PR #1927, opened 2026-07-27
feat(comments): edit your own comments inline

patrickweh:
## What

Comments could be created and deleted but never edited — a long-standing customer request. This adds inline editing of a comment's text, restricted to its author.

## Behaviour (agreed)

- **Who:** only the author may edit the text. Sticky/internal toggles stay open to anyone with the update permission.
- **When:** no time limit; the edit history is captured by the activity log.
- **Marker:** once the text is edited, the comment shows an *edited* label with the timestamp.
- **UX:** inline editor in place (save/cancel), no modal.

## Changes

- `UpdateCommentRuleset` now accepts `comment`; `UpdateComment` guards the text change so only the author can change it (reads the raw `created_by` through the base query, since the accessor resolves to the author's name).
- New `edited_at` column, set from the model `saving` hook only when the comment text actually changes — so a sticky toggle never counts as an edit.
- Comment view: an *edited* marker with the timestamp, plus an inline contenteditable editor with save/cancel. Mentions already embedded in the text are preserved.

## Note

The inline editor is a plain rich-text field, not the full mentions-autocomplete editor used when creating a comment. Existing mentions survive an edit; adding new ones with autocomplete during an edit is out of scope for this first version.

## Tests

- Actions: editing marks the comment edited, sticky toggle does not, editing another user's comment is rejected.
- Livewire: `updateComment` edits the text and returns `edited_at`.

## Summary by Sourcery

Add inline editing of a comment’s text with attribution and history tracking, restricted to the original author.

New Features:
- Allow users to edit their own comments inline in the comment list using a rich-text contenteditable field with save and cancel controls.
- Display an edited label and timestamp on comments whose text has been modified.

Enhancements:
- Record an edited_at timestamp on existing comments when their text changes, without affecting sticky or internal flag updates.
- Validate comment updates so that changes to the text are only allowed for the comment’s author while still permitting permission-based sticky and internal toggles.
- Expose a Livewire action for updating comment text that returns the updated comment payload for UI updates.

Tests:
- Add action tests to ensure text edits set edited_at, sticky toggles do not, and editing another user’s comment is rejected.
- Add Livewire tests to verify updateComment edits the text, returns edited_at, and persists the updated comment.


src/Models/Comment.php
```diff
@@ -33,6 +33,11 @@ protected static function booted(): void
     {
         static::saving(function (Comment $comment): void {
             if ($comment->isDirty('comment')) {
+                // Only an edit of an existing comment marks it; creating one does not.
+                if ($comment->exists) {
+                    $comment->edited_at = now();
```
SirSplasch: You know that you could have just put this line in the static::updating() closure.
Or didn't you do that, because there is non yet, and you didn't want to add it?

src/Models/Comment.php
```diff
@@ -64,6 +69,7 @@ public static function restoring($callback): void
     protected function casts(): array
     {
         return [
+            'edited_at' => 'datetime',
```
SirSplasch: edited_at after is_sticky to match database table column ordering.

src/Actions/Comment/UpdateComment.php
```diff
@@ -41,4 +46,28 @@ protected function prepareForValidation(): void
...
+            // The created_by accessor resolves to the author's name, so read the
+            // raw morph value through the base query instead.
+            $author = resolve_static(Comment::class, 'query')
+                ->whereKey($this->getData('id'))
+                ->toBase()
+                ->value('created_by');
+
+            $currentUser = auth()->user()?->getMorphClass() . ':' . auth()->id();
+
+            if ($author !== $currentUser) {
```
SirSplasch: What if a comment got created via the system or other means, where there is no authenticated user?
These comments can never get edited. Think on how to handle this situation.
Should these created_by = null comments only editable by non authenticated users, or by anyone, or by none?

SirSplasch (changes requested):

SirSplasch (approved):

PR #1928, opened 2026-07-27
Keep media folders created below a collection in place

patrickweh:
## Problem

Creating a subfolder below a collection node in the attachments tree looked like it worked, but the folder was gone from that position after a reload: it showed up at the root of the tree, still named "New folder".

Two causes, both in `FolderTree::saveFolder()`:

- A collection node is derived from media paths, not a record. Its uuid cannot be used as `parent_id`, so the parent was simply dropped and the folder was stored at the root.
- The rename branch for collections was entered whenever the **parent** id was a string. A real folder below a collection has a string parent, so renaming it only rewrote media collection names and never saved the folder.

## Change

- `media_folders.collection_name` (nullable) stores the collection path a folder was created in. It is cleared as soon as the folder has a parent folder, which then defines its position.
- The folder slug is built from that collection path when there is no parent folder, so the whole subtree keeps a consistent path.
- `getMediaAsTree()` places folders with a dotted slug below that path instead of at the root.
- The collection rename branch is keyed on the node itself being a collection.
- `saveFolder()` returns the collection node as the parent again, so the client side tree keeps the folder where the user created it, and no longer drops the children of a renamed folder.

## Tests

`tests/Livewire/FolderTreeTest.php` covers creating and renaming a folder below a collection. Both fail without the change.

## Summary by Sourcery

Keep media folders created under collection nodes anchored to their collection path and correctly represented in the media tree.

New Features:
- Preserve collection context for media folders via a new nullable collection_name field on media_folders and use it to build stable folder slugs and positions in the tree.

Bug Fixes:
- Ensure new folders created under a collection node remain under that collection instead of moving to the root after reloads.
- Fix renaming of folders under collection nodes so their names and slugs update correctly without losing their position.
- Render real media folders under virtual collection nodes in the media tree based on dotted slugs rather than treating them as root nodes.

Enhancements:
- Refine media tree construction to distinguish between virtual collection nodes and persisted folder nodes when resolving dotted slugs.
- Centralize media folder slug generation logic via a reusable buildSlug method on MediaFolder.

Build:
- Add database migration to introduce the collection_name column to the media_folders table and wire it through validation and forms.

Tests:
- Add Livewire tests covering creating and renaming media folders under a collection node and extend test permissions to include folder creation.

### DIFF
--- a/database/migrations/2026_07_27_000001_add_collection_name_to_media_folders_table.php
+++ b/database/migrations/2026_07_27_000001_add_collection_name_to_media_folders_table.php
@@ -1,0 +1,24 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class() extends Migration
+{
+    public function up(): void
+    {
+        Schema::table('media_folders', function (Blueprint $table): void {
+            $table->string('collection_name')
+                ->nullable()
+                ->after('parent_id');
+        });
+    }
+
+    public function down(): void
+    {
+        Schema::table('media_folders', function (Blueprint $table): void {
+            $table->dropColumn('collection_name');
+        });
+    }
+};
--- a/resources/views/livewire/folder-tree.blade.php
+++ b/resources/views/livewire/folder-tree.blade.php
@@ -209,6 +209,7 @@
                                             $wire
                                                 .saveFolder({
                                                     parent_id: selection.id,
+                                                    collection_name: getNodePath(selectionProxy, 'slug'),
                                                     name: '{{ __('New folder') }}',
                                                     is_new: true,
                                                     children: [],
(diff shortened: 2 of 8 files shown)


patrickweh (commented):
The fix itself reads well: buildSlug() pulls together logic that was copied three times, the saving hook keeps collection_name and parent_id from contradicting each other, and the ruleset and form are carried along.

What is missing is a backfill, and it decides whether this actually closes the ticket it came from. The migration only adds the column. On the VVO instance there are already **102 folders named "Neuer Ordner", 50 of them stranded at the top level** from exactly this bug. After the merge they stay where they are: new folders land correctly, the existing ones do not move.

Worse, they cannot be repaired later from within the schema. collection_name is null for all of them and the slug carries no trace of where they were meant to sit, so once this ships the information is gone for good. If a backfill is possible at all it has to be written now, while the wrong placements are still recognisable.

Two smaller things: buildSlug() is public although only booted() calls it, and this PR touches FolderTree.php and folder-tree.blade.php, as does #2024, so whichever merges second needs a rebase.

SirSplasch (approved):

PR #1934, opened 2026-07-28
Refuse to delete an order type that still has orders

patrickweh:
## Problem

Deleting an order type leaves every order of that type without the type it was created with. The order keeps its `order_type_id`, but the type itself is hidden from every query by the soft delete scope, so `Order::orderType()` resolves to `null`.

Reading such an order then fails outright:

```
Error: Call to a member function isPurchase() on null
  src/Livewire/Forms/OrderForm.php:245
```

`OrderForm::fill()` dereferences the type to decide whether the order is a purchase, and hits a null. The order detail page answers 500 and the record is unreachable — not just degraded.

Nothing prevented this. `DeleteOrderTypeRuleset` only checks that the type exists, so a type in daily use can be removed with a single click.

## Change

Refuse the delete while orders reference the type, the way `DeleteLanguage` already refuses a language that addresses or users point at — same `validateData()` override, same `423` status.

Merging the type into another one (`MergeRecords`, exposed through the `RecordMerging` component) moves the orders across first and then leaves the type free to delete. That is the path for the two real reasons to remove a type: it exists twice, or it is retired and its orders belong somewhere else.

Trashed orders count too — they can be restored, and would come back pointing at a type that is gone for good.

## Why not `withTrashed()` on the relation

Letting the order read its deleted type would paper over the 500, but a deleted order type is deliberately absent everywhere else — it is not offered in menus or selects. Making it reappear only for orders splits the meaning of the delete. The orders should not be left dangling in the first place.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by Sourcery

Prevent deletion of order types that are still referenced by orders to avoid leaving dangling orders without a valid type.

Bug Fixes:
- Reject deleting an order type when any (including soft-deleted) orders reference it, returning a 423 validation error instead of allowing broken orders.

Tests:
- Add feature tests asserting that order types in use by orders cannot be deleted and that unused order types remain deletable.

### DIFF
--- a/src/Actions/OrderType/DeleteOrderType.php
+++ b/src/Actions/OrderType/DeleteOrderType.php
@@ -5,6 +5,7 @@ namespace FluxErp\Actions\OrderType;
 use FluxErp\Actions\FluxAction;
 use FluxErp\Models\OrderType;
 use FluxErp\Rulesets\OrderType\DeleteOrderTypeRuleset;
+use Illuminate\Validation\ValidationException;
 
 class DeleteOrderType extends FluxAction
 {
@@ -24,5 +25,26 @@ class DeleteOrderType extends FluxAction
             ->whereKey($this->data['id'])
             ->first()
             ->delete();
+    }
+
+    protected function validateData(): void
+    {
+        parent::validateData();
+
+        $orderType = resolve_static(OrderType::class, 'query')
+            ->whereKey($this->data['id'])
+            ->first();
+
+        // An order keeps pointing at its type for the rest of its life, and the
+        // type is hidden from every query once it is deleted. The order would
+        // lose the type it was created with, so the type has to be merged into
+        // another one first.
+        if ($orderType->orders()->withTrashed()->exists()) {
+            throw ValidationException::withMessages([
+                'order' => ['Order type referenced by an order'],
+            ])
+                ->errorBag('deleteOrderType')
+                ->status(423);
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


src/Actions/OrderType/DeleteOrderType.php
```diff
@@ -25,4 +26,25 @@ public function performAction(): ?bool
...
+
+        $orderType = resolve_static(OrderType::class, 'query')
+            ->whereKey($this->data['id'])
+            ->first();
+
+        // An order keeps pointing at its type for the rest of its life, and the
+        // type is hidden from every query once it is deleted. The order would
+        // lose the type it was created with, so the type has to be merged into
+        // another one first.
+        if ($orderType->orders()->withTrashed()->exists()) {
```
SirSplasch: Do you really need to include soft-deleted orders in this query? The order type itself also is soft deleted. So if an order of a soft-deleted order type gets restored, the deleted order type must also be restored.

SirSplasch (changes requested):

SirSplasch (approved):

src/Actions/OrderType/DeleteOrderType.php
patrickweh: Agreed on dropping the trashed orders, and I checked the reasoning holds today: there is no `RestoreOrder` action at all, only Contact and Product have one, so a deleted order has no supported way back to find its type missing.

One thing worth knowing though: nothing currently implements "restoring such an order restores its type". `Order` declares `$cascadeDeletes` but no `$cascadeRestores`, and `CascadeSoftDeletes` cascades down to the declared child relations, never up to a parent like `orderType`.

So the invariant in the comment is a note for whoever adds a restore path later rather than something the code guarantees. Fine to leave as is while orders cannot be restored — just flagging it so the 500 does not come back through that door.

PR #1943, opened 2026-07-29
fix(addresses): drop the roles of a deleted address

patrickweh:
## Problem

A role of a contact (main, delivery, invoice) sits on exactly one address. An address never gives a role up by being switched off, the role moves when another address takes it over, which is why each role toggle is disabled once it is on.

`Address::deleted` already hands the roles of a deleted address to the address that survives it, and repoints the contact. The deleted row kept its own flags though. Restoring that address left the contact with two rows claiming the same role, and neither could be corrected afterwards: taking a role means turning it on somewhere else, and both were already on.

Deleting a whole contact hits the same path, because `CascadeSoftDeletes` deletes its addresses one by one through this hook. Restoring the contact then showed two main addresses, while the contact pointed at the one that had been promoted last.

## Change

The deleted address now gives up exactly the roles it handed over, so a restore always yields a single holder per role, and switching a role back works the normal way.

## Tests

`tests/Feature/Models/AddressRolesTest.php` covers the handover itself, restoring a single address, and restoring a whole contact. All three fail without the change. The Address and Contact tests of the Feature and Livewire suites pass.

## Summary by Sourcery

Ensure address roles are uniquely reassigned when an address or contact is soft-deleted and later restored.

Bug Fixes:
- Clear role flags on a deleted address after its roles are handed over so that no two addresses claim the same role after restore.

Tests:
- Add feature tests verifying role handover on address deletion and ensuring only a single main/delivery address exists after restoring an address or an entire contact.

### DIFF
--- a/src/Models/Address.php
+++ b/src/Models/Address.php
@@ -280,16 +280,7 @@ class Address extends FluxAuthenticatable implements Calendarable, HasLocalePref
 
                 $mainAddress->update($addressUpdates);
 
-                // A role belongs to exactly one address, and the address that
-                // just handed it over has to give up its own claim on it.
-                // Keeping the flag left the role on two rows as soon as the
-                // deleted address was restored, and neither of them could be
-                // switched over any more, because a role is taken by turning it
-                // on somewhere else and both were already on.
-                resolve_static(Address::class, 'query')
-                    ->withTrashed()
-                    ->whereKey($address->getKey())
-                    ->update(array_fill_keys(array_keys($addressUpdates), false));
+                $address->updateQuietly(array_fill_keys(array_keys($addressUpdates), false));
             }
         });
     }
--- a/src/Models/Address.php
+++ b/src/Models/Address.php
@@ -279,6 +279,17 @@ class Address extends FluxAuthenticatable implements Calendarable, HasLocalePref
                     ->update($contactUpdates);
 
                 $mainAddress->update($addressUpdates);
+
+                // A role belongs to exactly one address, and the address that
+                // just handed it over has to give up its own claim on it.
+                // Keeping the flag left the role on two rows as soon as the
+                // deleted address was restored, and neither of them could be
+                // switched over any more, because a role is taken by turning it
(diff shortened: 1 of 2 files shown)


src/Models/Address.php
```diff
@@ -279,6 +279,17 @@ protected static function booted(): void
...
 
                 $mainAddress->update($addressUpdates);
+
+                // A role belongs to exactly one address, and the address that
+                // just handed it over has to give up its own claim on it.
+                // Keeping the flag left the role on two rows as soon as the
+                // deleted address was restored, and neither of them could be
+                // switched over any more, because a role is taken by turning it
+                // on somewhere else and both were already on.
+                resolve_static(Address::class, 'query')
```
SirSplasch: Do you need to re-query the address? Can't you just updateQuietly() $address?

SirSplasch (changes requested):

src/Models/Address.php
patrickweh: Right, updateQuietly() does the same job without the second query and keeps the instance in sync instead of leaving it stale behind an update on the query builder. Changed in 618732ac9.

src/Models/Address.php
```diff
@@ -279,6 +279,17 @@ protected static function booted(): void
...
+                    ->withTrashed()
+                    ->whereKey($address->getKey())
+                    ->update(array_fill_keys(array_keys($addressUpdates), false));
```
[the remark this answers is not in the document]
patrickweh: Keeping this as is. The coupling to $addressUpdates is the point rather than an accident: the address has to give up exactly the roles it just handed over, and that array is built ten lines above from nothing but the three role flags. A fixed list would clear roles that were never transferred.

The suggested columns (billing, shipping, primary) do not exist on this model either, the flags are is_main_address, is_delivery_address and is_invoice_address.

If $addressUpdates ever grows non-role fields, that same change has to revisit this block anyway, since it decides what the surviving address takes over.

SirSplasch (approved):

PR #1945, opened 2026-07-30
Let ledger bookings settle the order they point at

patrickweh:
## Problem

An invoice can be settled without any bank transaction ever touching our own accounts. The case that surfaced this: a car was financed and the bank paid the dealer directly, so the purchase invoice in Flux had no payment to assign. The correct entry is a ledger-to-ledger booking (supplier liability against the loan account), and `LedgerBooking` already carries a polymorphic `source`.

The order kept its full balance and stayed open anyway, because `Order::totalPaid()` only summed assigned transactions:

```php
return $this->transactions()
    ->withPivot(['amount', 'order_currency_amount'])
    ->get()
    ->reduce(...);
```

So the books were right while the order list still claimed the invoice had to be paid. Setting the balance by hand does not help, since `calculateBalance()` recomputes it from `total_gross_price` minus `totalPaid()` on every touch.

## Change

- `Order::ledgerBookings()` exposes the bookings that point at the order.
- `Order::totalPaid()` adds `totalSettledByLedgerBookings()`. Booking amounts are always positive, so the order type multiplier decides the direction a booking settles in. A purchase invoice is settled by a negative contribution, a sales invoice by a positive one.
- `calculatePaymentState()` no longer forces an order back to Open when it has no transactions but is settled by a booking.
- `LedgerBooking` recalculates the payment state of the affected orders on save and delete, including the previous order when a booking is moved to another source.
- `source_type` and `source_id` are accepted by the create and update rulesets, so the reference can be set through the action rather than only in code. Both are validated with `MorphClassExists` and `MorphExists` and required together.

## Tests

`tests/Feature/Actions/LedgerBooking/LedgerBookingSettlesOrderTest.php` covers a settled purchase order, a settled sales order, a partial booking that leaves the remainder open, a deleted booking that reopens the order, and a booking without a source that leaves orders untouched.

```
Tests:    11 passed (19 assertions)   tests/Feature/Actions/LedgerBooking
Tests:    89 passed (208 assertions)  Order, OrderTransaction, Transaction, Livewire/Accounting
```

Pint clean on the touched files.

## Notes

The schedule generator still cannot express a final balloon installment, and `Loan` has no media support for the credit agreement. Both came up in the same session and are left for separate pull requests.

## Summary by Sourcery

Allow ledger-to-ledger bookings to settle the payment state and balance of the orders they reference, alongside bank transactions.

New Features:
- Expose a ledgerBookings relation on orders so ledger bookings can be associated directly with an order.
- Include amounts from ledger bookings in an order’s totalPaid calculation based on the order type’s multiplier.
- Support setting a ledger booking’s polymorphic source (including orders) via create and update actions with validation for morph type and existence.

Bug Fixes:
- Prevent settled orders from being reset to Open when they have no transactions but are fully settled via ledger bookings.
- Ensure saving, deleting, or re-pointing a ledger booking triggers recalculation of the payment state of affected orders.

Tests:
- Add feature tests covering order settlement via ledger bookings for purchase and sales orders, partial settlement, reopening on booking deletion, and bookings without a source leaving orders unchanged.

### DIFF
--- a/src/Models/LedgerBooking.php
+++ b/src/Models/LedgerBooking.php
@@ -11,10 +11,23 @@ use FluxErp\Traits\Model\HasUuid;
 use FluxErp\Traits\Model\SoftDeletes;
 use Illuminate\Database\Eloquent\Relations\BelongsTo;
 use Illuminate\Database\Eloquent\Relations\MorphTo;
+use Illuminate\Support\Collection;
 
 class LedgerBooking extends FluxModel
 {
     use Filterable, HasPackageFactory, HasTenantAssignment, HasUserModification, HasUuid, SoftDeletes;
+
+    protected static function booted(): void
+    {
+        $recalculateSettledOrders = function (LedgerBooking $ledgerBooking): void {
+            $ledgerBooking->settledOrders()->each(
+                fn (Order $order) => $order->calculatePaymentState()->save()
+            );
+        };
+
+        static::saved($recalculateSettledOrders);
+        static::deleted($recalculateSettledOrders);
+    }
 
     protected function casts(): array
     {
@@ -38,5 +51,21 @@ class LedgerBooking extends FluxModel
     public function source(): MorphTo
     {
         return $this->morphTo();
+    }
+
+    /**
+     * The orders whose payment state depends on this booking: the one it points at
+     * now and, when the source was moved, the one it pointed at before.
+     */
+    public function settledOrders(): Collection
+    {
+        return collect([
+            [$this->getOriginal('source_type'), $this->getOriginal('source_id')],
+            [$this->source_type, $this->source_id],
+        ])
+            ->filter(fn (array $source) => $source[1] && is_a(morphed_model($source[0]) ?? '', Order::class, true))
+            ->unique(fn (array $source) => $source[1])
+            ->map(fn (array $source) => resolve_static(Order::class, 'query')->whereKey($source[1])->first())
+            ->filter();
     }
 }
--- a/src/Models/LedgerBooking.php
+++ b/src/Models/LedgerBooking.php
@@ -11,7 +11,6 @@ use FluxErp\Traits\Model\HasUuid;
 use FluxErp\Traits\Model\SoftDeletes;
 use Illuminate\Database\Eloquent\Relations\BelongsTo;
 use Illuminate\Database\Eloquent\Relations\MorphTo;
-use Illuminate\Support\Collection;
 
 class LedgerBooking extends FluxModel
 {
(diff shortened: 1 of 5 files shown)


src/Rulesets/LedgerBooking/UpdateLedgerBookingRuleset.php
```diff
@@ -41,6 +43,18 @@ public function rules(): array
...
+                'nullable',
+                'required_with:source_id',
+                app(MorphClassExists::class),
+            ],
+            'source_id' => [
+                'integer',
+                'nullable',
+                'required_with:source_type',
+                app(MorphExists::class, ['modelAttribute' => 'source_type']),
+            ],
```
SirSplasch, line 49: order these two according to the table's column ordering.

src/Models/LedgerBooking.php
```diff
@@ -11,11 +11,24 @@
...
+    protected static function booted(): void
+    {
+        $recalculateSettledOrders = function (LedgerBooking $ledgerBooking): void {
+            $ledgerBooking->settledOrders()->each(
+                fn (Order $order) => $order->calculatePaymentState()->save()
+            );
+        };
+
+        static::saved($recalculateSettledOrders);
+        static::deleted($recalculateSettledOrders);
```
SirSplasch: I know that you want to reduce duplicate code, but this must change if something gets added to the saved callback for example. So why not simply handle this like every other booted method of a model and implement the callbacks for saved and deleted separately.

src/Models/LedgerBooking.php
```diff
@@ -39,4 +52,20 @@ public function source(): MorphTo
...
+     * The orders whose payment state depends on this booking: the one it points at
+     * now and, when the source was moved, the one it pointed at before.
+     */
+    public function settledOrders(): Collection
+    {
+        return collect([
+            [$this->getOriginal('source_type'), $this->getOriginal('source_id')],
+            [$this->source_type, $this->source_id],
+        ])
+            ->filter(fn (array $source) => $source[1] && is_a(morphed_model($source[0]) ?? '', Order::class, true))
```
SirSplasch: $source[0] === morph_alias(Order::class) instead of the is_a() check.
Also you are using the Order::class directly without taking customization into account.

src/Models/LedgerBooking.php
SirSplasch: In the deleted closure the source morph cannot change, as the model will only be deleted, not updated.

src/Models/LedgerBooking.php
```diff
@@ -11,11 +11,24 @@
...
 
+    protected static function booted(): void
+    {
+        $recalculateSettledOrders = function (LedgerBooking $ledgerBooking): void {
+            $ledgerBooking->settledOrders()->each(
+                fn (Order $order) => $order->calculatePaymentState()->save()
+            );
+        };
+
+        static::saved($recalculateSettledOrders);
```
SirSplasch: In the saved closure, check if the source morph was changed, and then handle the recalculation(s) accordingly.

src/Models/LedgerBooking.php
```diff
@@ -39,4 +52,20 @@ public function source(): MorphTo
     {
         return $this->morphTo();
     }
+
+    /**
+     * The orders whose payment state depends on this booking: the one it points at
+     * now and, when the source was moved, the one it pointed at before.
+     */
+    public function settledOrders(): Collection
```
SirSplasch: I don't think this method is needed, as the deleted closure does not need to call this method.

SirSplasch (changes requested):

SirSplasch (approved):

PR #1948, opened 2026-08-01
Normalize translated salutation labels in the order address snapshot

patrickweh:
## Problem

A monthly subscription order stopped producing its invoice. `ProcessSubscriptionOrder` failed every run with a validation error that only said:

```
Der ausgewählte Wert ist ungültig.   →   address_delivery.salutation
```

`orders.address_invoice` and `orders.address_delivery` are plain JSON snapshots cast to `array`. Nothing stops a caller from putting the translated label into them, and in a German instance the snapshot held `"Firma"` instead of `"company"`. `ReplicateOrder` builds its payload from `$originalOrder->toArray()` and the ruleset validates the snapshot with `EnumRule(SalutationEnum)`, so the label is rejected and every replication of that order fails.

It stays invisible until an order is replicated or saved through an action, which is why it surfaced on a subscription and not on the thousands of legacy orders carrying the same value.

## Change

`Order` normalizes the salutation inside both snapshots in its `saving` hook:

```php
foreach (['address_invoice', 'address_delivery'] as $addressSnapshot) {
    $order->{$addressSnapshot} = static::normalizeSnapshotSalutation($order->{$addressSnapshot});
}
```

`normalizeSnapshotSalutation()` returns the snapshot untouched when the salutation is empty or already a valid case. Otherwise it looks for a case whose label matches, checking both `__(Str::headline($case->value))` and `__($case->value)`, since JSON translation files key the plain value. Anything it cannot map stays as it is, so unknown input is never silently dropped.

Legacy rows heal on their next save, no migration required.

## Tests

`tests/Feature/Models/OrderAddressSnapshotSalutationTest.php`: a translated label is normalized in both snapshots while the rest of the address is preserved, an enum value is left alone, an unknown salutation is left alone, a snapshot without a salutation is untouched, and assigning a label to an existing order normalizes it on save.

```
Tests:    5 passed (7 assertions)    OrderAddressSnapshotSalutationTest
Tests:    43 passed (85 assertions)  tests/Feature/Actions/Order, tests/Feature/Models
```

Pint clean on the touched files.

## Summary by Sourcery

Normalize salutation values in order address snapshots to valid enum cases on save to prevent replication and validation failures caused by translated labels.

Enhancements:
- Add snapshot salutation normalization in the Order model saving hook for both invoice and delivery addresses, mapping known translated labels back to their corresponding SalutationEnum values while leaving unknown values unchanged.

Tests:
- Add feature tests ensuring address snapshot salutations are normalized from translated labels, left untouched when already valid or unknown, and updated correctly when assigned directly on existing orders.

### DIFF
--- a/src/Models/Order.php
+++ b/src/Models/Order.php
@@ -148,7 +148,17 @@ class Order extends FluxModel implements Calendarable, HasMedia, InteractsWithDa
             }
 
             foreach (['address_invoice', 'address_delivery'] as $addressSnapshot) {
-                $order->{$addressSnapshot} = static::normalizeSnapshotSalutation($order->{$addressSnapshot});
+                $address = $order->{$addressSnapshot};
+
+                if (! is_array($address)) {
+                    continue;
+                }
+
+                $normalized = static::normalizeSnapshotSalutation($address);
+
+                if ($normalized !== $address) {
+                    $order->{$addressSnapshot} = $normalized;
+                }
             }
 
             // reset to original
@@ -447,12 +457,8 @@ class Order extends FluxModel implements Calendarable, HasMedia, InteractsWithDa
         ];
     }
 
-    protected static function normalizeSnapshotSalutation(mixed $address): mixed
+    protected static function normalizeSnapshotSalutation(array $address): array
     {
-        if (! is_array($address)) {
-            return $address;
-        }
-
         $salutation = data_get($address, 'salutation');
 
         if (! $salutation || resolve_static(SalutationEnum::class, 'tryFrom', ['value' => $salutation])) {
--- a/src/Models/Order.php
+++ b/src/Models/Order.php
@@ -447,10 +447,6 @@ class Order extends FluxModel implements Calendarable, HasMedia, InteractsWithDa
         ];
     }
 
-    /**
-     * A snapshot holding the translated label instead of the enum value fails
-     * validation on every replication of the order.
-     */
     protected static function normalizeSnapshotSalutation(mixed $address): mixed
     {
         if (! is_array($address)) {
--- a/src/Models/Order.php
+++ b/src/Models/Order.php
@@ -16,6 +16,7 @@ use FluxErp\Contracts\OffersPrinting;
 use FluxErp\Contracts\Targetable;
 use FluxErp\Enums\OrderTypeEnum;
 use FluxErp\Enums\PaymentRunTypeEnum;
+use FluxErp\Enums\SalutationEnum;
 use FluxErp\Events\Order\OrderApprovalRequestEvent;
 use FluxErp\Models\Pivots\AddressAddressTypeOrder;
 use FluxErp\Models\Pivots\OrderPaymentRun;
@@ -76,6 +77,7 @@ use Illuminate\Support\Collection as SupportCollection;
 use Illuminate\Support\Facades\DB;
 use Illuminate\Support\Facades\Validator;
 use Illuminate\Support\Number;
+use Illuminate\Support\Str;
 use Illuminate\Support\Traits\Conditionable;
 use RoundingMode;
 use Spatie\MediaLibrary\HasMedia;
@@ -143,6 +145,10 @@ class Order extends FluxModel implements Calendarable, HasMedia, InteractsWithDa
                 && ! $order->address_delivery
             ) {
                 $order->address_delivery = $order->addressDelivery()->with('country')->first();
+            }
+
+            foreach (['address_invoice', 'address_delivery'] as $addressSnapshot) {
+                $order->{$addressSnapshot} = static::normalizeSnapshotSalutation($order->{$addressSnapshot});
             }
 
             // reset to original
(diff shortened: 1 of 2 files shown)


src/Models/Order.php
```diff
@@ -441,6 +447,32 @@ public static function toCalendar(): array
...
+     * The address snapshots are plain JSON, so nothing stops a caller from storing the
+     * translated salutation label instead of the enum value. Replicating such an order
+     * later fails validation, because the ruleset only accepts the enum value. Map a
+     * known label back to its case and leave anything unknown untouched.
+     */
+    protected static function normalizeSnapshotSalutation(?array $address): ?array
+    {
+        $salutation = data_get($address, 'salutation');
+
+        if (! $salutation || resolve_static(SalutationEnum::class, 'tryFrom', ['value' => $salutation])) {
```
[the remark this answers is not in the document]
patrickweh: Keeping `resolve_static()` here. It is the established idiom in this codebase for calling static enum APIs, and it is not just indirection: it resolves the concrete class through `App::getBindings()`, so an application can bind its own `SalutationEnum` with additional cases and this code picks it up. A direct static call would bypass that.

Existing usages follow the same shape, including the named parameter array:

- `src/Traits/Livewire/Widget/IsTimeFrameAwareWidget.php:116` - `resolve_static(TimeFrameEnum::class, 'getRange', ['case' => $this->timeFrame])`
- `src/Livewire/Settings/PriceLists.php:247` - `resolve_static(RoundingMethodEnum::class, 'valuesLocalized')`
- `src/Support/Enums/Traits/IsCastable.php` - `resolve_static(static::class, 'tryFrom', ['value' => $value])`

`resolve_static()` spreads the array, so the named key matches the `tryFrom(int|string|null $value)` signature.

patrickweh:
Went through the review.

**Non-array snapshots** - valid, fixed in 2072cfa. A legacy column can decode to something other than an array, and the `?array` type hint would have thrown before the method could bail out. It now takes `mixed` and returns anything that is not an array untouched, covered by a test.

**Caching the enum cases** - not needed. `FluxEnum::cases()` already memoizes through `Cache::memo()->rememberForever()`, and the lookup only runs when a salutation is present and is not already a valid case. Every normal save short-circuits on the first condition.

**`resolve_static()` for `tryFrom`** - keeping it, reasoning in the inline reply: it resolves container bindings so an application can swap the enum, and the same shape is used across the codebase.

Also dropped two comments in the test that only restated what the code does.

src/Models/Order.php
```diff
@@ -441,6 +447,30 @@ public static function toCalendar(): array
         ];
     }
 
+    protected static function normalizeSnapshotSalutation(mixed $address): mixed
```
SirSplasch: Isn't the datatype of $address ?array

src/Models/Order.php
```diff
@@ -145,6 +147,10 @@ protected static function booted(): void
                 $order->address_delivery = $order->addressDelivery()->with('country')->first();
             }
 
+            foreach (['address_invoice', 'address_delivery'] as $addressSnapshot) {
```
SirSplasch, line 150: Please check if these keys are dirty. No need to always run this. Also if only one of them is dirty, no need to do anything with the other key that has not changed.

src/Models/Order.php
SirSplasch: return type is probably also ?array

SirSplasch (changes requested):

src/Models/Order.php
patrickweh: Changed, but to `array` rather than `?array`.

The cast is `array`, so the getter hands back an array or null, never a scalar for any row we have: of 11720 orders, 0 have a null `address_invoice`, 81 have a null `address_delivery`, and exactly one holds JSON that is not an object (a plain list of address lines from an import). A scalar is still possible in principle, which is what the `mixed` was defending against.

That defence sat in the wrong place. The caller now skips anything that is not an array, so null and any scalar never reach the method and it can state exactly what it takes:

```php
$address = $order->{$addressSnapshot};

if (! is_array($address)) {
    continue;
}
```

The list shaped row still goes through and comes back untouched, since `data_get()` finds no salutation in it.

patrickweh: Return type is `array` now as well, for the same reason as the parameter: the caller filters out everything that is not an array, so the method never has a null to hand back.

src/Models/Order.php
patrickweh: Half of this is done, the other half would take the fix out.

Only writing back what changed is right, so the snapshot is no longer reassigned on every save:

```php
$normalized = static::normalizeSnapshotSalutation($address);

if ($normalized !== $address) {
    $order->{$addressSnapshot} = $normalized;
}
```

Guarding the whole block with `isDirty()` does not work, though. The rows this fixes were written long ago and carry the translated label in the database. On a normal save, of an invoice date or a state, that column is not dirty, and it never becomes dirty on its own. Skipping it means such an order is never repaired and the next replication fails exactly as before. The healing depends on reading the snapshot on a save that did not touch it.

The tests show it: with `isDirty()` in front of the loop, `a translated salutation label in the snapshot is normalized to the enum value` fails, because it writes the label through the query builder and then saves the model, which is precisely the legacy case.

What is left per save is a `data_get()` on an already decoded array plus a comparison, and no write. If you would rather have the strict `isDirty()` guard, then the legacy rows need a one off command to clean them up, and I would put that in a separate PR.

PR #1951, opened 2026-08-03
Keep translated salutations out of the order address snapshots

patrickweh:
The invoice and the delivery address are stored as plain JSON on the order, and a caller was free to write the translated label into them. In a German instance that meant `Firma` where the enum expects `company`, and since the ruleset validates the salutation as an enum, every replication of such an order failed. A monthly subscription stopped producing its invoice that way.

Only the delivery address was ever validated field by field:

```php
Arr::prependKeysWith(
    resolve_static(PostalAddressRuleset::class, "getRules"),
    "address_delivery."
),
```

The invoice address had nothing but `array` and `nullable`, so the label went in unnoticed. Both are validated the same way now, and an order that carries a label is refused where it is written rather than where it is read.

The rows already carrying one are cleaned up once by a migration. It maps a label back to its case through the translations of every locale the installation has, the app lang directory, the package lang directory, `app.locale` and `fallback_locale`, so an instance in another language is repaired as well. It works on `DB` in chunks of 500 and writes only where something actually changes. Anything it cannot map stays as it is.

On our own live instance this concerns 199 rows in `address_invoice` and 195 in `address_delivery`, values `Firma`, `Herr` and `Frau`. The older ones come from the FileMaker import, the ones from 2025 and 2026 do not, while the address records themselves hold the enum value throughout.

One consequence worth knowing: a caller that sends a label now fails validation instead of storing it silently.

## Summary by Sourcery

Validate order invoice and delivery address snapshots with postal address rules and normalize existing salutation values from translated labels to enum values.

Bug Fixes:
- Prevent invalid translated salutations from being stored in order invoice and delivery address snapshots by enforcing postal address validation.
- Repair existing orders whose address snapshots contain translated salutation labels by mapping them back to the correct enum values.

Tests:
- Add feature tests covering salutation normalization in order address snapshots and validation of invoice and delivery address salutations.

### DIFF
--- a/database/migrations/2026_08_03_000002_normalize_salutation_in_order_address_snapshots.php
+++ b/database/migrations/2026_08_03_000002_normalize_salutation_in_order_address_snapshots.php
@@ -1,0 +1,90 @@
+<?php
+
+use FluxErp\Enums\SalutationEnum;
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Support\Collection;
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Str;
+
+return new class() extends Migration
+{
+    public function up(): void
+    {
+        $labels = $this->labels();
+
+        if (! $labels) {
+            return;
+        }
+
+        DB::table('orders')
+            ->select(['id', 'address_invoice', 'address_delivery'])
+            ->orderBy('id')
+            ->chunk(500, function (Collection $orders) use ($labels): void {
+                foreach ($orders as $order) {
+                    $changes = [];
+
+                    foreach (['address_invoice', 'address_delivery'] as $column) {
+                        $address = json_decode($order->{$column} ?? '', true);
+                        $salutation = data_get($address, 'salutation');
+
+                        if (! is_array($address)
+                            || ! is_string($salutation)
+                            || ! array_key_exists($salutation, $labels)
+                        ) {
+                            continue;
+                        }
+
+                        $address['salutation'] = $labels[$salutation];
+                        $changes[$column] = json_encode($address);
+                    }
+
+                    if ($changes) {
+                        DB::table('orders')
+                            ->where('id', $order->id)
+                            ->update($changes);
+                    }
+                }
+            });
+    }
+
+    public function down(): void
+    {
+        //
+    }
+
+    protected function labels(): array
+    {
+        $values = array_map(
+            fn (object $case): string => $case->value,
+            resolve_static(SalutationEnum::class, 'cases')
+        );
+
+        $labels = [];
+
+        foreach ($this->locales() as $locale) {
+            foreach ($values as $value) {
+                foreach ([Str::headline($value), $value] as $key) {
+                    $label = trans($key, [], $locale);
+
+                    if ($label !== $key && ! in_array($label, $values, true)) {
+                        $labels[$label] = $value;
+                    }
+                }
+            }
+        }
+
+        return $labels;
+    }
+
+    protected function locales(): array
+    {
+        return collect(glob(lang_path('*.json')))
+            ->merge(glob(__DIR__ . '/../../lang/*.json'))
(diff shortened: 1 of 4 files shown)


database/migrations/2026_08_03_000002_normalize_salutation_in_order_address_snapshots.php
```diff
@@ -0,0 +1,90 @@
...
+    public function up(): void
+    {
+        $labels = $this->labels();
+
+        if (! $labels) {
+            return;
+        }
+
+        DB::table('orders')
+            ->select(['id', 'address_invoice', 'address_delivery'])
```
SirSplasch, line 20: Maybe add
->whereNotNull('address_invoice')
->orWhereNotNull('address_delivery')

database/migrations/2026_08_03_000002_normalize_salutation_in_order_address_snapshots.php
```diff
@@ -0,0 +1,90 @@
...
+        $labels = [];
+
+        foreach ($this->locales() as $locale) {
+            foreach ($values as $value) {
+                foreach ([Str::headline($value), $value] as $key) {
+                    $label = trans($key, [], $locale);
+
+                    if ($label !== $key && ! in_array($label, $values, true)) {
+                        $labels[$label] = $value;
+                    }
```
SirSplasch: Doesn't that bring some issues? Like what if for example a key is not localized in English, and therefore $label == $key. Then it will not appear in the $labels array.
Also what if for example different keys translate to the same $label, but in different languages. Then only the first $label will appear, and the others will be ignored.

To me this looks flawed.

database/migrations/2026_08_03_000002_normalize_salutation_in_order_address_snapshots.php
```diff
@@ -0,0 +1,90 @@
...
+                    if ($changes) {
+                        DB::table('orders')
+                            ->where('id', $order->id)
+                            ->update($changes);
+                    }
+                }
+            });
+    }
+
+    public function down(): void
```
SirSplasch: public function down(): void {}
no need for the empty comment line.

SirSplasch (changes requested):

database/migrations/2026_08_03_000002_normalize_salutation_in_order_address_snapshots.php
patrickweh: Added. On our own live instance it saves nothing, no order has both columns null, 0 for `address_invoice` and 81 for `address_delivery` out of 11720. On an installation that rarely fills them it does.

database/migrations/2026_08_03_000002_normalize_salutation_in_order_address_snapshots.php
patrickweh: Both hold, and both are fixed.

The untranslated key was silently dropped. `Str::headline("no_salutation")` gives `No Salutation`, English has no line for it, so `$label === $key` and the entry never entered the map, while that string is exactly what an English instance shows in its select. The key is a candidate on its own now, translated or not.

The collision was worse than being ignored, it was last one wins, so a label that two cases share would have rewritten data to whichever locale came last. Candidates are collected per label now and only unambiguous ones survive, everything else stays as it is.

```php
foreach ($values as $value) {
    foreach ([Str::headline($value), $value] as $key) {
        $candidates[$key][] = $value;

        foreach ($locales as $locale) {
            $candidates[trans($key, [], $locale)][] = $value;
        }
    }
}

return collect($candidates)
    ->except($values)
    ->map(fn (array $mapped): array => array_unique($mapped))
    ->filter(fn (array $mapped): bool => count($mapped) === 1)
    ->map(fn (array $mapped): string => reset($mapped))
    ->all();
```

Two tests came with it, one for each case. Both fail on the previous version.

SirSplasch (approved):

PR #1955, opened 2026-08-03
Report payment reminders that could not be sent

patrickweh:
## Problem

A payment reminder run reported success even when not a single reminder was mailed.

`SendPaymentReminderJob` drops an invoice silently when

- no `PaymentReminderText` exists for the reminder level, or
- the reminder text has no email template linked (`email_template_id` is nullable).

In both cases the job writes an activity log entry and returns normally. The monitored batch therefore finishes without failures, `PaymentReminderRun` hides the invoices as sent, and the user sees a green "all jobs succeeded" toast. Delivery failures behaved the same way: `SendMail` returns `success => false`, the job cleans up and returns, and the batch still reports success.

## Change

- `BundlePaymentReminders` resolves the reminder texts once and checks each invoice's next reminder level before dispatching. Unconfigured invoices are not queued and are returned as `unsendable` with `id`, `invoice_number`, `reminder_level` and a translated reason. Configured invoices in the same call are still sent.
- `PaymentReminderRun` marks only the queued invoices as sent, so the held back ones stay in the list, and raises a warning toast grouping the reasons.
- `SendPaymentReminderJob::abort()` now fails the batch job after cleaning up, so the finished-batch toast turns red on delivery failures instead of green.

## Tests

- `BundlePaymentRemindersTest`: reports missing reminder text, reports missing email template, and queues configured invoices while reporting only the unconfigured ones.
- `SendPaymentReminderJobTest`: fails the job when the mail cannot be sent.
- `PaymentReminderRunTest`: invoices that cannot be sent stay in the list and no job is queued.

## Summary by Sourcery

Report and surface unsendable payment reminders and delivery failures so users can see which invoices were not mailed and why.

New Features:
- Return unsendable payment reminders from the bundling action with invoice and reason details instead of silently skipping them.
- Show a warning toast in the payment reminder run UI summarizing reminders that were held back and why.

Bug Fixes:
- Prevent payment reminder runs from marking invoices as sent when their reminder level is misconfigured or their mail cannot be delivered.
- Fail the payment reminder send job and monitored batch on mail delivery errors so failed reminders no longer appear as successful.

Enhancements:
- Adjust the payment reminder bundling action to only queue jobs for fully configured reminder levels and to still send valid reminders within the same run.
- Update the payment reminder run component to mark only successfully queued invoices as sent based on the bundling result.

Tests:
- Extend payment reminder bundling tests to cover missing reminder texts, missing email templates, and mixed configured/unconfigured orders.
- Add a job test ensuring payment reminder mail failures fail the job and leave no reminder record.
- Add a Livewire test ensuring unsendable invoices remain visible in the reminder run and are not queued.


src/Livewire/Accounting/PaymentReminderRun.php
```diff
@@ -247,6 +243,43 @@ public function toggleGroup(string $key): void
...
+     * the run. The rest stay listed together with the reason they were held back, so
+     * a misconfigured reminder level can no longer look like a completed run.
+     */
+    protected function handleSendResult(?array $result, array $orderIds): void
+    {
+        if (is_null($result)) {
+            return;
+        }
+
+        $unsendable = data_get($result, 'unsendable', []);
```
SirSplasch: ?? [] instead of default

SirSplasch (changes requested):

SirSplasch (approved):

PR #1956, opened 2026-08-03
Allow editing a single payment reminder mail before sending

patrickweh:
## Problem

The reminder run only offers the bulk send, which mails every selected invoice without user interaction. The screen it replaced was built on the mail dialog, so the user could adjust the reminder text per customer before it went out. That option was lost, and there is currently no way to send a reminder with an individually written text.

## Change

- `PaymentReminderRun` uses `CreatesDocuments`, the same path `OrderList` and `Order` already use for their mails.
- A per invoice action opens the create documents modal, creates the payment reminder for the next level and hands it to the mail dialog.
- The reminder text of the matching level supplies the default email template and the subject, the invoice PDF is attached next to the reminder document, and the editable group recipient wins over the resolved address.
- Only the invoice that was actually handed over is marked as sent, the bulk run is untouched and keeps using the queued batch.

Editing a mail before sending only makes sense for a single reminder, so the action is deliberately limited to one invoice.

## Tests

`PaymentReminderRunTest`:
- a single invoice can be edited as a mail before it is sent (dispatches `createFromSession`, creates exactly one reminder at the next level)
- the reminder mail defaults to the template of the reminder level (asserts `default_template_id` and recipient in the session payload)

## Summary by Sourcery

Allow sending individual payment reminder emails via the mail dialog for a single invoice while keeping the existing bulk reminder run unchanged.

New Features:
- Add per-invoice action to open the create-documents mail dialog for a single payment reminder with editable content before sending.

Enhancements:
- Integrate PaymentReminderRun with the shared CreatesDocuments flow to generate reminder documents, attach invoices, and prefill email metadata (subject, template, recipients, language).

Tests:
- Add Livewire tests ensuring a single invoice can be mailed via the edit flow and that the reminder email defaults to the correct template and recipient based on the reminder level.

### DIFF
--- a/src/Livewire/Accounting/PaymentReminderRun.php
+++ b/src/Livewire/Accounting/PaymentReminderRun.php
@@ -4,23 +4,28 @@ namespace FluxErp\Livewire\Accounting;
 
 use Carbon\Carbon;
 use FluxErp\Actions\PaymentReminder\BundlePaymentReminders;
+use FluxErp\Actions\PaymentReminder\CreatePaymentReminder;
+use FluxErp\Contracts\OffersPrinting;
 use FluxErp\Models\Order;
 use FluxErp\Models\PaymentReminder;
 use FluxErp\Traits\Livewire\Actions;
+use FluxErp\Traits\Livewire\CreatesDocuments;
 use FluxErp\View\Printing\PaymentReminder\PaymentReminderView;
 use Illuminate\Contracts\View\View;
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Support\Collection;
 use Illuminate\Validation\ValidationException;
+use Laravel\SerializableClosure\SerializableClosure;
 use Livewire\Attributes\Computed;
 use Livewire\Attributes\Locked;
+use Livewire\Attributes\Renderless;
 use Livewire\Component;
 use Spatie\Permission\Exceptions\UnauthorizedException;
 use Throwable;
 
 class PaymentReminderRun extends Component
 {
-    use Actions;
+    use Actions, CreatesDocuments;
 
     #[Locked]
     public array $groups = [];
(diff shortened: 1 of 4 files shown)


src/Livewire/Accounting/PaymentReminderRun.php
```diff
@@ -200,6 +213,58 @@ public function preview(int $orderId): void
...
+
+    public function createDocuments(): void
+    {
+        $order = resolve_static(Order::class, 'query')
+            ->whereKey($this->documentOrderId)
+            ->wherePaymentReminderEligible()
+            ->first();
+
+        if (! $order) {
+            return;
```
SirSplasch, line 250: intentionally not giving user feedback?

src/Livewire/Accounting/PaymentReminderRun.php
```diff
@@ -247,6 +312,69 @@ public function toggleGroup(string $key): void
             : array_values(array_unique(array_merge($this->selectedOrders, $groupIds)));
     }
 
+    protected function getAttachments(OffersPrinting $item): array
+    {
+        $invoice = $item->order->invoice();
+
+        return $invoice
+            ? ['id' => $invoice->getKey(), 'name' => $invoice->file_name]
```
SirSplasch: wrap the array

src/Livewire/Accounting/PaymentReminderRun.php
```diff
@@ -247,6 +312,69 @@ public function toggleGroup(string $key): void
...
+    }
+
+    protected function getCommunicatableId(OffersPrinting $item): int
+    {
+        return $item->order_id;
+    }
+
+    protected function getCommunicatableType(OffersPrinting $item): string
+    {
+        return app(Order::class)->getMorphClass();
```
SirSplasch: morph_alias()

src/Livewire/Accounting/PaymentReminderRun.php
```diff
@@ -247,6 +312,69 @@ public function toggleGroup(string $key): void
...
+    }
+
+    /** The group recipient stays editable, so it wins over the resolved address. */
+    protected function getTo(OffersPrinting $item, array $documents): array
+    {
+        $groupKey = $item->order->contact_id . '-' . $item->reminder_level;
+        $email = data_get($this->recipientEmails, $groupKey)
+            ?: $item->order->resolveMailablePaymentReminderAddress()?->email_primary;
+
+        return array_filter([$email]);
```
SirSplasch: Arr::wrap()? If the data_get() returns an array, it will be an array in an array.

SirSplasch (changes requested):

SirSplasch (approved):

PR #1957, opened 2026-08-03
feat(order): bring back marking orders as paid in bulk

patrickweh:
## What is missing

The order list has no way to settle several invoices at once. Until the payment reminder screen was reworked in #1845 the removed `Livewire\Accounting\PaymentReminder` carried a **Mark as paid** select action that looped `UpdateLockedOrder` over the selected orders. The component went away with the route, and the function went with it.

## What this adds

A **Mark as paid** select action on the order list, next to Create Documents and Delete, gated by the `UpdateLockedOrder` permission and guarded by a confirmation.

It runs as a monitored batch through `BulkExecutor` rather than a synchronous loop, so settling a full page of invoices does not block the request and the user gets the same progress toast as other bulk work. That also continues the move of bulk select actions onto the bulk executor, following `Livewire\Task\TaskList`.

Orders that are already paid are filtered out before dispatching, they have no transition left to make and would only produce failing jobs. Orders in a payment run keep working, the state machine allows the transition from those states.

`UpdateLockedOrder` now extends `DispatchableFluxAction` and implements `SupportsBulkExecution`, which is what the bulk executor requires. The change is additive, every existing synchronous caller keeps working unchanged.

## Note for operations

The action is asynchronous now. The payment states change once the queue has worked through the batch, not within the request.

## Summary by Sourcery

Add an asynchronous bulk "Mark as paid" action to the order list using the bulk executor and an updated order update action.

New Features:
- Introduce a "Mark as paid" bulk action on the order list, gated by existing permissions and confirmation, to settle multiple invoices at once.

Enhancements:
- Adapt the UpdateLockedOrder action to be dispatchable and support bulk execution for use with the BulkExecutor.
- Provide user feedback via toasts when no eligible orders are selected for marking as paid.

Tests:
- Add Livewire and action tests to verify that marking orders as paid dispatches one job per open order, correctly settles payment state, and no jobs are dispatched without a selection.


src/Livewire/Order/OrderList.php
```diff
@@ -149,6 +161,47 @@ public function fetchContactData(): void
...
 
+    /**
+     * Settling invoices by hand is a bulk job in practice, so it runs as a
+     * monitored batch instead of blocking the request.
+     */
+    #[Renderless]
+    public function markAsPaid(): void
+    {
+        $orderIds = resolve_static(Order::class, 'query')
+            ->whereIntegerInRaw('id', $this->getSelectedValues())
```
SirSplasch: whereKey()

SirSplasch (changes requested):

SirSplasch (approved):

PR #1964, opened 2026-08-04
Let an address change contacts

patrickweh:
An address could not leave its contact. A person filed under a company who moves to another one, or whose private address should stand on its own, could only be fixed in the database.

`Address` has handled a changing `contact_id` for a while: it reassigns the main, invoice and delivery roles of the contact left behind, and deletes that contact once its last address is gone. Nothing exposed that path, so this adds the two ways in.

- **`MoveAddress`** hands an address to an existing contact.
- **`DetachAddress`** creates a new contact for it, inheriting agent, currency, language, payment type, price list and tenants from the contact it came from, and refuses to leave a contact without any address.

Two actions rather than one, so `action.address.move` and `action.address.detach` can be granted apart.

Both sit next to the existing buttons in the address panel of a contact, each behind its own `@canAction`. Moving redirects to the target contact, which also avoids the dead page you would be left on after moving the last address of a contact — that contact no longer exists at that point.

One thing worth knowing for reviewers: leaving a contact clears `is_main_address` through a query update the instance never sees, so setting the flag before the move is silently a no-op. `MoveAddress` refreshes after saving and only then sets it, otherwise a target that had no addresses ends up with no main address at all. Covered by a test.

## Summary by Sourcery

Enable reassigning addresses between contacts or into new contacts while preserving main address semantics and preventing contacts from being left without any address.

New Features:
- Allow moving an address to a different existing contact via a MoveAddress action and UI control.
- Allow detaching an address into a newly created contact via a DetachAddress action and UI control, while inheriting key commercial data from the original contact.

Bug Fixes:
- Ensure a contact without addresses cannot be created by detaching its last remaining address.
- Guarantee that a contact receiving a moved address always ends up with a main address set.

Enhancements:
- Surface address move and detach operations in the contact address panel with appropriate permissions and navigation redirects.

Tests:
- Add Livewire tests for moving and detaching addresses from the contact address component.
- Add action-level tests for MoveAddress and DetachAddress, covering validation and main-address handling.


src/Actions/Address/DetachAddress.php
```diff
@@ -0,0 +1,72 @@
...
+            ->whereKey($this->getData('id'))
+            ->first(['id', 'contact_id']);
+
+        if (resolve_static(Address::class, 'query')
+            ->where('contact_id', $address->contact_id)
+            ->whereKeyNot($address->getKey())
+            ->doesntExist()
+        ) {
+            throw ValidationException::withMessages([
+                'id' => [__('The contact would be left without an address.')],
```
SirSplasch: remove __()

src/Actions/Address/DetachAddress.php
```diff
@@ -0,0 +1,72 @@
...
+
+    protected function getRulesets(): string|array
+    {
+        return DetachAddressRuleset::class;
+    }
+
+    public function performAction(): Address
+    {
+        $address = resolve_static(Address::class, 'query')
+            ->whereKey($this->data['id'])
```
SirSplasch: $this->getData('id')

src/Actions/Address/DetachAddress.php
```diff
@@ -0,0 +1,72 @@
...
+    {
+        parent::validateData();
+
+        $address = resolve_static(Address::class, 'query')
+            ->whereKey($this->getData('id'))
+            ->first(['id', 'contact_id']);
+
+        if (resolve_static(Address::class, 'query')
+            ->where('contact_id', $address->contact_id)
+            ->whereKeyNot($address->getKey())
```
SirSplasch, line 62: whereKeyNot() before where('contact_id')

src/Rulesets/Address/MoveAddressRuleset.php
```diff
@@ -0,0 +1,29 @@
...
+        return [
+            'id' => [
+                'required',
+                'integer',
+                app(ModelExists::class, ['model' => Address::class]),
+            ],
+            'contact_id' => [
+                'required',
+                'integer',
+                app(ModelExists::class, ['model' => Contact::class, 'subject' => Address::class]),
```
SirSplasch: subject check can be removed, as this is no "true" update.

src/Actions/Address/MoveAddress.php
```diff
@@ -0,0 +1,64 @@
...
+    {
+        parent::validateData();
+
+        $currentContactId = resolve_static(Address::class, 'query')
+            ->whereKey($this->getData('id'))
+            ->value('contact_id');
+
+        if ($currentContactId === $this->getData('contact_id')) {
+            throw ValidationException::withMessages([
+                'contact_id' => [__('The address already belongs to this contact.')],
```
SirSplasch: remove __()

src/Actions/Address/MoveAddress.php
```diff
@@ -0,0 +1,64 @@
...
+    {
+        return [Address::class];
+    }
+
+    protected function getRulesets(): string|array
+    {
+        return MoveAddressRuleset::class;
+    }
+
+    public function performAction(): Address
```
SirSplasch, line 22: use $this->getData() instead of direct array access.

SirSplasch (changes requested):
Add a Test for the DetachAddress action

PR #1965, opened 2026-08-04
Revert "Let an address change contacts"

patrickweh:
Reverts #1964, which was merged before its review had happened. Nothing was released, so `v1.7.2` is unaffected and no installation has this code.

The branch `feature/move-address-between-contacts` is restored and unchanged, so the work is not lost and can be reviewed there.

Note for whoever brings it back: git records these commits as reverted, so merging the feature branch again would bring in nothing. Revert this revert first, or rebuild the branch on top of main.

## Summary by Sourcery

Revert the previously introduced feature that allowed moving or detaching addresses between contacts, removing its UI, backend actions, and related tests.

Enhancements:
- Remove address move/detach handling from the contact addresses Livewire component and view to restore prior behavior.

Tests:
- Delete tests covering moving addresses between contacts and detaching addresses into new contacts.

Chores:
- Remove the MoveAddress and DetachAddress actions and their associated rulesets and localization entries.


patrickweh:
Not needed after all — the review had nothing outstanding, so the merge stands.

PR #1968, opened 2026-08-05
Tell the user why an upload was rejected

patrickweh:
Both upload components answered every failure with the same sentence: "Your file upload failed. Please try again." A tenant had no way to tell whether the file was too large, of the wrong type, or something else.

## Why the reason has to come from the browser

Three things were checked before writing this:

- Livewire calls the error callback with no arguments at all (`markUploadErrored` in `livewire.esm.js`), so the `error` parameter the components accepted was always `undefined`.
- The component error bag stays empty. Polled 40 times over 6 seconds in a live browser after a failed upload, it never filled.
- The 422 of the upload endpoint carries only `files.0 could not be uploaded`, which would not help anyone either.

The only place where a precise reason exists is the browser, before the upload starts, where name, size and type are known.

## What changes

`$nuxbe.validateFiles()` checks each file against the effective upload limit and the accept attribute of the input, and both components show what is actually wrong: the file name with its size and the limit, or the accepted types.

`Helper::getMaxUploadSizeInBytes()` derives that limit from the smallest of the livewire rule, `upload_max_filesize` and `post_max_size`. It parses those values itself instead of using the `fromFileSizeToBytes` macro, which throws on legal ini forms like a plain byte count, `0` or `-1`, and would have turned every page containing an upload into a 500 on such a host.

Two details worth naming: the file name is escaped, because the dialog renders its description through `x-html` and a name is fully user controlled. And the accept list is split on commas as well as semicolons, because the purchase invoice upload passes `application/pdf;image/*` and a comma only split would have rejected every file there.

## Tests

Browser tests cover size, type, the semicolon list, a file whose mime type the browser leaves empty, and the escaping. Unit tests cover the limit calculation including the ini forms that used to throw.

## Summary by Sourcery

Improve file upload UX by validating files client-side and surfacing specific rejection reasons to users.

New Features:
- Add client-side file validation for uploads based on maximum size and accepted types, returning detailed rejection metadata for the first invalid file.
- Display contextual error dialogs that name the rejected file, its size, and either the configured size limit or the allowed file types.
- Expose shared upload utilities (file-size formatter, HTML escaping, and validation) via the global $nuxbe object for reuse across upload components.

Bug Fixes:
- Prevent server errors caused by parsing edge-case PHP ini upload size values by introducing a resilient size parsing helper and using it to compute the effective upload limit.
- Ensure file names in upload error dialogs are safely escaped to avoid markup injection in HTML-rendered messages.
- Correct handling of accept lists separated by semicolons or containing wildcard and extension-based patterns so valid files are no longer incorrectly rejected.

Enhancements:
- Centralize computation of the effective maximum upload size by combining Livewire rules with PHP ini limits and using the minimum applicable value.
- Extend both media upload components to clear invalid file inputs and reuse shared client-side validation and error presentation logic.

Tests:
- Add browser tests covering client-side upload validation scenarios, including size and type rejections, semicolon-separated accept lists, empty MIME types, and HTML escaping of file names.
- Add unit tests for maximum upload size calculation and ini value parsing, including previously problematic forms like plain byte counts, zero, negative, and malformed values.

### DIFF
--- a/resources/js/nuxbe/uploads.js
+++ b/resources/js/nuxbe/uploads.js
@@ -1,3 +1,5 @@
+import { getLocale } from './format.js';
+
 export function formatFileSize(bytes) {
     const units = ['B', 'KB', 'MB', 'GB'];
     let size = Number(bytes) || 0;
@@ -10,13 +12,10 @@ export function formatFileSize(bytes) {
 
     const decimals = size < 10 && unit > 0 ? 1 : 0;
 
-    const formatted = new Intl.NumberFormat(
-        document.documentElement.lang || 'de',
-        {
-            minimumFractionDigits: decimals,
-            maximumFractionDigits: decimals,
-        },
-    ).format(size);
+    const formatted = new Intl.NumberFormat(getLocale(), {
+        minimumFractionDigits: decimals,
+        maximumFractionDigits: decimals,
+    }).format(size);
 
     return `${formatted} ${units[unit]}`;
 }
@@ -58,9 +57,13 @@ function matchesAccept(file, patterns) {
     }
 
     return patterns.some((pattern) => {
-        if (pattern.startsWith('.')) return name.endsWith(pattern);
-        if (pattern.endsWith('/*'))
+        if (pattern.startsWith('.')) {
+            return name.endsWith(pattern);
+        }
+
+        if (pattern.endsWith('/*')) {
             return type.startsWith(pattern.slice(0, -1));
+        }
 
         return type === pattern;
     });
@@ -80,7 +83,11 @@ export function validateFiles(files, { maxSize = 0, accept = '' } = {}) {
         }
 
         if (patterns.length && !matchesAccept(file, patterns)) {
-            return { file, reason: 'type', accept: patterns.join(', ') };
+            return {
+                file,
+                reason: 'type',
+                accept: patterns.join(', '),
+            };
         }
     }
 
--- a/resources/js/nuxbe/uploads.js
+++ b/resources/js/nuxbe/uploads.js
@@ -1,0 +1,88 @@
+export function formatFileSize(bytes) {
+    const units = ['B', 'KB', 'MB', 'GB'];
+    let size = Number(bytes) || 0;
+    let unit = 0;
+
+    while (size >= 1024 && unit < units.length - 1) {
+        size /= 1024;
+        unit++;
+    }
+
+    const decimals = size < 10 && unit > 0 ? 1 : 0;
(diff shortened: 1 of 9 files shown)


resources/js/nuxbe/uploads.js
```diff
@@ -0,0 +1,88 @@
...
+
+    while (size >= 1024 && unit < units.length - 1) {
+        size /= 1024;
+        unit++;
+    }
+
+    const decimals = size < 10 && unit > 0 ? 1 : 0;
+
+    const formatted = new Intl.NumberFormat(
+        document.documentElement.lang || 'de',
```
SirSplasch: Why not fallback to 'en'?

resources/js/nuxbe/uploads.js
```diff
@@ -0,0 +1,88 @@
...
+
+        return (
+            extensions.length === 0 ||
+            extensions.some((pattern) => name.endsWith(pattern))
+        );
+    }
+
+    return patterns.some((pattern) => {
+        if (pattern.startsWith('.')) return name.endsWith(pattern);
+        if (pattern.endsWith('/*'))
```
SirSplasch: please add {} to the if statements and wrap it accordingly.

resources/js/nuxbe/uploads.js
```diff
@@ -0,0 +1,88 @@
...
+            return {
+                file,
+                reason: 'size',
+                size: formatFileSize(file.size),
+                maxSize: formatFileSize(maxSize),
+            };
+        }
+
+        if (patterns.length && !matchesAccept(file, patterns)) {
+            return { file, reason: 'type', accept: patterns.join(', ') };
```
SirSplasch: Why is this object not wrapped, while the maxSize object is?

src/Helpers/Helper.php
```diff
@@ -106,6 +107,27 @@ public static function classExists(string $classString,
         }
     }
 
+    public static function getMaxUploadSizeInBytes(): int
```
SirSplasch: after getHtmlInputFieldTypes() method

SirSplasch (changes requested):

resources/js/nuxbe/uploads.js
SirSplasch: Question not answered, just moved the issue to a different location.

SirSplasch (changes requested):
I still need an answer on why the default locale is 'de' instead of 'en'.

SirSplasch (approved):

PR #1972, opened 2026-08-06
Say what the payment run fields do

patrickweh:
Collective, Single Booking, BASIC and B2B are terms somebody has to already know, and the toggle that greys out until another one is on explains itself nowhere. Six decisions stand between a user and collecting money.

Hints only. No default, no behaviour and no field changed, so nothing about what reaches the bank moves.

## Summary by Sourcery

Clarify payment run form fields by adding explanatory hints for key options without altering their behavior.

Enhancements:
- Add descriptive hints to execution date, SEPA mandate type, collective, single booking, and instant payment fields in the payment run form to explain their effects to users.

Documentation:
- Extend German and English translation resources with copy for the new payment run field hints.


lang/en.json
SirSplasch, line 1: Are you intentionally adding these keys although the keys and the translations do not differ?

SirSplasch (changes requested):

lang/en.json
patrickweh: Not intentional. Checked `en.json` on main: 309 entries, none where key equals value. The file maps symbolic keys like `action.address.create` onto English sentences, and these source strings are already English, so the fallback covers them. Removed, `en.json` is back to 309 entries.

`de.json` keeps its five, where the convention is the opposite: 2236 of its keys are English sentences.

SirSplasch (approved):

PR #1973, opened 2026-08-06
Resolve address and media routes by primary key

patrickweh:
Clicking an address inside a contact answered with a 404 on every contact, while every other link on the page worked.

## Cause

Three routes use implicit route model binding, so laravel looks the record up through `getRouteKeyName()`. An installation that runs a public site next to flux in the same application can legitimately point that at a slug, and one of ours does exactly that: it binds its own Address model, which returns `slug`, into the container.

The url itself is built by flux with the id. `Address::detailRouteParams()` returns `'address' => $this->id`, and `/address/{address}` then tries to resolve that id through the slug column. Flux therefore hands out a link it can no longer read itself, and the result is a 404 that looks like a missing record.

## Change

`/address/{address}` and the two media routes resolve their record through `whereKey()` and only accept numeric parameters. The route key stays untouched and remains free for whatever the surrounding application needs it for.

A test binds an Address whose route key is a slug and asserts the redirect still works, next to one asserting an unknown id still gives a 404.

## Summary by Sourcery

Resolve address and media routes by primary key instead of implicit route model binding to avoid conflicts with customized route keys.

Bug Fixes:
- Fix address detail links returning 404 when the application overrides the Address route key to use a slug.
- Fix media download and public media routes failing when Media uses a non-numeric route key.

Enhancements:
- Restrict address and media routes to numeric parameters and look up records via primary key queries instead of implicit binding.

Tests:
- Add feature tests ensuring the address redirect route works with customized Address route keys and still returns 404 for unknown IDs.

### DIFF
--- a/routes/frontend/web.php
+++ b/routes/frontend/web.php
@@ -198,15 +198,22 @@ Route::middleware('web')
                     });
                 Route::get(
                     '/address/{address}',
-                    fn (Address $address) => redirect()
-                        ->route(
-                            'contacts.id?',
-                            [
-                                'id' => $address->contact_id,
-                                'address' => $address->getKey(),
-                            ]
-                        )
+                    function (int $address) {
+                        $address = resolve_static(Address::class, 'query')
+                            ->whereKey($address)
+                            ->firstOrFail();
+
+                        return redirect()
+                            ->route(
+                                'contacts.id?',
+                                [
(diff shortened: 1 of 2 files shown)


patrickweh:
Closing this. The route key is a decision of the surrounding application, so the fix belongs there, not in core.

PR #1978, opened 2026-08-08
feat(loan): extra repayments with allowance and savings

patrickweh:
Stacked on #1976, it uses the fixed installment mode and the grace period from there. Merge that one first.

A loan could only follow the plan it was created with. An extra repayment had nowhere to go, and what the contract allows as an extra repayment was not recorded at all.

## Changes

- `loans.allows_extra_repayments` covers the loan that permits none. When they are permitted, `extra_repayment_allowance_percentage` (share of the original amount) and `extra_repayment_allowance_amount` cap them per calendar year; both empty means uncapped. The three fields stay editable after creation, they are contract terms and not schedule maths.
- `loan_extra_repayments` records the payment with its date, a note and the adjustment it triggers:
  - `shorten_term` keeps the installment and ends the schedule earlier
  - `reduce_installment` keeps the term and lowers the installment
- Only the open tail is rescheduled. Installments already settled, and the transactions assigned to them, are untouched, and the sequence continues where they stop.
- An extra repayment during the grace period keeps the remaining interest only installments and only lowers their interest, it does not start the repayment early.
- `interest_saved` and `installments_saved` are stored on the record and shown live in the modal while the amount is entered, together with the new remaining principal and end date.
- Validation rejects a loan that allows none, an amount beyond the outstanding principal, and an amount beyond what the allowance leaves for that year.

## Notes

- Deleting an extra repayment does not roll the schedule back, installments settled in the meantime would no longer match. There is no delete action for that reason.

## Summary by Sourcery

Add support for recording and applying extra loan repayments with configurable yearly allowances and schedule adjustments.

New Features:
- Introduce contract-level fields to control whether a loan allows extra repayments and to define capped or unlimited yearly extra repayment allowances.
- Add a LoanExtraRepayment model, action, Livewire form, and UI tab to capture extra repayments, including schedule adjustment type, note, and calculated savings.
- Provide live preview of the impact of a proposed extra repayment on interest, installment count, remaining principal, and loan end date.

Enhancements:
- Implement an ExtraRepaymentScheduler to reschedule only the open tail of a loan when an extra repayment is made, preserving settled installments and sequences.
- Extend loan domain logic to compute yearly extra repayment allowance, track used allowance per year, and derive remaining allowance.
- Register loan extra repayments in the polymorphic morph map for consistent linking and navigation.

Tests:
- Add feature and Livewire tests covering extra repayment booking, allowance enforcement, schedule behavior across annuity, linear, grace-period scenarios, and validation of disallowed or excessive repayments.


database/migrations/2026_08_08_000003_add_extra_repayment_allowance_to_loans_table.php
```diff
@@ -0,0 +1,34 @@
...
+use Illuminate\Support\Facades\Schema;
+
+return new class() extends Migration
+{
+    public function up(): void
+    {
+        Schema::table('loans', function (Blueprint $table): void {
+            $table->boolean('allows_extra_repayments')
+                ->default(true)
+                ->after('grace_period_installments');
```
SirSplasch: after ends_at, as this is a boolean

database/migrations/2026_08_08_000003_add_extra_repayment_allowance_to_loans_table.php
```diff
@@ -0,0 +1,34 @@
...
+            $table->decimal('extra_repayment_allowance_percentage', 40, 10)
+                ->nullable()
+                ->after('allows_extra_repayments');
```
SirSplasch: As allows_extra_repayments is moved, this should be after another column.

database/migrations/2026_08_08_000003_add_extra_repayment_allowance_to_loans_table.php
```diff
@@ -0,0 +1,34 @@
...
+
+return new class() extends Migration
+{
+    public function up(): void
+    {
+        Schema::table('loans', function (Blueprint $table): void {
+            $table->boolean('allows_extra_repayments')
+                ->default(true)
+                ->after('grace_period_installments');
+            $table->decimal('extra_repayment_allowance_percentage', 40, 10)
```
SirSplasch: Why is this column 40, 10, while this is a percentage value? Can this exceed 100%?

database/factories/LoanFactory.php
```diff
@@ -19,6 +20,9 @@ public function definition(): array
             'interest_rate' => fake()->randomFloat(4, 0, 0.1),
             'repayment_type_enum' => fake()->randomElement(RepaymentTypeEnum::cases()),
             'number_of_installments' => fake()->numberBetween(12, 120),
+            'installment_interval_enum' => InstallmentIntervalEnum::Monthly,
+            'grace_period_installments' => 0,
+            'allows_extra_repayments' => true,
```
SirSplasch: move this accordingly

database/factories/LoanFactory.php
```diff
@@ -19,6 +20,9 @@ public function definition(): array
             'interest_rate' => fake()->randomFloat(4, 0, 0.1),
             'repayment_type_enum' => fake()->randomElement(RepaymentTypeEnum::cases()),
             'number_of_installments' => fake()->numberBetween(12, 120),
+            'installment_interval_enum' => InstallmentIntervalEnum::Monthly,
+            'grace_period_installments' => 0,
```
SirSplasch: Don't you want to add the extra_repayment columns?

resources/views/components/loan/extra-repayments.blade.php
SirSplasch: Make the $this->allowance and $this->extraRepaymentPreview array accesses null-safe.

src/Models/Loan.php
```diff
@@ -31,6 +31,9 @@ protected function casts(): array
             'amount' => Money::class,
             'repayment_type_enum' => RepaymentTypeEnum::class,
             'installment_interval_enum' => InstallmentIntervalEnum::class,
+            'allows_extra_repayments' => 'boolean',
+            'extra_repayment_allowance_percentage' => 'decimal:10',
+            'extra_repayment_allowance_amount' => Money::class,
```
SirSplasch: move accordingly

src/Rulesets/Loan/CreateLoanRuleset.php
```diff
@@ -67,6 +67,15 @@ public function rules(): array
                 app(EnumRule::class, ['type' => InstallmentIntervalEnum::class]),
             ],
             'grace_period_installments' => 'nullable|integer|min:0',
+            'allows_extra_repayments' => 'nullable|boolean',
```
SirSplasch: Move accordingly

src/Rulesets/Loan/CreateLoanRuleset.php
```diff
@@ -67,6 +67,15 @@ public function rules(): array
...
+            'extra_repayment_allowance_percentage' => [
+                'nullable',
+                app(Numeric::class, ['min' => 0, 'max' => 1]),
+            ],
+            'extra_repayment_allowance_amount' => [
+                'nullable',
+                app(Numeric::class, ['min' => 0]),
+            ],
```
SirSplasch: Maybe exclude these, if allows_extra_repayments is null or not given?

src/Rulesets/Loan/UpdateLoanRuleset.php
SirSplasch: Same things as in the CreateLoanRuleset.

src/Actions/LoanExtraRepayment/CreateLoanExtraRepayment.php
```diff
@@ -0,0 +1,119 @@
...
+                    __('The extra repayment exceeds the allowance of :amount left for :year.', [
+                        'amount' => $remainingAllowance,
+                        'year' => Carbon::parse($this->getData('executed_at'))->year,
+                    ]),
+                ],
+            ]);
+        }
+    }
+
+    protected function loan(): Loan
```
SirSplasch: I don't know about this method, it is inconsistent to every other action.

src/Actions/LoanExtraRepayment/CreateLoanExtraRepayment.php
```diff
@@ -0,0 +1,119 @@
...
+                ],
+            ]);
+        }
+
+        $remainingAllowance = $loan->remainingExtraRepaymentAllowance(
+            Carbon::parse($this->getData('executed_at'))->year
+        );
+
+        if (! is_null($remainingAllowance) && bccomp($amount, $remainingAllowance, 2) === 1) {
+            throw ValidationException::withMessages([
```
SirSplasch: Does it make sense to pool the last two validation exceptions and throw one validation exception at the end?

src/Actions/LoanExtraRepayment/CreateLoanExtraRepayment.php
```diff
@@ -0,0 +1,119 @@
...
+                'amount' => [
+                    __('The extra repayment exceeds the outstanding principal of :amount.', [
+                        'amount' => $outstanding,
+                    ]),
+                ],
+            ]);
+        }
+
+        $remainingAllowance = $loan->remainingExtraRepaymentAllowance(
+            Carbon::parse($this->getData('executed_at'))->year
```
SirSplasch: Can write the year into a variable, so that this statement does not need to be executed twice.

src/Models/Loan.php
```diff
@@ -71,9 +79,57 @@ public function tenant(): BelongsTo
     }
 
     // Public methods
-    /**
-     * The repaid share of the loan, between 0 and 1.
-     */
+    public function extraRepaymentAllowance(): ?string
```
SirSplasch: These newly added methods are not ordered alphabetical.

src/Models/Loan.php
```diff
@@ -71,9 +79,57 @@ public function tenant(): BelongsTo
...
+    {
+        if (! $this->allows_extra_repayments) {
+            return '0.00';
```
SirSplasch: Why return '0.00' here, while you return null in the next if condition?

src/Models/Loan.php
```diff
@@ -71,9 +79,57 @@ public function tenant(): BelongsTo
...
+        }
+
+        if (is_null($this->extra_repayment_allowance_percentage)
+            && is_null($this->extra_repayment_allowance_amount)
+        ) {
+            return null;
+        }
+
+        return bcround(
```
SirSplasch: How does this work if extra_repayment_allowance_percentage and extra_repayment_allowance_amount is given?
Currently the results are added, but that does not make sense to me. I would interpret this as the maximum is the lower of these two values or maybe simply the flat amount takes precedence.

src/Rulesets/LoanExtraRepayment/CreateLoanExtraRepaymentRuleset.php
```diff
@@ -0,0 +1,38 @@
...
+    public function rules(): array
+    {
+        return [
+            'uuid' => 'nullable|string|uuid|unique:loan_extra_repayments,uuid',
+            'loan_id' => [
+                'required',
+                'integer',
+                app(ModelExists::class, ['model' => Loan::class]),
+            ],
+            'executed_at' => 'required|date',
```
SirSplasch: Just to be sure, this can be any date, like even years in the past?

src/Actions/LoanExtraRepayment/CreateLoanExtraRepayment.php
```diff
@@ -0,0 +1,119 @@
...
+        ]);
+        $loan->calculateRemaining()
+            ->calculateTotalInterest()
+            ->calculateProgress()
+            ->save();
+
+        return $extraRepayment->refresh();
+    }
+
+    protected function validateData(): void
```
SirSplasch: I am not sure, but what about validating the executed_at date. Like should it be ensured, that the executed_at date is actually after the loan has been started and possibly before it ends?

src/Actions/LoanExtraRepayment/CreateLoanExtraRepayment.php
```diff
@@ -0,0 +1,119 @@
...
+            $extraRepayment->amount,
+            $extraRepayment->schedule_adjustment_type_enum,
+            $open
+        );
+
+        $extraRepayment->fill($scheduler->savings($schedule, $open));
+        $extraRepayment->save();
+
+        $loan->installments()
+            ->whereIn('id', $open->modelKeys())
```
SirSplasch: whereKey()?

src/Actions/LoanExtraRepayment/CreateLoanExtraRepayment.php
```diff
@@ -0,0 +1,119 @@
...
+            ->delete();
+
+        foreach ($schedule as $installment) {
+            $loan->installments()->create($installment);
+        }
+
+        $lastInstallment = array_last($schedule);
+        $firstRepayment = array_first(
+            array_filter($schedule, fn (array $installment): bool => bccomp($installment['principal_amount'], '0', 2) === 1)
```
SirSplasch: Maximum line length exceeded

src/Livewire/Accounting/Loan.php
```diff
@@ -165,6 +187,76 @@ public function finance(): bool
...
+            ['index' => 'note', 'label' => __('Note')],
+        ];
+    }
+
+    #[Computed]
+    public function extraRepaymentPreview(): array
+    {
+        $amount = (float) $this->extraRepayment->amount;
+
+        if ($amount <= 0) {
```
SirSplasch: Why not use bccomp?

src/Support/Calculation/ExtraRepaymentScheduler.php
SirSplasch: Please refactor this class the following way.
- Add the Makeable trait
- Pass Loan to this class as it's being instanced and store it in a protected property
- As you have stored the loan, no need to pass it to every method
- What about making the core methods chainable?
    - Then you might need to add accessors if you want to have access to for example the schedule or the open installments.
- order methods alphabetical (sorted by public, protected, etc.)

SirSplasch (changes requested):

src/Models/Loan.php
patrickweh: '0.00' means nothing is allowed, null means no limit is configured at all. `remainingExtraRepaymentAllowance()` passes the null through so the ui can show unlimited, the flag off case has to cap at zero

src/Rulesets/Loan/CreateLoanRuleset.php
patrickweh: no, the column defaults to true, so a missing flag means extra repayments are allowed and excluding would drop a valid allowance. same in the update ruleset, the flag is `sometimes` there, so you could never change the allowance on its own

src/Models/Loan.php
patrickweh: made them mutually exclusive instead, `prohibits` on the percentage. either a percentage or a flat amount, so theres nothing left to add up

src/Rulesets/LoanExtraRepayment/CreateLoanExtraRepaymentRuleset.php
patrickweh: not any more, executed_at has to be on or after the loans starts_at now. no upper bound on purpose, ends_at moves with every extra repayment and booking one ahead is fine

src/Actions/LoanExtraRepayment/CreateLoanExtraRepayment.php
```diff
@@ -0,0 +1,117 @@
...
+            ->save();
+
+        return $extraRepayment->refresh();
+    }
+
+    protected function validateData(): void
+    {
+        parent::validateData();
+
+        $loan = $this->loan = resolve_static(Loan::class, 'query')
```
SirSplasch: Do you need to create a local variable instead of working with the property?

src/Actions/LoanExtraRepayment/CreateLoanExtraRepayment.php
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        $executedAt = Carbon::parse($this->getData('executed_at'));
+        $amount = bcadd((string) $this->getData('amount'), '0', 2);
+        $outstanding = bcadd((string) $loan->remaining, '0', 2);
+        $remainingAllowance = $loan->remainingExtraRepaymentAllowance($executedAt->year);
+        $errors = [];
+
+        if ($executedAt->lt($loan->starts_at)) {
+            $errors['executed_at'][] = __('The extra repayment cannot be executed before the loan starts.');
```
SirSplasch: No need to use __()

src/Actions/LoanExtraRepayment/CreateLoanExtraRepayment.php
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    protected function getRulesets(): string|array
+    {
+        return CreateLoanExtraRepaymentRuleset::class;
+    }
+
+    public function performAction(): LoanExtraRepayment
+    {
+        $loan = $this->loan ?? resolve_static(Loan::class, 'query')
```
SirSplasch: Why not $this->loan ??= resolve_static() and work with $this->loan instead of the local variable?

src/Actions/LoanExtraRepayment/CreateLoanExtraRepayment.php
```diff
@@ -0,0 +1,117 @@
...
+            ]);
+        }
+
+        $executedAt = Carbon::parse($this->getData('executed_at'));
+        $amount = bcadd((string) $this->getData('amount'), '0', 2);
+        $outstanding = bcadd((string) $loan->remaining, '0', 2);
+        $remainingAllowance = $loan->remainingExtraRepaymentAllowance($executedAt->year);
+        $errors = [];
+
+        if ($executedAt->lt($loan->starts_at)) {
```
SirSplasch: It should also probably not be after the loan ends_at, or is the ends_at more vague and of cosmetic nature?

SirSplasch (changes requested):

SirSplasch (approved):

PR #1993, opened 2026-08-11
feat(accounting): list outstanding items again after the reminder run rework

patrickweh:
Reported by a customer after #1845: the **Mahnungen** page was not only the dunning entry point, it doubled as the list of every unpaid invoice. The reminder run cannot fill that role, it deliberately holds only what is due for a reminder today, so an invoice that is open but not yet due disappeared from view.

Adds **Buchhaltung > Offene Posten**, a page on top of `Order\OrderList` whose builder adds the two scopes the model already has:

```php
return parent::getBuilder($builder)
    ->unpaid()
    ->revenue();
```

`unpaid()` is invoice number present, payment state not paid, balance not zero. `revenue()` drops the purchase side. Together they are what the removed page queried.

Building on the order list means filters, column selection, saved filters, export and the existing actions come along, and the page carries its own cache key so column settings do not collide with the order list.

The dashboard widget already redirected here in spirit: #1845 pointed its **Show** and **Show overdue** options at the order list through a `SessionFilter`. This gives the same view a permanent, discoverable place in the menu.

Note for deployment: the route brings a new permission that has to be assigned, same as the reminder run itself.

## Summary by Sourcery

Add an accounting view that lists outstanding revenue invoices independently of the payment reminder run.

New Features:
- Introduce an Outstanding Items accounting page based on the order list to show all unpaid revenue invoices.
- Expose the Outstanding Items view via a new frontend route and register it in the accounting menu.

Tests:
- Add Livewire tests to verify that the Outstanding Items view includes only unpaid revenue invoices and excludes non-invoiced, settled, purchase, or not-yet-remindable orders.

Chores:
- Introduce a new permission associated with the Outstanding Items route that must be assigned on deployment.

### DIFF
--- a/lang/de.json
+++ b/lang/de.json
@@ -1959,6 +1959,7 @@
     "Other": "Sonstige",
     "Other media": "Sonstige Dateien",
     "Outstanding": "Ausstehend",
+    "Outstanding Items": "Offene Posten",
     "Overall Lead Won Lost Ratio": "Gesamtes Lead Gewinn-Verlust-Verhältnis",
     "Overdue": "Überfällig",
     "Overdue Installments": "Überfällige Raten",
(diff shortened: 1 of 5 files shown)


patrickweh:
Wrong approach. The dunning action belongs in the existing order list instead of a separate page.

PR #1994, opened 2026-08-11
feat(order): create payment reminders straight from the order list

patrickweh:
Follow-up to #1845. The removed dunning page carried a bulk action the reminder run cannot replace: it created payment reminders for a freely filtered selection, while the run deliberately holds only what is due today. A customer noticed the gap.

The action moves to **Alle Aufträge**:

- **Zahlungserinnerung erstellen** appears next to **Dokumente erstellen**, gated by `CreatePaymentReminder::canPerformAction()`
- orders without a mailable reminder address are skipped, with a toast naming how many, exactly as the old page did
- the reminder level is taken from the order and raised by one, so a second run produces the next level

Both flows share one modal. `createsPaymentReminders` decides which print layouts are offered, and the item hooks branch on `$item instanceof PaymentReminder`, so mail template, subject, language, blade parameters and recipient resolve per document type. `openCreateDocumentsModal()` resets the flag, so the regular document flow cannot inherit the reminder state. The trait method is aliased to `openDocumentsModal()` because a trait method cannot be reached through `parent::`.

Tests cover the layout switch, the reset on the regular modal, and that an order whose contact has no email address gets no reminder while its neighbour does. 241 tests in `Livewire/Order` and `Livewire/Accounting` pass.

## Summary by Sourcery

Add a bulk action to create payment reminders directly from the order list, sharing the existing document creation flow while handling reminder-specific layouts, templates, and recipients.

New Features:
- Introduce a "Create Payment Reminder" bulk action on the order list to generate reminders for selected orders with mailable addresses.
- Support payment reminder document creation using appropriate print layouts, email templates, blade parameters, languages, and recipient resolution based on the reminder context.

Enhancements:
- Gate the payment reminder action with existing permission checks and ensure the regular document flow cannot accidentally inherit reminder state.
- Show a warning toast indicating how many selected orders were skipped due to missing mailable reminder addresses.

Tests:
- Add livewire tests verifying reminder layout selection in the modal, state reset when switching back to regular documents, and skipping orders without mailable addresses while processing eligible neighbours.

### DIFF
--- a/src/Livewire/Order/OrderList.php
+++ b/src/Livewire/Order/OrderList.php
@@ -406,8 +406,6 @@ class OrderList extends \FluxErp\Livewire\DataTables\OrderList
                 continue;
             }
 
-            // The action returns a fresh model, so the document hooks would query
-            // the order again for every single reminder.
             $reminder->setRelation('order', $order);
 
             $reminders->push($reminder);
--- a/src/Livewire/Order/OrderList.php
+++ b/src/Livewire/Order/OrderList.php
@@ -394,17 +394,23 @@ class OrderList extends \FluxErp\Livewire\DataTables\OrderList
         $reminders = collect();
         foreach ($orders as $order) {
             try {
-                $reminders->push(
-                    CreatePaymentReminder::make([
-                        'order_id' => $order->getKey(),
-                        'reminder_level' => (int) $order->payment_reminder_current_level + 1,
-                    ])
-                        ->validate()
-                        ->execute()
-                );
+                $reminder = CreatePaymentReminder::make([
+                    'order_id' => $order->getKey(),
+                    'reminder_level' => (int) $order->payment_reminder_current_level + 1,
+                ])
+                    ->validate()
+                    ->execute();
             } catch (ValidationException $e) {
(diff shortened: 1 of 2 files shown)


src/Livewire/Order/OrderList.php
```diff
@@ -330,8 +373,64 @@ public function updatedShowMap(): void
...
+        $orders = (clone $baseQuery)
+            ->whereHasMailablePaymentReminderAddress()
+            ->get();
+
+        $skippedCount = (clone $baseQuery)->count() - $orders->count();
+
+        if ($skippedCount > 0) {
+            $this->toast()
+                ->warning(
+                    __(':count order(s) skipped due to missing email address.', ['count' => $skippedCount])
```
SirSplasch: Does this key exist in de.json?

SirSplasch (changes requested):

src/Livewire/Order/OrderList.php
patrickweh: yes, its in de.json since `979f21940`

SirSplasch (approved):
resolve merge conflicts

PR #2021, opened 2026-08-18
Allow renaming a media file in the folder tree

patrickweh:
## Problem

The file detail panel of the folder tree offers download, copy link and delete, but no way to change the name of an uploaded file. Files that arrive with a generated name stay unreadable in the tree, and there is no way to correct them from the UI.

## Change

- `FolderTree::saveMedia()` renames a media item through the existing `UpdateMedia` action, writing both `name` and `file_name`, which also moves the stored file and its generated conversions.
- The media is resolved through `findMedia()` first, so a component can only rename media that belongs to its own model.
- The blade adds a name input plus save button to the file detail panel, mirroring the folder rename block that already exists, gated by `@canAction(UpdateMedia::class)` and hidden for read-only components and collections.
- Two tests cover the rename and the rejection of a foreign media id.

## Summary by Sourcery

Enable secure, permission-controlled renaming of media files from the folder tree.

New Features:
- Allow users to rename media files directly from the folder tree file details panel.

Bug Fixes:
- Prevent renaming media that does not belong to the folder tree's model.

Enhancements:
- Restrict media renaming to writable components with the media update permission and return updated file metadata to refresh the tree.

Tests:
- Add coverage for successful media renaming and rejection of foreign media identifiers.

### DIFF
--- a/resources/views/livewire/folder-tree.blade.php
+++ b/resources/views/livewire/folder-tree.blade.php
@@ -430,6 +430,48 @@
                                     ></div>
                                 </template>
                             </div>
+                            @canAction(\FluxErp\Actions\Media\UpdateMedia::class)
+                                <div
+                                    class="flex flex-col gap-3 md:flex-row md:items-end"
+                                    x-cloak
+                                    x-show="!$wire.isReadonly && !readOnly"
+                                >
+                                    <div class="w-full md:flex-1">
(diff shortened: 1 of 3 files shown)


patrickweh:
Superseded by the branch without the ticket number.

PR #2024, opened 2026-08-18
Allow renaming a media file in the folder tree

patrickweh:
## Problem

The file detail panel of the folder tree offers download, copy link and delete, but no way to change the name of an uploaded file. A file that arrives with a generated name stays unreadable in the tree, and there is no way to correct it from the UI.

## Change

- `FolderTree::saveMedia()` renames a media item through the existing `UpdateMedia` action.
- Only `name` is written. Renaming `file_name` would move the stored file and every generated conversion, which invalidates links that were already copied out, while the tree and the detail panel render `name`.
- The media is resolved through `findMedia()` first, so a component can only rename media that belongs to its own model.
- The blade adds a name input plus save button to the file detail panel, mirroring the folder rename block that already exists, gated by `@canAction(UpdateMedia::class)` and hidden for read-only components and collections.
- Two tests cover the rename and the rejection of a foreign media id.

## Summary by Sourcery

Enable authorized users to rename media files from the folder tree while protecting unrelated media and read-only views.

New Features:
- Add the ability to rename media files directly from the folder tree detail panel.

Enhancements:
- Restrict media renaming to items belonging to the current model and preserve the stored file name.

Tests:
- Add coverage for successful media renaming and rejection of media belonging to another model.

### DIFF
--- a/resources/views/livewire/folder-tree.blade.php
+++ b/resources/views/livewire/folder-tree.blade.php
@@ -430,6 +430,48 @@
                                     ></div>
                                 </template>
                             </div>
+                            @canAction(\FluxErp\Actions\Media\UpdateMedia::class)
+                                <div
+                                    class="flex flex-col gap-3 md:flex-row md:items-end"
+                                    x-cloak
+                                    x-show="!$wire.isReadonly && !readOnly"
+                                >
+                                    <div class="w-full md:flex-1">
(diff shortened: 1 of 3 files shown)


SirSplasch (approved):

SirSplasch:
resolve merge conflicts

